PR #2, opened 2017-11-27
rename castLinear in comments too

gelisam:
"castLinear" was renamed to "toLinear" but comments were still referring
to the function as "castLinear".


aspiwack:
Thanks.

PR #5, opened 2017-12-04
Add a primitive for pairs of unrestricted and linear results

aspiwack:
@mboes adding a useful idiom, but I have little idea about the name. I put one, but it's not good.

@YPares maybe you want to have a go at proposing something.

### DIFF
--- a/src/Prelude/Linear.hs
+++ b/src/Prelude/Linear.hs
@@ -12,6 +12,7 @@ module Prelude.Linear
     -- $ unrestricted
   , Unrestricted(..)
   , unUnrestricted
+  , WithUnrestricted(..)
     -- * Typeclasses for non-linear actions
     -- $ comonoid
   , Consumable(..)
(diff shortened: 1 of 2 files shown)


src/Prelude/Linear.hs
```diff
@@ -70,6 +71,11 @@ data Unrestricted a where
 unUnrestricted :: Unrestricted a ->. a
 unUnrestricted (Unrestricted a) = a
 
+-- | A variant of 'Unrestricted' which captures the very common pattern of
+-- returning a pair of an unrestricted value and a linear value.
+data WithUnrestricted a l where
+  WithUnrestricted :: a -> l ->. WithUnrestricted a l
```
facundominguez: If this was used as an infix operator, the second component should be unrestricted. Not sure if there is a rationale to make it the first.

aspiwack: Forgot to reply, sorry: the rationale was "it looks like the result of the state monad".

I don't care either way. But I'd rather have a shorter name. These things are everywhere.

@mboes once proposed `Poly` instead of `Unrestricted` for "polynomial", which is meaningful (though even more unusual than unrestricted), so why not. We could have this be `WithPoly`, or even something better. I don't really have a good idea. And the name does matter!

mboes: In fact they're in so many places that I'm increasingly partial to adding syntactic sugar for this. e.g.

`a ->. IO ~(Maybe Int)`

We can't use `!` like in linear logic. Not because `!` is already legal in type sigs (it's not). But it already means something completely different in data type field type declarations, as well as in patterns.

mboes: I agree with @facundominguez. It feels natural to allow `Handle ``WithUnrestricted`` Char`

aspiwack:
I need to think more. So let me close this PR to avoid clutter.

PR #6, opened 2017-12-05
Set up a CI

aspiwack:
Just building with `--pedantic` for the moment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,8 +4,6 @@ jobs:
   build:
     docker:
       - image: tweag/linear-types:0.1.6
-    environment:
-      - stack: stack --no-docker
     steps:
       - checkout
       # Restore last version of the dependencies in cache
@@ -17,8 +15,8 @@ jobs:
           # --jobs=1 because of limitted memory
           name: Build dependencies
           command: |
-            $stack build --jobs=1 --only-dependencies --no-haddock linear-base:lib
-            $stack build --jobs=1 --test --bench --only-snapshot
+            stack --no-docker build --jobs=1 --only-dependencies --no-haddock linear-base:lib
+            stack --no-docker build --jobs=1 --test --bench --only-snapshot
             find . -name "*.cabal" -o -name "stack.yaml" -type f | xargs sha256sum > /tmp/dps_checksum
       - save_cache:
           key: linear-base-stackcache-{{ arch }}-v171204-{{ checksum "/tmp/dps_checksum" }}
@@ -26,4 +24,4 @@ jobs:
             - ~/.stack
       - run:
           name: Main build
-          command: $stack build --pedantic
+          command: stack --no-docker build --pedantic
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,8 @@ jobs:
   build:
     docker:
       - image: tweag/linear-types:0.1.6
+    environment:
+      - stack: stack --no-docker
     steps:
       - checkout
       # Restore last version of the dependencies in cache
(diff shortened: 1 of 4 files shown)


aspiwack:
@dleprovost any suggestion to factor writing `stack --no-docker` at every line?

Maybe the best is to write
```yaml
docker:
   enable: false
```
(or whatever the actual syntax is) in ths `/etc/stack/config.yaml`.

We're root on these images right? Maybe I ought to do that directly in the Dockerfile?

src/System/IO/Resource.hs
```diff
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-name-shadowing #-}
```
mboes: Why?

src/Prelude/Linear.hs
```diff
@@ -50,6 +49,8 @@ infixr 0 $
 const :: a ->. b -> a
 const x _ = x
 
+-- In `base`, this is not a prelude function (it's in `Data.Tuple`), maybe we
```
mboes: This comment seems out-of-place. It's neither connected to the goal of this PR nor actionable because not tagged with `TODO` or `XXX`.

.circleci/config.yml
```diff
@@ -0,0 +1,29 @@
+version: 2
+
+jobs:
+  build:
+    docker:
+      - image: tweag/linear-types:0.1.6
+    environment:
+      - stack: stack --no-docker
```
mboes: I don't find the factoring out of `stack --no-docker` helpful. Just write `stack --no-docker` inline everywhere. Overzealous application of DRY sometimes has readability cost through obfuscation while bringing no benefit.

.circleci/config.yml
```diff
@@ -0,0 +1,29 @@
...
+          key: linear-base-stackcache-{{ arch }}-v171204-
+      - run:
+          # --jobs=1 because of limitted memory
+          name: Build dependencies
+          command: |
+            $stack build --jobs=1 --only-dependencies --no-haddock linear-base:lib
+            $stack build --jobs=1 --test --bench --only-snapshot
+            find . -name "*.cabal" -o -name "stack.yaml" -type f | xargs sha256sum > /tmp/dps_checksum
+      - save_cache:
+          key: linear-base-stackcache-{{ arch }}-v171204-{{ checksum "/tmp/dps_checksum" }}
```
mboes: What is the date for?

src/Prelude/Linear.hs
aspiwack: It's just something I noted during this PR. As part of the warning hunting bit.

.circleci/config.yml
aspiwack: To be able to invalidate the cache.

src/System/IO/Resource.hs
aspiwack: Because of `RebindableSyntax` shenanigans.

.circleci/config.yml
aspiwack: Really? Because I find it positively obnoxious to write `--no-docker` everywhere. It is not part of the logic of the test, but of the overall architecture of the CI, so the repetition is obfuscation, in my view.

aspiwack:
Dependencies are not cached because they are not part of the snapshot (and I'm only caching the snapshot). I'll try and piggyback on an LTS. If that doesn't work, I'll go for the new custom snapshot feature. Though, I'll do it after merge.

.circleci/config.yml
mboes: Just change it. It's what we do in virtually every other CircleCI deployment we have (usually concerning the `--nix` argument).

src/System/IO/Resource.hs
mboes: Can you clarify? At any rate - add a comment. When `RebindableSyntax` goes away we'll remember to reenable this warning that way.

aspiwack:
Anything else?

aspiwack:
@mboes anything else?

PR #8, opened 2017-12-20
First prototype of off-heap data management

aspiwack:
Depends on #7 .

This only ships a few tests. I will need to spend a bit of time devising a functor/monad hierarchy before I can write more. Otherwise managing the `Pool` is a pain.

There is no exception management yet. Though the path is explained. I'd rather get a few more tests and example before this.

So it's atomic enough, if not definitive. More soon.

### DIFF
--- a/src/Data/OffHeap.hs
+++ b/src/Data/OffHeap.hs
@@ -1,3 +1,6 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
 -- | This module introduces primitives to /safely/ store data off-heap. The
 -- benefit of off-heap data is that it does not add to the GC pressure, and help
 -- ensure predictable latency (/e.g./ in distributed applications). The cost is
@@ -12,7 +15,7 @@
 -- GC. Despite calling @malloc@ and @free@ under the hood, the entire API is
 -- pure, and does not make calls in IO.
 --
--- The allocation API is organised around a notion of memory pool. From the API
+-- The allocation API is organised around a notion of memory 'Pool'. From the API
 -- point of view, a pool serves as a source of linearity. That is: it ensures
 -- that the allocation primitive need not take a continuation to delimit its
 -- lifetime. Besides convenience, it avoids needlessly preventing functions from
@@ -20,7 +23,7 @@
 --
 -- Pools play another role: resilience to exceptions. If an exception is raised,
 -- all the data in the pool is deallocated. It does not, however, impose a stack
--- discipline: data in pools is normally freed by calls to the 'free' function,
+-- discipline: data in pools is normally freed by the destruction primitives of ,
 -- only in case of exception are the pool deallocated in a stack-like
 -- manner. Moreover, pool A can have data pointing to pool B, while at the same
 -- time, pool B having data pointing to pool A.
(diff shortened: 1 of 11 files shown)


aspiwack:
Thoughts from few minutes later.

The cost of this `Pool` thing in state-passing style is that allocations are unnecessarily sequentialised, and that lists are not functors or traversables.

The pool is necessary, but we can alleviate some of the pain by changing the state-passing style into an environment-passing style with `Pool` splitting (in the style of random). We can implement that by making `Pool` `Dupable`(pretty sure it's correct), which would nicely complete the `Consumable` that we need anyway.

Off-heap containers would still not be functors or traversables. But almost: we can define a _traversal_ taking an additional `Pool` as an argument. Overall seems more practical. And would no require monadic combinator to reduce the boilerplate.

aspiwack:
Just did the above. It is much more usable an interface. And no need for a monadic interface.

aspiwack:
@m Any comment before I merge? I'll commit exception-resilience in another PR.

stack.yaml
```diff
@@ -1,6 +1,6 @@
 # Configures GHC to be the 8.2 variant from the Docker image
 # See https://hub.docker.com/r/tweag/linear-types/
-resolver: ghc-8.2
+resolver: nightly-2017-12-20
```
mboes: Can use LTS-10 now.

aspiwack: Ha, I merged this in a separate PR before you brought it up. Well… no worries, I'll fix it in the afternoon.

src/Data/OffHeap.hs
```diff
@@ -0,0 +1,117 @@
+{-# LANGUAGE GADTs #-}
```
mboes: I think `Data.OffHeap` is not a good name. I either doesn't mean anything to people, or is at best confusing because off-heap data is still on-heap, it's just not on on-GC-heap. We have an existing `Foreign.*` hierarchy, which among other things already provides `malloc`. This module is essentially a different API for `malloc`, so let's insert ourselves into the same hierarchy!

Call it e.g. `Foreign.Marshal.Alloc.Linear`. Now, you could argue that this is a different beast because it's pure. But I think this module should *not* be pure, making that point moot. At any rate, you cannot use the `Storable` type class if you're going to make this interface pure. `Storable` instances make no guarantees about referential transparency. Because methods are in `IO`, they can have arbitrary effects and that would still be fine. In fact `Storable` has no laws.

Also, I'm not sure you need the `Pool` concept anymore if you stick to `(R)IO`.

src/Data/OffHeap.hs
```diff
@@ -0,0 +1,117 @@
...
+
+instance Consumable Pool where
+  consume Pool = ()
+
+instance Dupable Pool where
+  dup Pool = (Pool, Pool)
+
+-- XXX: this indirection is possibly not necessary. It's here because the inner
+-- Ptr must be unrestricted (in order to implement deconstruct at the moment).
+data Box a where
```
mboes: Why the non-descript `Box` name rather than just using `Unrestricted` here? Just because `Box a` shorter than `Unrestricted (Ptr a)`? Or as an abstraction barrier (hide the `Ptr`)?

src/Data/OffHeap.hs
```diff
@@ -0,0 +1,117 @@
...
+alloc :: forall a. Storable a => a ->. Pool ->. Box a
+alloc a Pool =
+    Unsafe.toLinear mkPtr a
+  where
+    mkPtr :: a -> Box a
+    mkPtr a' = unsafeDupablePerformIO $ do
+      ptr <- malloc
+      poke ptr a'
+      return (Box ptr)
+
```
mboes: extraneous whitespace.

src/Data/OffHeap.hs
aspiwack: It's an abstraction barrier around `Ptr`. It differs from `Ptr` in that it is pure. Also it will hold some more data for exception tolerance.

aspiwack:
@mboes please take another look (I'm done with exception management as well, I'll upload on a different PR for easy review).

PR #9, opened 2017-12-22
Manually managed data: make exception safe

aspiwack:
_Depends on #8 . There is only one commit in this PR._

Memory is freed in case of exception. A list of pointers is kept in a
non-gc'd region in order not to add to the GC pressure, and not
contribute to less predictable latencies.


aspiwack:
@mboes just rebased on top of master.

PR #10, opened 2018-01-26
A linear monad hierarchy

aspiwack:
This PR adds a linear (enriched) monad hierarchy.

Highlights:

- The linear monad hiearchy is in its own module, in order to reuse the same notations as regular monads, in particular the infix notations. Defining instance feels quite native this way.
- I put the entire functor/applicative/monad/monad-fail hierarchy in the same module (`Control.Monad.Linear`). This is unlike the usual hierarchy which has one module per type-class. We could want to instead follow the same module layout as `base` (except that this kind of enriched functors should be in `Control.Functor.Linear`, `Data.Functor.Linear` will have a different type). I decided this way because they are generally used together (an alternative is to define them in separate files and re-export the preceding modules)
- This uses F#-style builders to benefit from the `do` notation (it's a bit awkward, but works). Due to how `-XRebindableSyntax` works the builder is defined in a distinguished module `Control.Monad.Linear.Builder`: the projections from the builder must also be named `(>>=)` and so one. It would conflict with the definition of `Monad` if defined in the same place.
- For `-XRebindableSyntax` to work well, `Monad(..)` is excluded from `Prelude.Linear`
- A builder for regular monads is provided in `Control.Monad.Builder`.

### DIFF
--- a/examples/Spec.hs
+++ b/examples/Spec.hs
@@ -9,10 +9,11 @@ import Control.Monad (void)
 import Data.Typeable
 -- TODO: restore (see #18)
 -- import qualified Foreign.Heap as Heap
-import qualified Foreign.List as List
 import Foreign.List (List)
-import qualified Foreign.Marshal.Pure as Manual
+import qualified Foreign.List as List
 import Foreign.Marshal.Pure (Pool)
+import qualified Foreign.Marshal.Pure as Manual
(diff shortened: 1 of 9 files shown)


aspiwack:
I'm merging this, as I'm increasingly needing Monads everywhere.

PR #14, opened 2018-07-03
η-expand a linear function used as an unrestricted one

aspiwack:
It is necessary with the new specification which doesn't auto η-expand anymore.


mpickering:
This should fix it.

aspiwack:
There may be others. So I may need to fix more after tweag/ghc#92 is merged. But let's start with this.

PR #15, opened 2018-07-04
"Fix" an assertion failure caused by unsafeCoerce

mpickering:
I don't understand how this code is meant to work. I haven't tried to
run any programs.


aspiwack:
It just makes a linear version of `unsafeCoerce`. I wonder why this fails.

mpickering:
It caused an assertion to fail. I didn't look any deeper than that as I wasn't sure what unsafeCoercing unsafeCoerce would do operationally.

aspiwack:
I think I managed to bypass this somehow. But there are still related bugs in the linter. Closing for now.

PR #26, opened 2018-12-14
Vectors of known length & refreshed Dupable

aspiwack:
This comes in three part

- A data-functor hierarchy representing containers (especially a `Data.Applicative` which represents zippable things)
- Vectors of known length, which are `Data.Applicative`
- An `Dupable`'s one method now has type `KnownNat n => a ->. V n a`.

Here is the rationale

> One immediate benefit is that the instances are much easier to write, using the applicative instance (of course, it would have worked just fine, if that was the only goal, to make a V2 applicative functor).
>
> The other benefit is that there was no pleasant way to produce more than two things from a dupable. It is possible to write a function `(a ->. (a,a)) ->. a ->. V n a`, but at this point we already have `V n` defined. And producing a `V n a` directly is, I expect, significantly more efficient.
>
> Another approach would be to write a function `(a ->. (a,a)) ->. Int ->. a ->. [a]`, and rely on incomplete pattern-matchings. But this is rather yucky.

### DIFF
--- a/src/Data/Vector/Linear.hs
+++ b/src/Data/Vector/Linear.hs
@@ -28,7 +28,7 @@
 
 module Data.Vector.Linear
   ( V
-  , Elim
+  , FunN
   , elim
   , make
   ) where
@@ -61,7 +61,7 @@ newtype V (n :: Nat) (a :: *) = V (Vector a)
   -- offers many convenience function. Since all these unsafeCoerces probably
   -- kill the fusion rules, it may be worth it going lower level since I
   -- probably have to write my own fusion anyway. Therefore, starting from
-  -- Arrays at the moment.
+  -- Vectors at the moment.
 
 theLength :: forall n. KnownNat n => Int
 theLength = fromIntegral (natVal' @n (proxy# @_))
@@ -79,9 +79,9 @@ instance KnownNat n => Data.Traversable (V n) where
     (V . Unsafe.toLinear (Vector.fromListN (theLength @n))) Control.<$>
     Data.traverse f (Unsafe.toLinear Vector.toList xs)
 
-type family Elim (n :: Nat) (a :: *) (b :: *) :: * where
-  Elim 0 a b = b
-  Elim n a b = a ->. Elim (n-1) a b
+type family FunN (n :: Nat) (a :: *) (b :: *) :: * where
+  FunN 0 a b = b
+  FunN n a b = a ->. FunN (n-1) a b
 
 split :: 1 <= n => V n a ->. (# a, V (n-1) a #)
 split = Unsafe.toLinear split'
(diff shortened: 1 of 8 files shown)


src/Data/Vector/Linear.hs
```diff
@@ -0,0 +1,128 @@
...
+import Prelude.Linear.Internal.Simple
+import qualified Unsafe.Linear as Unsafe
+
+newtype V (n :: Nat) (a :: *) = V (Vector a)
+  deriving (Eq, Ord, Prelude.Functor)
+  -- Using vector rather than, say, 'Array' (or directly 'Array#') because it
+  -- offers many convenience function. Since all these unsafeCoerces probably
+  -- kill the fusion rules, it may be worth it going lower level since I
+  -- probably have to write my own fusion anyway. Therefore, starting from
+  -- Arrays at the moment.
```
b-mehta: Should this say "starting from Vectors at the moment"?

src/Prelude/Linear/Internal/Simple.hs
```diff
@@ -0,0 +1,55 @@
...
+curry f x y = f (x, y)
+
+-- | Beware, 'uncurry' is not compatible with the standard one because it is
+-- higher-order and we don't have multiplicity polymorphism yet.
+uncurry :: (a ->. b ->. c) ->. (a, b) ->. c
+uncurry f (x,y) = f x y
+
+-- | Beware: @(.)@ is not compatible with the standard one because it is
+-- higher-order and we don't have multiplicity polymorphism yet.
+(.) :: (b ->. c) -> (a ->. b) ->. a ->. c
```
b-mehta: I believe this type signature should be
```(b ->. c) ->. (a ->. b) ->. a ->. c```

**Comment by aspiwack** [comment]
You are right. There needs a `make`. But I'd give it type `KnownNat n=> Elim n a (Vec n a)`. `Elim` becomes a terrible name though :) .

**Comment by b-mehta** [comment]
```
build :: forall n a. KnownNat n => Elim n a (V n a)
build = case caseNat @n of
          Left Refl -> V Vector.empty
          Right Refl -> Unsafe.coerce prepend :: Elim n a (V n a)
            where prepend :: a -> Elim (n-1) a (V n a)
                  prepend t = case predNat @n of
                                Dict -> continue @(n-1) @a @(V (n-1) a) (cons t) (build @(n-1) @a)

cons :: forall n a. a ->. V (n-1) a ->. V n a
cons = Unsafe.toLinear $ \x -> Unsafe.coerce (Vector.cons x)

continue :: forall n a b c. KnownNat n => (b ->. c) ->. Elim n a b ->. Elim n a c
continue = case caseNat @n of
             Left Refl -> id
             Right Refl -> Unsafe.coerce continueS
               where continueS :: (KnownNat n, 1 <= n) => (b ->. c) ->. (a ->. Elim (n-1) a b) ->. (a ->. Elim (n-1) a c)
                     continueS f' x a = case predNat @n of
                                          Dict -> continue @(n-1) @a @b f' (x a)
```
Here was my implementation of that function

File: src/Data/Vector/Linear.hs (hunk 1 above)
Comment (aspiwack) [reply src/Data/Vector/Linear.hs]: Yes.

File: src/Data/Vector/Linear.hs [hunk 3]
```diff
@@ -0,0 +1,149 @@
...
+  pure a = V $ Vector.replicate (theLength @n) a
+  (V fs) <*> (V xs) = V $
+    Unsafe.toLinear2 (Vector.zipWith @(_ ->. _) (\f x -> f x)) fs xs
+
+instance KnownNat n => Data.Traversable (V n) where
+  traverse f (V xs) =
+    (V . Unsafe.toLinear (Vector.fromListN (theLength @n))) Control.<$>
+    Data.traverse f (Unsafe.toLinear Vector.toList xs)
+
+type family Elim (n :: Nat) (a :: *) (b :: *) :: * where
```
aspiwack: Let's find a better name for this family. Like `FunN` or something.

PR #28, opened 2019-07-19
Define Applicatives by `liftA2` and add default (>>)

b-mehta:
Added some minimal pragmas as in standard base to allow alternate definition of Applicative. Added default definition of `(>>)`.

### DIFF
--- a/src/Data/Functor/Linear.hs
+++ b/src/Data/Functor/Linear.hs
@@ -14,6 +14,7 @@
 
 module Data.Functor.Linear where
 
+import Prelude.Linear.Internal.Simple
 import qualified Control.Monad.Linear as Control
 
 class Functor f where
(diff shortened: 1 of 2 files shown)


src/Data/Functor/Linear.hs
```diff
@@ -28,7 +29,7 @@ class Functor f where
 -- drops values, which we are not allowed to do in a linear container).
 --
 -- In fact, an applicative functor is precisely a functor equipped with (pure
--- and) @zipWith :: (a ->. b ->. c) -> f a ->. f b ->. f c@
+-- and) @liftA2 :: (a ->. b ->. c) -> f a ->. f b ->. f c@
```
aspiwack: Well, the reference to `zipWith` was rather on purpose. We probably want to say that `liftA2` has the same signature as `zipWith` does for lists or something.

aspiwack:
Mmm… I see that there is no CI running. I'll have to remedy that.

Also: don't forget to delete branch after merging.

b-mehta:
It looks like CI did run here: https://circleci.com/gh/tweag/linear-base/107, but it wasn't reported for some reason.

b-mehta:
It turns out github status updates were off for this repo, I've turned them back on and CircleCI now reports the results of checks.

PR #30, opened 2019-07-19
Traversable changes

b-mehta:
Modify Data.Traversable to use Data.Applicatives and use this to write cleaner `Dupable a => Dupable [a]` and same for `Movable`.

### DIFF
--- a/src/Data/Functor/Linear.hs
+++ b/src/Data/Functor/Linear.hs
@@ -15,7 +15,6 @@
 module Data.Functor.Linear where
 
 import Prelude.Linear.Internal.Simple
-import qualified Control.Monad.Linear as Control
 
 class Functor f where
   fmap :: (a ->. b) -> f a ->. f b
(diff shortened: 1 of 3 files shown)


src/Data/Functor/Linear.hs
```diff
@@ -50,9 +49,17 @@ class Functor f => Applicative f where
   f <*> x = liftA2 ($) f x
   liftA2 :: (a ->. b ->. c) -> f a ->. f b ->. f c
   liftA2 f x y = f <$> x <*> y
+-- TODO: Add monad and monad fail
+-- (>>=) :: m a ->. (a ->. m b) -> m b
```
aspiwack: Data Monad is not very interesting, as far as I can tell (as do notation doesn't work). Let's not rush there until we have an actual use-case.

b-mehta:
I'll try to add these relations in my next PR!

Issue #27, opened 2018-12-20
Polymorphism big plans for Data and Control functor hierarchy


aspiwack:
Currently, data functors and control functors have type

```haskell
fmap :: (a ->. b) -> f a ->. f b
```

and

```haskell
fmap :: (a ->. b) ->. f a ->. f b
```

respectively. (that is data functors are regular functors and control functors are enriched functors).

But with polymorphism, it seem that we may want to split them further. Based more on how they are used (data functors are data containers of sorts, which control functors are effect-bearing computations).

My thoughts are the following: just like for lists the type of data functor's `fmap` would become

```haskell
fmap :: (a # p -> b) -> f a # p -> f b
```

(it specialises to linear data functor when `p ~ One` and to regular (unrestricted) functor when `p ~ Omega`). It is not equivalent, but it seems to make sense. The only downside, though, is that it may be a bit weird for types which are always linear in the context (like a pure mutable array, for instance). An alternative is to do:

```haskell
class Functor (p :: Multiplicity) f where
  fmap :: (a # p -> b) -> f a # p -> f b
```

This way we have polymorphism without the weirdness.

On the other hand, for control functors, we care about linearity of the control, but not linearity of the data. Therefore, we can the multiplicity of the data variable, like `IO` in the paper.

```haskell
class Functor (f :: Multiplicity -> * -> *) where
  fmap :: (a ->. b) ->. f p a ->. f p b
```
(maybe a variant of type `(a # p ->. b) ->. f q a ->. f (p * b)`? Really the general form would be `(a #p ->. Mult q b) ->. f p a ->. f q b`).

The bottom-line is that control functors become functors from the category where objects are pairs `(a, p)` of an object and a type, and maps from `(a, p)` to `(b, q)` are `a # p ->. Mult q b`, to the category of types and linear functions (enriched in the category of types and linear functions).

This is not a restriction because you can always compose your old-style control functor with `Mult p` to obtain a new one. It also extends to a notion of monad (more precisely, of relative monad).

Which is the whole point of the change: it will make do notation much more pleasant, with the ability to return linear or unrestricted values.

sjoerdvisscher:
I think for data functors the domain and codomain should be allowed to differ? So:

```haskell
fmap :: (a %p -> b) -> f a %q -> f b
```

PR #31, opened 2019-07-22
WIP: Prelude Functor => Data Functor constraint

b-mehta:
First brush at a (temporary) functor hierarchy.
(WIP: does not yet compile)

Prelude.Functor is now a superclass of Data.Functor. This is temporary, until multiplicity polymorphism is added, when the type of `Data.fmap` includes the type of `Prelude.fmap`. 
Until then, to `fmap` a non-linear arrow we must use `Prelude.fmap`, so it is now a required constraint.

Data.Functor is now a superclass of Control.Functor, similarly for Applicative.

### DIFF
--- a/src/Control/Monad/Linear.hs
+++ b/src/Control/Monad/Linear.hs
@@ -33,14 +33,14 @@ import Prelude (String)
 -- TODO: make the laws explicit
 
 -- | Enriched linear functors.
-class Functor f where
+class Data.Functor f => Functor f where
   fmap :: (a ->. b) ->. f a ->. f b
 
 (<$>) :: Functor f => (a ->. b) ->. f a ->. f b
 (<$>) = fmap
 
 -- | Enriched linear applicative functors
-class Functor f => Applicative f where
+class (Data.Applicative f, Functor f) => Applicative f where
   {-# MINIMAL pure, ((<*>) | liftA2) #-}
   pure :: a ->. f a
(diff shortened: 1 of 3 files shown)


b-mehta:
One advantage of this change is that it allows us to define `Data.Traversable Unrestricted`, so we have 
```sequence @Unrestricted :: Applicative e => Unrestricted (e a) ->. e (Unrestricted a)```.

However, to define a `Prelude.Functor IO` instance, it seems like we need to have multiplicity-parametrised `IORes`...

```newtype IOL p a = IOL {unIOL :: World ->. IORes p a}
data IORes p a where
IOR :: World ->. a →_p IORes p a```

b-mehta:
On second thoughts, a `Prelude.Functor f => Data.Functor f` constraint probably isn't the best way to go about this, since the intended plan in #27 is to have `Functor One f` and `Functor Omega f` as independent constraints, which correspond to the current `Data.Functor f` and `Prelude.Functor f`.

aspiwack:
Yeah, I don't think this is a good idea. If you need both constraint, just require them both at the site which needs both constraints.

Note that this PR is incompatible with making `Data.Functor` a superclass of `Control.Functor`. Because many `Control.Functor` are, in fact, not `Prelude.Functor` (any which has a linear arrow in its definition, basically, such as `State`, and `Reader`).

b-mehta:
Right, I'll close this.

PR #33, opened 2019-07-25
Superclass relations

b-mehta:
Add superclasses Data.Functor => Control.Functor and similarly for Applicative; along with default definitions to conveniently define such instances.

### DIFF
--- a/src/Control/Monad/Linear.hs
+++ b/src/Control/Monad/Linear.hs
@@ -16,6 +16,7 @@ module Control.Monad.Linear
   , return
   , join
   , ap
+  , DataFromControl(..)
   ) where
 
 import Prelude.Linear.Internal.Simple (id)
@@ -79,3 +80,13 @@ join action = action >>= id
 -- | Convenience operator to define Applicative instances in terms of Monad
 ap :: Monad m => m (a ->. b) ->. m a ->. m b
 ap f x = f >>= (\f' -> fmap f' x)
+
+-- | DerivingVia combinators for Data.XXX in terms of Control.XXX
+newtype DataFromControl f a = DataFromControl (f a)
+
+instance Functor f => Data.Functor (DataFromControl f) where
+  fmap f (DataFromControl x) = DataFromControl (fmap f x)
+
+instance Applicative f => Data.Applicative (DataFromControl f) where
+  pure x = DataFromControl (pure x)
+  DataFromControl f <*> DataFromControl x = DataFromControl (f <*> x)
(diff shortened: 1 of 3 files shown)


src/Control/Monad/Linear.hs
```diff
@@ -34,21 +37,27 @@ import Prelude (String)
...
+class (Data.Applicative f, Functor f) => Applicative f where
   {-# MINIMAL pure, ((<*>) | liftA2) #-}
   pure :: a ->. f a
   (<*>) :: f (a ->. b) ->. f a ->. f b
   (<*>) = liftA2 id
   liftA2 :: (a ->. b ->. c) ->. f a ->. f b ->. f c
   liftA2 f x y = f <$> x <*> y
 
+dataPureDefault :: Applicative f => a -> f a
+dataPureDefault x = pure x
```
aspiwack: Maybe we could also straight up give `DerivingVia` combinator for these.

Like, define a data type

```haskell
newtype Forget f a = Forget (f a)
```

Then prove

```haskell
instance Functor f => Data.Forget (Control f)
instance Applicative f => Data.Applicative (Control f)
```

Then we can use it in `deriving` clauses.

b-mehta: Do you mean 
```haskell
instance Functor f => Data.Functor (Forget f)
instance Applicative f => Data.Applicative (Forget f)
```

aspiwack: Yes. Let's blame it on the heat.

b-mehta: I've added this, with a different name - since the name is exported I felt something more evocative would be nicer.

b-mehta: I wonder if it is worth adding 
```
Linear.coerce :: Coercible a b => a ->. b
```
also.

src/Control/Monad/Linear.hs
```diff
@@ -79,3 +80,13 @@ join action = action >>= id
 -- | Convenience operator to define Applicative instances in terms of Monad
 ap :: Monad m => m (a ->. b) ->. m a ->. m b
 ap f x = f >>= (\f' -> fmap f' x)
+
+-- | DerivingVia combinators for Data.XXX in terms of Control.XXX
+newtype DataFromControl f a = DataFromControl (f a)
```
aspiwack: It's a bit wordy as it's supposed to be imported qualified. So how about `Control.Data` instead?

b-mehta: Ah good point!

PR #35, opened 2019-07-30
Merge profunctors and very basic lenses

b-mehta:


src/Control/Optics/Linear/Internal.hs
```diff
@@ -0,0 +1,53 @@
...
+{-# LANGUAGE TypeOperators #-}
+
+module Control.Optics.Linear.Internal where
+
+import qualified Data.Profunctor.Linear as Monoidal
+import Data.Profunctor.Linear hiding (swap)
+import Data.Void
+import Prelude.Linear
+
+type Optic_ arr a b s t = a `arr` b -> s `arr` t
```
aspiwack: We said we wanted a `newtype` for that.

src/Control/Optics/Linear/Internal.hs
```diff
@@ -0,0 +1,53 @@
...
+type Lens a b s t = Optic (Strong (,) ()) a b s t
+type Lens' a s = Lens a a s s
+type Prism a b s t = Optic (Strong Either Void) a b s t
+type Prism' a s = Prism a a s s
+
+-- | Because it is inconvenient to pair up parametric constraints, 'Traversal'
+-- is defined in terms of 'Optic_' rather than 'Optic'.
+type Traversal a b s t =
+  forall arr.
+  ( Strong (,) () arr , Strong Either Void arr , Monoidal (,) () arr)
```
aspiwack: I'm not sure this is the right definition. This is something I was experimenting with, but it may not work.

In the lens-as-yoneda paper, there is a suggested type class, a similar one is used in the [mezzolens](https://hackage.haskell.org/package/mezzolens) library (`Wandering` in [Mezzolens.Profunctor](https://hackage.haskell.org/package/mezzolens-0.0.0/docs/Mezzolens-Profunctor.html)). It's probably what we should use too.

src/Control/Optics/Linear/Internal.hs
```diff
@@ -0,0 +1,53 @@
...
+_2 = second
+
+_Left :: Prism' a (Either a b)
+_Left = first
+
+_Right :: Prism' b (Either a b)
+_Right = second
+
+-- traversed :: Traversable t => Traversal a b (t a) (t b)
+-- traversed p = _
```
aspiwack: Dead code.

aspiwack:
I should add that the history ought to be cleaned before merging.

b-mehta:
Subsumed by #37 and #39.

PR #37, opened 2019-07-30
Profunctors

b-mehta:
Basic profunctor classes

Replaces part of #35, with a cleaner history. Only includes profunctor classes and no lenses.

### DIFF
--- a/src/Data/Bifunctor/Linear.hs
+++ b/src/Data/Bifunctor/Linear.hs
@@ -1,21 +1,54 @@
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE TypeOperators #-}
 
 module Data.Bifunctor.Linear where
 
+import Prelude.Linear
 import Data.Void
 
--- TODO: Is a bifunctor
 -- TODO: associators
 -- TODO: remove `swap` from Prelude
+
+-- | Bifunctors on the category of linear types
+--
+-- Laws:
+-- If 'bimap' is supplied, then
+-- * @'bimap' 'id' 'id' = 'id'@
+-- If 'first' and 'second' are supplied, then
+-- * @
+-- 'first' 'id' ≡ 'id'
+-- 'second' 'id' ≡ 'id'
+-- @
+-- If all are supplied, then
+-- * @'bimap' f g = 'first' f '.' 'second' g
+class Bifunctor p where
+  {-# MINIMAL bimap | (first , second) #-}
+  bimap :: (a ->. b) -> (c ->. d) -> a `p` c ->. b `p` d
+  bimap f g = first f . second g
+
+  first :: (a ->. b) -> a `p` c ->. b `p` c
+  first f = bimap f id
+
+  second :: (b ->. c) -> a `p` b ->. a `p` c
+  second = bimap id
+
+instance Bifunctor (,) where
+  bimap f g (x,y) = (f x, g y)
+  first f (x,y) = (f x, y)
+  second g (x,y) = (x, g y)
+
+instance Bifunctor Either where
+  bimap f _ (Left x) = Left (f x)
+  bimap _ g (Right y) = Right (g y)
 
 -- | Symmetric monoidal products on the category of linear types
 --
 -- Laws
--- * @swap . swap = id@
-class SymmetricMonoidal (m :: * -> * -> *) (u :: *) | m -> u, u -> m where
+-- * @'swap' . 'swap' = 'id'@
+class Bifunctor m => SymmetricMonoidal (m :: * -> * -> *) (u :: *) | m -> u, u -> m where
   swap :: a `m` b ->. b `m` a
 
 instance SymmetricMonoidal (,) () where
--- a/src/Data/Bifunctor/Linear.hs
+++ b/src/Data/Bifunctor/Linear.hs
@@ -25,12 +25,15 @@ class Bifunctor p where
   {-# MINIMAL bimap | (first , second) #-}
   bimap :: (a ->. b) -> (c ->. d) -> a `p` c ->. b `p` d
   bimap f g = first f . second g
+  {-# INLINE bimap #-}
 
   first :: (a ->. b) -> a `p` c ->. b `p` c
   first f = bimap f id
+  {-# INLINE first #-}
 
   second :: (b ->. c) -> a `p` b ->. a `p` c
   second = bimap id
+  {-# INLINE second #-}
 
 instance Bifunctor (,) where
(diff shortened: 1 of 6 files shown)


src/Data/Profunctor/Linear.hs
```diff
@@ -0,0 +1,68 @@
...
+
+  lmap :: (s ->. a) -> a `arr` t -> s `arr` t
+  lmap f = dimap f id
+  {-# INLINE lmap #-}
+
+  rmap :: (b ->. t) -> s `arr` b -> s `arr` t
+  rmap = dimap id
+  {-# INLINE rmap #-}
+
+-- TODO: Move to a dedicated module
```
aspiwack: Let's make a `Data.Bifunctor.Linear` module. Add a (linear) `Bifunctor` type class there. And make `SymmetricMonoidal` a super class.

src/Data/Profunctor/Linear.hs
```diff
@@ -0,0 +1,68 @@
...
+-- TODO: Is a bifunctor
```
aspiwack: Scrap and see above.

src/Data/Profunctor/Linear.hs
```diff
@@ -0,0 +1,68 @@
...
+-- TODO: associators
```
aspiwack: Let's add these for completeness.

src/Data/Profunctor/Linear.hs
```diff
@@ -0,0 +1,68 @@
...
+-- TODO: remove `swap` from Prelude
```
aspiwack: And remove `swap` from `Prelude.Linear` indeed.

src/Data/Profunctor/Linear.hs
```diff
@@ -0,0 +1,68 @@
...
+  swap :: a `m` b ->. b `m` a
+
+instance SymmetricMonoidal (,) () where
+  swap (x, y) = (y, x)
+
+instance SymmetricMonoidal Either Void where
+  swap (Left x) = Right x
+  swap (Right x) = Left x
+
+class (SymmetricMonoidal m u, Profunctor arr) => Monoidal m u arr where
```
aspiwack: This is where the moving stop: `Monoidal` profunctors is a type class for the `Profunctor` module.

src/Data/Bifunctor/Linear.hs
```diff
@@ -0,0 +1,59 @@
...
+-- If 'first' and 'second' are supplied, then
+-- * @
+-- 'first' 'id' ≡ 'id'
+-- 'second' 'id' ≡ 'id'
+-- @
+-- If all are supplied, then
+-- * @'bimap' f g = 'first' f '.' 'second' g
+class Bifunctor p where
+  {-# MINIMAL bimap | (first , second) #-}
+  bimap :: (a ->. b) -> (c ->. d) -> a `p` c ->. b `p` d
```
b-mehta: I wonder now if the final arrow here should actually be linear, comparing with the type

```haskell
dimap :: (s ->. a) -> (b ->. t) -> a `arr` b -> s `arr` t
```
in which the final arrow is non-linear.

Consistency feels like the correct choice, but I'm not sure if both should be linear or both non-linear.

b-mehta: I think linear is the correct choice now. In particular, asking for 
```haskell
dimap :: (s ->. a) -> (b ->. t) -> a `arr` b ->. s `arr` t
```
would mean `rmap` witnesses that a `Profunctor` is a `Data.Functor` in its second argument, which feels more correct.

aspiwack: Ah, `Bifunctor` has the `Data` vs `Control` issue again.

Profunctors are not endofunctors (they are functors to Set). So they don't have (quite) the same issue. `dimap` is supposed to be a non-linear arrow at the end in theory, and quite possibly, that would include more potential profunctors, which we may need (I'd say probably, even) to define `set` and `view` (these are not linear operations, but they still work when the data is not linear).

Looking concretely, we are considering `(,)` and `Either`, `Either` is not a Control bifunctor. So I'd say we can restrict our attention to Data Bifunctors for the time being. Hence `bimap :: (a ->. b) -> (c ->. d) -> a `p` c -> b `p` d`.

src/Data/Bifunctor/Linear.hs
```diff
@@ -44,16 +41,27 @@ instance Bifunctor Either where
...
 -- | Symmetric monoidal products on the category of linear types
 --
 -- Laws
 -- * @'swap' . 'swap' = 'id'@
 class Bifunctor m => SymmetricMonoidal (m :: * -> * -> *) (u :: *) | m -> u, u -> m where
+  {-# MINIMAL swap, (assoc | unassoc) #-}
+  assoc :: (a `m` b) `m` c ->. a `m` (b `m` c)
+  assoc = swap . unassoc . swap . unassoc . swap
+  unassoc :: a `m` (b `m` c) ->. (a `m` b) `m` c
+  unassoc = swap . assoc . swap . assoc . swap
```
aspiwack: I'd call them `rassoc` and `lassoc`, personally.

src/Data/Bifunctor/Linear.hs
b-mehta: I don't think this is the `Data` vs `Control` issue, since the `(a ->. b)` is not necessarily consumed linearly. Concretely, there's no issue making `Either` a `Bifunctor` with the `bimap` I gave above (as witnessed by the CI check passing).

src/Data/Bifunctor/Linear.hs
```diff
@@ -23,14 +23,17 @@ import Data.Void
 -- * @'bimap' f g = 'first' f '.' 'second' g
 class Bifunctor p where
   {-# MINIMAL bimap | (first , second) #-}
-  bimap :: (a ->. b) -> (c ->. d) -> a `p` c ->. b `p` d
-  bimap f g = first f . second g
+  bimap :: (a ->. b) -> (c ->. d) -> a `p` c -> b `p` d
```
aspiwack: Mmm wouldn't these be

```haskell
   bimap :: (a ->. b) -> (c ->. d) -> a `p` c -> b `p` d
```

I'd even expect `forall a. (Functor p a)` to be a super-constraint. But this last point can wait until another PR.

b-mehta: I agree completely - this is what started the discussion above - I changed it back after your comment: 
> Looking concretely, we are considering `(,)` and `Either`, `Either` is not a Control bifunctor. So I'd say we can restrict our attention to Data Bifunctors for the time being. Hence 
```haskell
bimap :: (a ->. b) -> (c ->. d) -> a `p` c -> b`p` d
```

aspiwack: I think that is was a typo on my part :slightly_frowning_face: . Quite unfortunate.

PR #38, opened 2019-07-31
Use either combinator to simplify instances

b-mehta:
Use the linear `either` consumer to have nicer `Either` instances

### DIFF
--- a/src/Data/Bifunctor/Linear.hs
+++ b/src/Data/Bifunctor/Linear.hs
@@ -41,8 +41,7 @@ instance Bifunctor (,) where
   second g (x,y) = (x, g y)
 
 instance Bifunctor Either where
-  bimap f _ (Left x) = Left (f x)
-  bimap _ g (Right y) = Right (g y)
+  bimap f g = either (Left . f) (Right . g)
 
 -- TODO: assoc laws
 -- | Symmetric monoidal products on the category of linear types
@@ -63,8 +62,7 @@ instance SymmetricMonoidal (,) () where
   rassoc ((x,y),z) = (x,(y,z))
 
 instance SymmetricMonoidal Either Void where
-  swap (Left x) = Right x
-  swap (Right x) = Left x
+  swap = either Right Left
   rassoc (Left (Left x)) = Left x
   rassoc (Left (Right x)) = Right (Left x)
   rassoc (Right x) = Right (Right x)


src/Data/Bifunctor/Linear.hs
```diff
@@ -63,8 +62,5 @@ instance SymmetricMonoidal (,) () where
   rassoc ((x,y),z) = (x,(y,z))
 
 instance SymmetricMonoidal Either Void where
-  swap (Left x) = Right x
-  swap (Right x) = Left x
-  rassoc (Left (Left x)) = Left x
-  rassoc (Left (Right x)) = Right (Left x)
-  rassoc (Right x) = Right (Right x)
+  swap = either Right Left
+  rassoc = either (either Left (Right . Left)) (Right . Right)
```
aspiwack: I don't really like this `rassoc` definition. It's terse, but it's hard to read. The previous one is clearer in my eyes.

The other definitions, on the other hand, are convincing!

b-mehta: Ahh, something frustrating about this is that the older one doesn't compile in ghc-8.9.0.20190613, but the new one does...

aspiwack: I'm confused. What is the problem?

b-mehta: If I try to compile
```haskell
instance SymmetricMonoidal Either Void where
  swap = either Right Left
  rassoc (Left (Left x)) = Left x
  rassoc (Left (Right x)) = Right (Left x)
  rassoc (Right x) = Right (Right x)
```
with ghc-8.9.0.20190613, linearity errors show up:
 ```
/Users/bmehta/Programming/haskell/tweag/linear-base/src/Data/Bifunctor/Linear.hs:67:23: error:
    • Couldn't match expected multiplicity ‘'One’ of variable ‘x’ with actual multiplicity ‘'Omega’
    • In the pattern: Right x
      In the pattern: Left (Right x)
      In an equation for ‘rassoc’:
          rassoc (Left (Right x)) = Right (Left x)
   |
67 |   rassoc (Left (Right x)) = Right (Left x)
   |                       ^

/Users/bmehta/Programming/haskell/tweag/linear-base/src/Data/Bifunctor/Linear.hs:68:17: error:
    • Couldn't match expected multiplicity ‘'One’ of variable ‘x’ with actual multiplicity ‘'Omega’
    • In the pattern: Right x
      In an equation for ‘rassoc’: rassoc (Right x) = Right (Right x)
      In the instance declaration for ‘SymmetricMonoidal Either Void’
   |
68 |   rassoc (Right x) = Right (Right x)
   |                 ^
```

b-mehta: Notably, there's no error for the pattern `Left`, only for `Right`.

aspiwack: What compiler are you using presently?

b-mehta: For general work I'm using tweag/linear-types:1.0.0 (not entirely sure how to check the exact version past that) - for some reason this started working the other day for me so I've stuck with it.

b-mehta: For that error message I used 1.0.5, with `compiler: ghc-8.9` in my `stack.yaml`.

aspiwack: Ok, I've spoken with @monoidal and we know what the issue is. We will fix it.

In the meantime, `rassoc (Right x) = (Right :: _ ->. _) (Right x)` will work.

(with `-XPartialTypeSignature` the warning on partial type signature deactivated, or by filling the `_`)

aspiwack: Or use the defnition in terms of `either`, but I'm not terribly fond of it.

b-mehta: Thanks! No rush on the fix, I'm happy working with 1.0.0

monoidal: In the next version of the image (not released yet), `Right (Right x)` should work.

PR #39, opened 2019-08-01
Basic optics

b-mehta:
Add basic lens newtypes and simple examples.

Covers the other half of #35, in conjunction with #37.


b-mehta:
Merged.

PR #44, opened 2019-08-05
Add more basic prisms and isos and match/build

b-mehta:
Prisms:

- Add `prism` constructor
- Add `_Just`, `_Nothing`
- `Prism` consumers: `match`, `build` and non-linear `match'` (I haven't yet found any use for `build'`, so it's omitted for now)

Isos:
- `iso` constructor and `withIso` deconstructor

(This leaves the question of a `lens` constructor, which I will address in a forthcoming PR)

### DIFF
--- a/src/Control/Optics/Linear/Internal.hs
+++ b/src/Control/Optics/Linear/Internal.hs
@@ -40,8 +40,8 @@ import Data.Functor.Const
 import Data.Functor.Linear
 import Data.Monoid
 import Data.Profunctor.Linear
-import qualified Data.Profunctor.Kleisli.Linear as L
-import qualified Data.Profunctor.Kleisli.NonLinear as NL
+import qualified Data.Profunctor.Kleisli.Linear as Linear
+import qualified Data.Profunctor.Kleisli.NonLinear as NonLinear
 import Data.Void
 import Prelude.Linear
 import qualified Prelude as P
@@ -96,27 +96,27 @@ traversed = Optical wander
 over :: Optic_ LinearArrow a b s t -> (a ->. b) -> s ->. t
 over (Optical l) f = getLA (l (LA f))
 
-traverseOf :: Optic_ (L.Kleisli f) a b s t -> (a ->. f b) -> s ->. f t
-traverseOf (Optical l) f = L.runKleisli (l (L.Kleisli f))
+traverseOf :: Optic_ (Linear.Kleisli f) a b s t -> (a ->. f b) -> s ->. f t
+traverseOf (Optical l) f = Linear.runKleisli (l (Linear.Kleisli f))
 
-get :: Optic_ (NL.Kleisli (Const a)) a b s t -> s -> a
+get :: Optic_ (NonLinear.Kleisli (Const a)) a b s t -> s -> a
 get l = gets l P.id
 
-gets :: Optic_ (NL.Kleisli (Const r)) a b s t -> (a -> r) -> s -> r
-gets (Optical l) f s = getConst' (NL.runKleisli (l (NL.Kleisli (Const P.. f))) s)
+gets :: Optic_ (NonLinear.Kleisli (Const r)) a b s t -> (a -> r) -> s -> r
+gets (Optical l) f s = getConst' (NonLinear.runKleisli (l (NonLinear.Kleisli (Const P.. f))) s)
 
 set :: Optic_ (->) a b s t -> b -> s -> t
 set (Optical l) x = l (const x)
 
-match :: Optic_ (L.Kleisli (Either a)) a b s t -> s ->. Either t a
-match (Optical l) = withIso swap (\x _ -> x) . L.runKleisli (l (L.Kleisli Left))
+match :: Optic_ (Linear.Kleisli (Either a)) a b s t -> s ->. Either t a
+match (Optical l) = withIso swap (\x _ -> x) . Linear.runKleisli (l (Linear.Kleisli Left))
 
 -- will be redundant with multiplicity polymorphism
-match' :: Optic_ (NL.Kleisli (Either a)) a b s t -> s -> Either t a
-match' (Optical l) = withIso swap (\x _ -> forget x) P.. NL.runKleisli (l (NL.Kleisli Left))
+match' :: Optic_ (NonLinear.Kleisli (Either a)) a b s t -> s -> Either t a
+match' (Optical l) = withIso swap (\x _ -> forget x) P.. NonLinear.runKleisli (l (NonLinear.Kleisli Left))
 
-build :: Optic_ (L.CoKleisli (Const b)) a b s t -> b ->. t
-build (Optical l) x = L.runCoKleisli (l (L.CoKleisli getConst')) (Const x)
+build :: Optic_ (Linear.CoKleisli (Const b)) a b s t -> b ->. t
+build (Optical l) x = Linear.runCoKleisli (l (Linear.CoKleisli getConst')) (Const x)
 
 -- XXX: move this to Prelude
 -- | Linearly typed patch for the newtype deconstructor. (Temporary until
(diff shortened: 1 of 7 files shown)


src/Control/Optics/Linear/Internal.hs
```diff
@@ -53,6 +56,12 @@ _Left = Optical first
 _Right :: Prism a b (Either c a) (Either c b)
 _Right = Optical second
 
+_Just :: Prism a b (Maybe a) (Maybe b)
+_Just = prism Just (maybe (Left Nothing) Right)
```
aspiwack: For the records: I discourage the temptation to use pointfree at all cost. This one is still ok though. But only just.

Conciseness is not always what you want to shoot for. Clarity is.

src/Control/Optics/Linear/Internal.hs
```diff
@@ -66,11 +75,29 @@ get :: Optic_ (Kleisli (Const a)) a b s t -> s -> a
...
 
 gets :: Optic_ (Kleisli (Const r)) a b s t -> (a -> r) -> s -> r
-gets (Optical l) f s = getConst (runKleisli (l (Kleisli (Const P.. f))) s)
+gets (Optical l) f s = getConst' (runKleisli (l (Kleisli (Const P.. f))) s)
 
 set :: Optic_ (->) a b s t -> b -> s -> t
 set (Optical l) x = l (const x)
 
+-- get can't return a linear arrow, so `withIso swap const` needs to be used
+-- instead of `get swap`
```
aspiwack: I'm not sure what his means. You may want to expand on it until it's crystal clear.

> The lens library defines @match@ as … . But this is not possible here, because … .

Etc…

src/Data/Functor/Linear.hs
```diff
@@ -70,3 +71,6 @@ instance Functor [] where
 instance Traversable [] where
   traverse _f [] = pure []
   traverse f (a : as) = (:) <$> f a <*> traverse f as
+
+instance Functor (Const x) where
+  fmap _ (Const x) = Const x
```
aspiwack: It's also `Applicative` (when `x` is a monoid) and `Traversable`. We might as well define these.

Also, we should definitly reexport `Const` (as well as, eventually, many functors) in `Functor.Linear`.

src/Data/Profunctor/Linear.hs
```diff
@@ -116,3 +122,20 @@ instance Strong (,) () (->) where
 instance Strong Either Void (->) where
   first f (Left x) = Left (f x)
   first _ (Right y) = Right y
+
+-- XXX: Since CoLKleisli has uses, it might be better to replace all this
+-- with a Bif-like structure...
+--
+newtype CoLKleisli w a b = CoLKleisli { runCoLKleisli :: w a ->. b }
```
aspiwack: I'd like to reiterate my dislike of mangling names like this. I'm _much_ rather have two different files and qualifed imports for the two Kleisli arrows (and Cokleisli).

src/Prelude/Linear.hs
```diff
@@ -71,6 +72,12 @@ maybe :: b -> (a ->. b) -> Maybe a ->. b
 maybe x _ Nothing = x
 maybe _ f (Just y) = f y
 
+-- XXX: temporary
```
aspiwack: Ok. It's temporary. But what will obsolete it?

b-mehta: Multiplicity polymorphism. This function is useful only because we don't have enough polymorphism to give a linear function to one expecting a non-linear arrow, so it's a temporary fix to make that work.
I was thinking of changing its name to `eta` since it has the same effect as eta-expanding `f` at use-sites, if you think that would be more useful.

src/Data/Profunctor/Linear.hs
b-mehta: Okay sure!

src/Control/Optics/Linear/Internal.hs
b-mehta: Ah this was much more of a throwaway comment: I just meant that in the two definitions `match` and `match'` given here, both need a `swap` to convert `Either a t` to `Either t a`. To do this, one could use `get swap` to swap it, but `get swap` isn't linear so wouldn't work for `match`. 

In hindsight it would probably be better to write
```haskell
match' (Optical l) = forget (withIso swap (\x _ -> x)) P.. runKleisli (l (Kleisli Left))
```

src/Data/Functor/Linear.hs
b-mehta: If it's okay, I'll hold off on defining the `Applicative` instance in this PR and add it when I add linear monoids.

src/Control/Optics/Linear/Internal.hs
b-mehta: I've changed the definition and removed this comment.

src/Data/Profunctor/Linear.hs
b-mehta: I've made two different files with qualified imports like you suggest - it's a lot cleaner.

src/Prelude/Linear.hs
b-mehta: I've added a comment to reflect this.

src/Data/Functor/Linear.hs
aspiwack: Sure.

src/Control/Optics/Linear/Internal.hs
```diff
@@ -6,14 +6,42 @@
...
+  )
+  where
 
 import qualified Data.Bifunctor.Linear as Bifunctor
 import Data.Bifunctor.Linear (SymmetricMonoidal)
 import Data.Functor.Const
 import Data.Functor.Linear
 import Data.Monoid
 import Data.Profunctor.Linear
+import qualified Data.Profunctor.Kleisli.Linear as L
```
aspiwack: Please no single-letter imports. `Linear` will do :slightly_smiling_face:

src/Control/Optics/Linear/Internal.hs
```diff
@@ -6,14 +6,42 @@
...
+import qualified Data.Profunctor.Kleisli.NonLinear as NL
```
aspiwack: There is no good name for this one. `Hask` could be? `Unrestricted` would be a little confusing… I guess `NonLinear` is clear enough in this context though.

src/Data/Profunctor/Kleisli/Linear.hs
```diff
@@ -0,0 +1,68 @@
...
+-- profunctorial properties still hold in this weaker setting.
+-- However stronger requirements on `f` are needed for profunctorial
+-- strength, so we have fewer instances.
+--
+-- Category theoretic remark: duality doesn't work in the obvious way, since
+-- (,) isn't the categorical product. Instead, we have a product (&), called
+-- "With", defined by
+-- > type With a b = forall r. Either (a ->. r) (b ->. r) ->. r
+-- which satisfies the universal property of the product of `a` and `b`.
+-- CoKleisli arrows are strong with respect to this monoidal structure,
```
aspiwack: Are they really? That's quite interesting. I don't know about useful. But interesting.

src/Prelude/Linear.hs
```diff
@@ -72,7 +72,9 @@ maybe :: b -> (a ->. b) -> Maybe a ->. b
 maybe x _ Nothing = x
 maybe _ f (Just y) = f y
 
--- XXX: temporary
+-- XXX: temporary: with multiplicity polymorphism functions expecting a
+-- non-linear arrow would allow a linear arrow passed, so this would be
```
aspiwack: Well, not always. But manually eta-expanding is probably not very costly anyway.

src/Data/Profunctor/Kleisli/Linear.hs
b-mehta: Oops, you're right - this needs `w` to be a copointed functor: it's exactly dual to the `Strong Either Void (Kleisli f)` instance.

src/Prelude/Linear.hs
b-mehta: This is true, but I think
```haskell
  dimap f g (Kleisli h) = Kleisli (\x -> forget g Prelude.<$> h (f x))
```
is clearer than
```haskell
  dimap f g (Kleisli h) = Kleisli (\x -> (\y -> g y) Prelude.<$> h (f x))
```
to read.

aspiwack: Which is a use case which would, indeed, disapear with polymorphism.

b-mehta:
Merged.

PR #46, opened 2019-08-07
Linear semigroups

b-mehta:
Add linear semigroups and the Top monoid.

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -23,6 +23,7 @@ library
     Control.Optics.Linear.Internal
     Data.Bifunctor.Linear
     Data.Functor.Linear
+    Data.Semigroup.Linear
     Data.Vector.Linear
     Data.Profunctor.Linear
     Data.Profunctor.Kleisli.Linear
--- a/src/Data/Semigroup/Linear.hs
+++ b/src/Data/Semigroup/Linear.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE NoImplicitPrelude #-}
@@ -12,6 +13,7 @@ module Data.Semigroup.Linear
   ( Semigroup(..)
   , Monoid(..)
   , LEndo(..), appLEndo
+  , Top, throw
   , module Data.Semigroup
   )
   where
@@ -29,7 +31,8 @@ class (Semigroup a, Prelude.Monoid a) => Monoid a where
   {-# MINIMAL #-}
   mempty :: a
   mempty = mempty
-  -- convenience redefine
+  -- convenience redefine, so that nonlinear Data.Semigroup doesn't have
+  -- to be imported to use mempty
 
 ---------------
 -- Instances --
@@ -38,6 +41,7 @@ class (Semigroup a, Prelude.Monoid a) => Monoid a where
 instance Semigroup () where
   () <> () = ()
 
+-- | The type of linear endomorphisms.
 newtype LEndo a = LEndo (a ->. a)
 
 -- TODO: have this as a newtype deconstructor once the right type can be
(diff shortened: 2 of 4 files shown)


b-mehta:
The use case I had in mind for `Top` doesn't apply like I thought it did - so I'm less strongly in favour of adding it. I still think it might come in handy at some point, but I'd be okay leaving it out for the moment also.

linear-base.cabal
```diff
@@ -23,6 +23,7 @@ library
     Control.Optics.Linear.Internal
     Data.Bifunctor.Linear
     Data.Functor.Linear
+    Data.Semigroup.Linear
```
aspiwack: Actually: rename this `Monoid`.

src/Data/Semigroup/Linear.hs
```diff
@@ -0,0 +1,106 @@
...
+instance Semigroup a => Semigroup (Dual a) where
+  Dual x <> Dual y = Dual (y <> x)
+instance Monoid a => Monoid (Dual a)
+
+newtype LWrap a = LWrap a
+  deriving (Prelude.Semigroup, Prelude.Monoid)
+
+-- This instance is unsafe: do not export LWrap so it cannot be used.
+instance Prelude.Semigroup a => Semigroup (LWrap a) where
+  LWrap a <> LWrap b = LWrap (Unsafe.toLinear2 (Prelude.<>) a b)
```
aspiwack: It's actually unsafe for `movable` things too. Unless `(<>)` is hyperstrict in both arguments. Which it rarely is. You really need to force all the linear thunks that you may encounter. Sorry I'd missed that on my first review.

It's totally ok to define these instances manually. They are not big.

src/Data/Semigroup/Linear.hs
```diff
@@ -0,0 +1,106 @@
...
+instance Prelude.Monoid a => Monoid (LWrap a)
+
+-- XXX: I think these are safe but I'm not fully confident
```
aspiwack: No, they unfortunately are not.

linear-base.cabal
b-mehta: Is there a particular reason this is better? I picked `Semigroup` since it matches the pattern in base where importing Semigroup also brings into scope Monoids too

src/Data/Semigroup/Linear.hs
b-mehta: Ahh this is frustrating - I'll edit. Could you explain why this would fail for movable things?

src/Data/Semigroup/Linear.hs
b-mehta: Does this include the boolean instances below as well? My hope was that even though `LWrap` is unsafe, the simple cases of `Sum` and `Product` would still be safe

linear-base.cabal
aspiwack: Just that `Monoid` is a more common structure than `Semigroup`. So it's a better namespace.

src/Data/Semigroup/Linear.hs
aspiwack: Suppose you have a thunk of type `True`. It may have, in its closure, some linear variables. As long as you haven't pattern-matched on the thunk, then this linear variable hasn't been consumed. Which breaks linearity.

Concretely, let's take boolean disjunction

```haskell
or :: Bool -> Bool -> Bool
or True _ = True
or False b = b
```

I claim that `or`, thus defined, is not a linear function. For it to be a linear function, it would need, per the defintion, that consuming `or x y` exactly once would consume `x` and `y` exactly once. But if `x` is `True`, then consuming `or x y` would not consume `y`.

Booleans are movable alright, but moving them _does_ require deep copy (which is not too deep in this case, but still).

src/Data/Semigroup/Linear.hs
aspiwack: As long as you don't define them with functions of the `Unsafe` module, the boolean instances are safe.

src/Data/Semigroup/Linear.hs
b-mehta: I see, thanks!

b-mehta:
Resolved by #52

PR #50, opened 2019-08-08
Linear traversables

### LINKED ISSUE #48: Linear traversables
In the past, I'd edited `traverse` to have type signature
```haskell
  traverse :: Data.Applicative f => (a ->. f b) -> t a ->. f (t b)
```
(and similar for `sequence`) - it now seems this was misguided, since it prevents `Const`, `((,) a)`, `Either e` from being `Traversable`. (And these last two instances are why `Strong (,) ()` and `Strong Either Void` are superclasses of `Wandering`). Instead, the correct signature probably ought to be
```haskell
  traverse :: Control.Applicative f => (a ->. f b) -> t a ->. f (t b)
```
This change should not prevent any instance from still working, but might need some definitions to be edited.

b-mehta:
Resolves #48

### DIFF
--- a/src/Data/Traversable/Linear.hs
+++ b/src/Data/Traversable/Linear.hs
@@ -1,0 +1,73 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Data.Traversable.Linear
+  ( -- * Linear traversable hierarchy
+    -- $ traversable
+  Traversable(..)
+  ) where
+
+import qualified Control.Monad.Linear as Control
+import qualified Data.Functor.Linear as Data
+import Data.Functor.Const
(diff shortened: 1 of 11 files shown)


src/Data/Traversable/Linear.hs
```diff
@@ -0,0 +1,76 @@
...
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Data.Traversable.Linear
```
aspiwack: I'll make you work a bit more. I realise you pulled this out of the `Functor.Linear` module because of the dependency cycle.

But the correct answer is to make a `Functor.Linear.Internal` module, which is reexported in `Functor.Linear` (only the `Functor.Linear` module should have an export list, `Functor.Linear.Internal` should just expose everything).

That way you can also make a `Data.Functor.Linear.Internal.Traversable` functor, which is later in the dependency chain, and reexport it from `Data.Functor.Linear` and you avoid the multiplication of things to import.

That gives you a bit more work, sorry about that, but it's for a good cause.

b-mehta: Sounds good, should be done!

PR #51, opened 2019-08-08
Linear semigroups (attempt 2)

b-mehta:
Moved `Unrestricted`, `Movable`, `Dupable`, `Consumable` to their own module (still exported by `Prelude.Linear` though) and added `Data.Monoid.Linear`.

Replaces #46.

### DIFF
--- a/src/Data/Monoid/Linear.hs
+++ b/src/Data/Monoid/Linear.hs
@@ -1,0 +1,103 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE StandaloneDeriving #-}
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- | = The linear semigroup hierarchy
+--
+-- TODO: documentation
+
+module Data.Monoid.Linear
+  ( Semigroup(..)
+  , Monoid(..)
+  , LEndo(..), appLEndo
+  , module Data.Semigroup
+  )
+  where
+
+import Prelude.Linear.Internal.Simple
+import Data.Unrestricted.Linear
+import Data.Functor.Linear as Data
+import Data.Functor.Const
+import Data.Semigroup hiding (Semigroup(..))
+import qualified Data.Semigroup as Prelude
+import qualified Prelude
+
+class Prelude.Semigroup a => Semigroup a where
+  (<>) :: a ->. a ->. a
+
+class (Semigroup a, Prelude.Monoid a) => Monoid a where
+  {-# MINIMAL #-}
+  mempty :: a
+  mempty = mempty
+  -- convenience redefine
+
+---------------
+-- Instances --
+---------------
+
+instance Semigroup () where
+  () <> () = ()
+
+newtype LEndo a = LEndo (a ->. a)
+
+-- TODO: have this as a newtype deconstructor once the right type can be
+-- correctly inferred
+appLEndo :: LEndo a ->. a ->. a
+appLEndo (LEndo f) = f
+
+instance Prelude.Semigroup (LEndo a) where
+  LEndo f <> LEndo g = LEndo (f . g)
+instance Prelude.Monoid (LEndo a) where
+  mempty = LEndo id
+instance Semigroup (LEndo a) where
+  LEndo f <> LEndo g = LEndo (f . g)
+instance Monoid (LEndo a) where
+
+instance (Semigroup a, Semigroup b) => Semigroup (a,b) where
+  (a,x) <> (b,y) = (a <> b, x <> y)
+instance (Monoid a, Monoid b) => Monoid (a,b)
+
+instance Semigroup a => Semigroup (Dual a) where
+  Dual x <> Dual y = Dual (y <> x)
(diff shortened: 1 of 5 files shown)


src/Data/Monoid/Linear.hs
```diff
@@ -0,0 +1,103 @@
...
+deriving instance Movable a => Movable (Sum a)
+deriving instance Consumable a => Consumable (Product a)
+deriving instance Dupable a => Dupable (Product a)
+deriving instance Movable a => Movable (Product a)
+deriving instance Consumable All
+deriving instance Dupable All
+deriving instance Movable All
+deriving instance Consumable Any
+deriving instance Dupable Any
+deriving instance Movable Any
```
b-mehta: These instances are orphans if they're here - do you think it's sensible to move them to Data.Unrestricted.Linear instead of putting them here?

src/Data/Monoid/Linear.hs
```diff
@@ -0,0 +1,103 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE StandaloneDeriving #-}
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- | = The linear semigroup hierarchy
```
aspiwack:
```suggestion
-- | = The linear monoid hierarchy
```

src/Data/Monoid/Linear.hs
```diff
@@ -0,0 +1,103 @@
...
+  -- convenience redefine
+
+---------------
+-- Instances --
+---------------
+
+instance Semigroup () where
+  () <> () = ()
+
+newtype LEndo a = LEndo (a ->. a)
```
aspiwack: Again with the `L` :stuck_out_tongue: .

```suggestion
newtype Endo a = Endo (a ->. a)
```

src/Data/Monoid/Linear.hs
aspiwack: Aren't these instance supposed to move the the appropriate module? It's ok if you do it later, but then, make yourself an issue as you close this (also to remove the `-fno-warn-orphans` from the top of the file).

src/Data/Monoid/Linear.hs
```diff
@@ -0,0 +1,103 @@
...
+deriving instance Dupable All
+deriving instance Movable All
+deriving instance Consumable Any
+deriving instance Dupable Any
+deriving instance Movable Any
+
+deriving via (LWrap (Sum a)) instance (Movable a, Prelude.Num a) => Semigroup (Sum a)
+deriving via (LWrap (Sum a)) instance (Movable a, Prelude.Num a) => Monoid (Sum a)
+deriving via (LWrap (Product a)) instance (Movable a, Prelude.Num a) => Semigroup (Product a)
+deriving via (LWrap (Product a)) instance (Movable a, Prelude.Num a) => Monoid (Product a)
```
aspiwack: These four are bad instances. It is unclear that you want to define these that way if `Move` is not trivially fast. It's fine to provide these as `deriving via` constructors, but not as blanket instances, I don't think.

src/Data/Monoid/Linear.hs
```diff
@@ -0,0 +1,103 @@
...
+
+deriving via LWrap All instance Semigroup All
+deriving via LWrap All instance Monoid All
+deriving via LWrap Any instance Semigroup Any
+deriving via LWrap Any instance Monoid Any
+
+instance Monoid x => Data.Applicative (Const x) where
+  pure _ = Const mempty
+  Const x <*> Const y = Const (x <> y)
```
aspiwack: I'd imagine this instance to be rather in `Data.Functor.Linear` though. Is there an obstacle to that?

src/Data/Unrestricted/Linear.hs
```diff
@@ -0,0 +1,175 @@
...
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE RankNTypes #-}
+module Data.Unrestricted.Linear
+  ( -- * Unrestricted
+    -- $ unrestricted
+    Unrestricted(..)
+  , unUnrestricted
+    -- * Typeclasses for non-linear actions
```
aspiwack: I'm not sure “non-linear" is quite appropriate here. I'd rather use something to the effect of

```suggestion
    -- * Duplicating and consuming value linearly
```

src/Data/Unrestricted/Linear.hs
```diff
@@ -0,0 +1,175 @@
...
+
+-- | Like 'seq' but the first argument is restricted to be 'Consumable'. Hence the
+-- first argument is 'consume'-ed and the result consumed.
+lseq :: Consumable a => a ->. b ->. b
+lseq a b = seqUnit (consume a) b
+
+-- | The laws of @Dupable@ are dual to those of 'Monoid':
+--
+-- * @first consume (dup a) ≃ a ≃ first consume (dup a)@ (neutrality)
+-- * @first dup (dup a) ≃ (second dup (dup a))@ (associativity)
```
aspiwack: We will really have to go through the documentation of this module one day: it's at best stale.

b-mehta:
Resolved by #52.

PR #54, opened 2019-08-13
Deriving Via combinator for Semigroup/Monoid

b-mehta:
Add a deriving via combinator for Prelude.Semigroup given Linear.Semigroup, and similar for Monoid.

### DIFF
--- a/src/Data/Monoid/Linear.hs
+++ b/src/Data/Monoid/Linear.hs
@@ -13,6 +13,7 @@ module Data.Monoid.Linear
   ( Semigroup(..)
   , Monoid(..)
   , Endo(..), appEndo
+  , NonLinear(..)
   , module Data.Semigroup
   )
   where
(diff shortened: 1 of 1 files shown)


src/Data/Monoid/Linear.hs
```diff
@@ -72,3 +71,13 @@ instance Semigroup Any where
   Any False <> Any True = Any True
   Any True  <> Any False = Any True
   Any True  <> Any True = Any True
+
+-- | DerivingVia combinator for Prelude.Semigroup given (linear) Semigroup.
+-- To use for Monoid, you must supply mempty in the Linear.Monoid instance,
+-- even though GHC will not warn you if this is omitted.
```
b-mehta: This is annoying. If `mempty` is not supplied here, then (I'm pretty sure) there'll be an infinite loop if it's used.
Possible solutions are:

  - Remove the minimal pragma, which would be frustrating for those manually writing instances.
  - Don't supply `instance Monoid a => Prelude.Monoid (NonLinear a)`, which would mean users have to write a `Prelude.Monoid` instance by hand and include `instance Monoid where` also.
  - Give a default instance?

aspiwack: You can remove the `mempty` method from `Monoid`, and define it as a function

```haskell
mempty :: Monoid a => a
mempty = Prelude.mempty
```

It's tad redundant there.

You can't really do a deriving via for `Prelude.Monoid` then either. However, what you can do is a blanket instance in that case:

```haskell
instance (Semigroup a, Prelude.Monoid a) => Monoid a
```

Because it really is canonical (in fact you can do that even keeping the mempty method; nobody will need to care since there is really just one instance).

b-mehta: Hmm, using this seems somewhat awkward also: it requires `UndecidableInstances` (which normally I wouldn't mind, but it feels like overkill for just monoids...) and creates warnings for things like
```haskell
instance Monoid x => Applicative (Const x)
```

aspiwack: Ah yeah, the infamous unreduced type class constraint thing. Of course we could decide that `Monoid` is defined as `type Monoid a = (Semigroup a, Prelude.Monoid a)`. But I'm not sure whether this is a good idea.

b-mehta:
Do you think this is okay to merge now?

PR #55, opened 2019-08-14
(WIP) Experimental linear num

b-mehta:
First brush at what a linear num typeclass might look like

### DIFF
--- a/src/Data/Num/Linear.hs
+++ b/src/Data/Num/Linear.hs
@@ -1,0 +1,54 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE RebindableSyntax #-}
+{-# LANGUAGE StandaloneDeriving #-}
+
+module Data.Num.Linear
+  ( Num(..)
+  )
+  where
+
+import qualified Prelude
+import Data.Unrestricted.Linear
+import qualified Unsafe.Linear as Unsafe
+
+-- XXX: subclass of Prelude.Num? subclass of Eq?
+class Num a where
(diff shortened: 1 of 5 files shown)


src/Data/Num/Linear.hs
```diff
@@ -0,0 +1,54 @@
...
+liftU f x = lifted f (move x)
+  where lifted :: (a -> b) ->. (Unrestricted a ->. b)
+        lifted g (Unrestricted a) = g a
+
+liftU2 :: (Movable a, Movable b) => (a -> b -> c) ->. (a ->. b ->. c)
+liftU2 f x y = lifted f (move x) (move y)
+  where lifted :: (a -> b -> c) ->. (Unrestricted a ->. Unrestricted b ->. c)
+        lifted g (Unrestricted a) (Unrestricted b) = g a b
+
+deriving via MovableNum Prelude.Int instance Num Prelude.Int
```
aspiwack: Frankly, I'm rather worried of deriving that one because of how it will probably not inline very well. It's not particularly urgent a concern. But it may make numeric code very slow.

b-mehta: Hmm, but don't we need to use explicit `move`s to make sure the linear thunks are forced, similar to our discussion for semigroups?
And the intention is not to export this combinator, only to use it here for easy Num instances, since they should be defined by `move`.

b-mehta:
This is a reasonable concern, I was looking at [Purescript's Num hierarchy](https://pursuit.purescript.org/packages/purescript-numerics/0.1.1/docs/Data.Num) for possible alternatives: where `Semiring` has `(+)` and `(*)` and `Ring` has `(-)`.

b-mehta:
Summarising discussion: 

- an even more fine-grained hierarchy would be nicer, with Additive and Multiplicative as independent superclasses of Semiring, along with a Group class, and a separate FromInteger class (consider a different name)
- The Num Int instance given here would be slow, but there's no real way to fix it at the moment so we'll go with it for now. Ideally, `I#` would behave in a similar way to unrestricted, since an unboxed unlifted `Int#` is completely movable

b-mehta:
Closing in favour of #69.

PR #56, opened 2019-08-16
Destination, pull and push arrays

b-mehta:


src/Data/Array/Destination.hs
```diff
@@ -0,0 +1,112 @@
...
+  | otherwise =
+      mirrorHeadAndTail (Vector.head as) (Vector.tail as) (split 1 ds)
+  where
+    -- /!\ The first DArray has length 1.
+    mirrorHeadAndTail :: a -> Vector a -> (DArray b, DArray b) ->. ()
+    mirrorHeadAndTail x xs (dl, dr) =
+      fill (f x) dl `lseq` mirror xs f dr
+
+-- XXX: would it be better to implement this directly, instead of in terms of
+-- mirror?
```
aspiwack: Yes, much better. The extra allocation is rather unwarranted. Though you can still merge and make an issue of it.

src/Data/Array/Polarized.hs
```diff
@@ -0,0 +1,28 @@
+{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Data.Array.Polarized
+  ( transfer
+  )
+  where
+
+-- XXX: there is lots of scope for testing array functions
```
aspiwack: Quite true. This is the sort of things which can be quickchecked pretty well.

src/Data/Array/Polarized/Pull.hs
```diff
@@ -0,0 +1,54 @@
...
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+
+-- | This module is intended to be imported qualified, e.g.
+-- > import qualified Data.Array.Polarized.Pull as Pull
+module Data.Array.Polarized.Pull
+  ( Array
+  , zip, zipWith
+  , append
```
aspiwack: `reverse` is conspicuously missing.

src/Data/Array/Polarized/Pull/Internal.hs
```diff
@@ -0,0 +1,73 @@
...
+import Prelude.Linear
+import qualified Prelude
+import Data.Vector (Vector)
+import qualified Data.Vector as Vector
+
+import qualified Unsafe.Linear as Unsafe
+
+-- | A pull array represents arrays which are easy to extract elements from.
+-- They provide an indexing function, and a length. If the indexing function
+-- is called at a point out of range, the result is unspecified.
```
aspiwack: It should be specified to be an error, I believe.

src/Data/Array/Polarized/Pull/Internal.hs
```diff
@@ -0,0 +1,73 @@
...
+data Array a where
+  Array :: (Int -> a) -> Int -> Array a
+-- Int is movable, so this is the same as Array :: (Int -> a) -> ...
```
aspiwack: These are syntactically the same thing. One of these you surely intended to be different. (probably the comment).

src/Data/Array/Polarized/Pull/Internal.hs
```diff
@@ -0,0 +1,73 @@
...
+
+instance Data.Functor Array where
+  fmap f (Array g n) = Array (\x -> f (g x)) n
+
+-- XXX: This doesn't match how PullArrays should work, but it is accepted
```
aspiwack: The type of pull arrays is abstract specifically because some things which are accepted are not valid. So I'm not sure it's a problem.

src/Data/Array/Polarized/Pull/Internal.hs
```diff
@@ -0,0 +1,73 @@
...
+-- ap :: Array (x ->. y) ->. Array x ->. Array y
+-- ap (Array g n) (Array h m) = Array (\k -> let (q,r) = k `quotRem` n in g q (h r)) (n*m)
```
aspiwack: Yeah, let's not.

src/Data/Array/Polarized/Pull/Internal.hs
```diff
@@ -0,0 +1,73 @@
...
+singleton = Unsafe.toLinear (\x -> Array (\_ -> x) 1)
+
+-- | /!\ Partial! Only works if both arrays have the same length.
+zip :: Array a ->. Array b ->. Array (a,b)
+zip (Array g n) (Array h m)
+  | n /= m    = error "Polarized.zip: size mismatch"
+  | otherwise = Array (\k -> (g k, h k)) n
+
+append :: Array a ->. Array a ->. Array a
+append (Array f m) (Array g n) = Array (\k -> if k < m then f k else g k) (m + n)
```
aspiwack: This is suspicious. My intuition tells me that we shouldn't be able to append pull-arrays.

src/Data/Array/Polarized/Pull.hs
```diff
@@ -0,0 +1,54 @@
...
+  , make
+  , foldr
+  , foldMap
+  , findLength
+  , asList
+  , singleton
+  , fromFunction
+  , fromVector
+  , toVector
```
aspiwack: We're lacking a `split` as well. I see splitting functions as the defining characteristic of pull arrays.

src/Data/Array/Polarized/Push.hs
```diff
@@ -0,0 +1,62 @@
...
+instance Prelude.Semigroup (Array a) where
+  a <> b = append a b
+instance Semigroup (Array a) where
+  (<>) = append
+
+-- XXX: the use of Vector in the type of alloc is temporary (see also "Data.Array.Destination")
+alloc :: Array a ->. Vector a
+alloc (Array k n) = DArray.alloc n (k id)
+
+walk :: Vector a -> Array a
```
aspiwack: This is a shortcut, isn't it? It should be split into a function from vector to push array and a function for push array to pull array, right?

src/Data/Array/Polarized/Push.hs
```diff
@@ -0,0 +1,62 @@
...
+walk as =
+  Array
+    (\g dest -> DArray.mirror as g dest)
+    (Vector.length as)
+
+make :: a -> Int -> Array a
+make x n = Array (\k -> DArray.replicate (k x)) n
```
aspiwack: Shouldn't we use the `fromFunction` thing instead?

src/Data/Array/Polarized/Pull/Internal.hs
b-mehta: Oops, yeah, it was meant to be a linear arrow in the comment

src/Data/Array/Polarized/Pull/Internal.hs
b-mehta: This itself isn't a problem, but it's concerning that `singleton` below isn't accepted...

src/Data/Array/Polarized/Pull/Internal.hs
b-mehta: Sure.

src/Data/Array/Polarized/Push.hs
b-mehta: I'm not really sure what you mean here - fromFunction right now is for PullArrays not push arrays right now. I guess we could define `make` by creating a pull array and `transfer`ring?

src/Data/Array/Polarized/Push.hs
b-mehta: Again I'm not clear what you mean, this is a function from vector to push array.

src/Data/Array/Polarized/Push.hs
aspiwack: But there is one in destination, right? And it would make, I assume, for a better `make` primitive to take an `(Int -> a)`, instead of the `a`. That's what I was, admittedly clumsily, attempting to say.

src/Data/Array/Polarized.hs
```diff
@@ -0,0 +1,39 @@
...
+import Data.Vector (Vector)
+
+
+-- See:
+--
+-- - http://lopezjuan.com/limestone/vectorcomp.pdf
+-- - http://jyp.github.io/posts/controlled-fusion.html
+--
+-- The general spirit is: `Array` are those arrays which are friendly in
+-- returned-value possition. And `PullArray` are those arrays which are friendly
```
aspiwack:
```suggestion
-- returned-value position. And `PullArray` are those arrays which are friendly
```

The whole documentation is fairly out of date, though. But it contains an nice intuition :slightly_smiling_face: .

src/Data/Array/Polarized/Push.hs
b-mehta: I see! It would, and this would just be implemented by making a PullArray and transferring, I'll make an issue for this.

PR #58, opened 2019-08-16
A few functions used for the jni package

facundominguez:


src/Prelude/Linear.hs
```diff
@@ -74,3 +83,29 @@ maybe _ f (Just y) = f y
...
+flip :: (a -->.(p) b -->.(q) c) -->.(r) b -->.(q) a -->.(p) c
+flip f b a = f a b
+
+-- | Linearly typed replacement for the standard '(<$)' function.
+(<$) :: (Applicative f, Consumable b) => a ->. f b ->. f a
+a <$ fb = pure (flip lseq a) <*> fb
+
+-- | Linearly typed replacement for the standard '(<*)' function.
+(<*) :: (Applicative f, Consumable b) => f a ->. f b ->. f a
+fa <* fb = fmap (flip lseq) fa <*> fb
```
facundominguez: I wanted to place `(<*)` and `(<$)` in `Control.Monad` but that creates a circular dependency because of the use of Consumable.

src/Prelude/Linear.hs
```diff
@@ -74,3 +83,29 @@ maybe _ f (Just y) = f y
...
+
+-- | Linearly typed replacement for the standard 'foldM' function.
+foldM :: forall m a b. Monad m => (b ->. a ->. m b) -> b ->. [a] ->. m b
+foldM f z0 xs = foldr f' return xs z0
+  where
+    f' :: a ->. (b ->. m b) ->. b ->. m b
+    f' x k z = f z x >>= k
```
facundominguez: I also wanted to add `foldM` to `Control.Monad.Linear` but then I was unsure where to put its dependency `foldr`.

src/Data/Unrestricted/Linear.hs
```diff
@@ -48,6 +49,9 @@ data Unrestricted a where
 unUnrestricted :: Unrestricted a ->. a
 unUnrestricted (Unrestricted a) = a
 
+uliftA2 :: (a -> b -> c) -> Unrestricted a ->. Unrestricted b ->. Unrestricted c
+uliftA2 f (Unrestricted a) (Unrestricted b) = Unrestricted (f a b)
```
aspiwack: rename: `liftA2`. This module is imported qualified.

src/Prelude/Linear.hs
```diff
@@ -74,3 +83,29 @@ maybe _ f (Just y) = f y
 -- arrow but we have a linear arrow
 forget :: (a ->. b) ->. a -> b
 forget f x = f x
+
+-- | Linearly typed replacement for the standard 'foldr' function, to allow
```
aspiwack: This `'foldr'` will point to this same function. Use `'Prelude.foldr'` for correct hyperlinks. I think.

src/Prelude/Linear.hs
aspiwack: Time for more crazy `Internal` cycle breaking? Or maybe these have their rightful place in the prelude?

If cycle breaking is preferable: don't hold this PR, and make an issue for it.

src/Prelude/Linear.hs
aspiwack: `Prelude.Linear.Internal.Simple`

src/Unsafe/Linear.hs
```diff
@@ -32,3 +32,7 @@ toLinear2
      (a :: TYPE r1) (b :: TYPE r2) (c :: TYPE r3).
      (a -> b -> c) ->. (a ->. b ->. c)
 toLinear2 = coerce
+
+-- | Like 'toLinear' but for three-argument functions
+toLinear3 :: (a -> b -> c -> d) ->. (a ->. b ->. c ->. d)
+toLinear3 f = toLinear2 (\a b -> toLinear (f a b))
```
aspiwack:
```suggestion
toLinear3 = coerce
```

No need to be overly conservative here. Everything is unsafe in this module anyway.

src/Prelude/Linear.hs
b-mehta: I think `Internal` is a sensible thing to do here, so that they could be imported qualified by `Control` when the `Data` version is also necessary

src/Data/Unrestricted/Linear.hs
b-mehta: I'm not sure I think `liftA2` is the right name for this function, it suggests that the argument ought to be `a ->. b ->. c` (since this is what's expected of Data and Control `liftA2`)

aspiwack: Well, `liftA2` is alright. Though `lift2` is fine with me too.

 `Unrestricted` is pretty special because it's a functor from the category of unrestricted functions to the category of linear functions. We don't really have an abstraction for that. This is saying that it preserves (monoidal) products, which is sort of like applicative (if we don't decide that _applicative_ is not only monoidal but also endo).

I don't care either way to be honest :slightly_smiling_face: . I just want the `u` to go as I want linear-base to use namespaces for namespacing rather than name conventions.

b-mehta: I'm happy with `lift2`

b-mehta: Similar to this, it would also be nice to have an unary version `(a -> b) -> Unrestricted a ->. Unrestricted b`.

src/Prelude/Linear.hs
b-mehta: On further thoughts (apologies I missed this earlier), both `(<*)` and `(<$)` can be given more general type signatures:
```haskell
(<$) :: (Control.Functor f, Consumable b) => a ->. f b ->. f a
a <$ fb = Control.fmap (`lseq` a) fb

(<*) :: (Data.Applicative f, Consumable b) => f a ->. f b ->. f a
fa <* fb = Data.fmap (flip lseq) fa Data.<*> fb
```
(and these cover the ones you've given).

PR #64, opened 2019-08-22
Add MonadIO.

facundominguez:
Using `Linear.IO` in linear interfaces does not make them easy to use in applications which use other monads.

I propose here a class to lift non-linear monads into linear monads. jni uses this to offer operations that run on any linear monad that can lift the non-linear `IO`.

```haskell
class (Linear.Monad m, NonLinear.Monad (LiftedM m)) => MonadLift m where
  type LiftedM m :: * -> *
  lift :: LiftedM m a -> m a
  liftU :: LiftedM m a -> m (Unrestricted a)
```

Thus, an application can provide an instance of `Linear.Monad` for any wrapper of a regularly looking monadic stack. e.g.

```haskell
data App a = App (ReaderT Env IO a)
  deriving (Monad, Functor, Applicative, MonadIO)

data LinearApp a = LinearApp (App a)

instance Linear.Monad LinearApp where
  (>>=) = Unsafe.toLinear2 Linear.$ \(LinearApp m) (f :: a ->. LinearApp b) ->
    LinearApp $ m >>= \x -> case f x of LinearApp n -> n
  (>>) = Unsafe.toLinear2 Linear.$ \(LinearApp m0) (LinearApp m1) ->
    LinearApp (m0 >> m1)

instance Linear.MonadLift LinearApp where
  type LiftedM LinearApp = App
  lift = LinearApp
  liftU = LinearApp . fmap Unrestricted
```

This is shown in action in https://github.com/tweag/inline-java/tree/master/examples/directory-server

I'm not sure about the name of `MonadLift`, but I don't have any better names so far.


b-mehta:
I'd (weakly) prefer if the `MonadLift IO` instance were placed in `IO.Class.Linear`, and this module imports `System.IO.Linear` instead: it seems to me that System.IO.Linear shouldn't depend on monad lifting modules and should be lower in the import hierarchy. 

As an aside, I plan to add linear monad instances for the standard monad transformers, (where applicable) so the `Linear.Monad` instance for, say `LinearApp`, would be safe and newtype-derivable.

b-mehta:
> I plan to add linear monad instances for the standard monad transformers, (where applicable) so the `Linear.Monad` instance for, say `LinearApp`, would be safe and newtype-derivable.

This is proposed in #68.

facundominguez:
I deleted the branch, convinced that it had been merged long ago. Therefore I reopen it.

src/Control/Monad/IO/Class/Linear.hs
```diff
@@ -0,0 +1,21 @@
...
+{-# LANGUAGE NoImplicitPrelude #-}
+module Control.Monad.IO.Class.Linear where
+
+import qualified Control.Monad.IO.Class as NonLinear
+import Control.Monad.Lift.Linear
+import Prelude.Linear
+
+-- | Like 'NonLinear.MonadIO' but allows to lift non-linear 'IO'
+-- actions into a linear monad
+type MonadIO m = (MonadLift m, NonLinear.MonadIO (LiftedM m))
```
aspiwack: This doesn't feel right. What we want of a linear `MonadIO` is to be able to `liftIO` linear `IO` functions. Not non-linear ones.

If we can lift linear functions, we can in particular lift non-linear `IO` functions (via the `MonadLift IO` instance). We may or may not want to have a corresponding non-iinear monad to lift from. But it should be an orthogonal concern.

facundominguez:
> We want to write functions parametrised over arbitray MonadLift m

This is the case in inline-java.

```Haskell
newLocalRef :: MonadIO m => J ty ->. m (J ty, J ty)
```

This function works on any monad that is linear and which can lift IO. Could it use a concrete monad? Sure.

```Haskell
newLocalRef :: J ty ->. Linear.IO (J ty, J ty)
```

But then, how do you combine it with other APIs which need different monads?

aspiwack:
Ok, I see the motivation. But I do object to `MonadIO` being defined using `MonadLift`, as argued above.

src/Control/Monad/IO/Class/Linear.hs
facundominguez: Let me see if I understand your proposal. `MonadLift` is ok as it is, but MonadIO should be made a type class that allows lifting Linear.IO into a linear monad.

```Haskell
class Linear.Monad m => MonadIO m where
  liftIO :: Linear.IO a ->. m a
```

That seems to work.

aspiwack: Indeed, but if we do that, I'm not sure we need `MonadLift` as an abstraction anymore.

facundominguez: `directory-server` is using `MonadLift` to run a MonadLogger. I don't think that would be possible with `MonadIO` alone.

facundominguez:
With the current state of this PR, the `jni` package had to define some auxiliary functions:
```
liftPreludeIO :: Linear.MonadIO m => Prelude.IO a -> m a
liftPreludeIO io = liftIO (Linear.fromSystemIO io)
  
liftPreludeIOU :: Linear.MonadIO m => Prelude.IO a -> m (Unrestricted a)
liftPreludeIOU io = liftIO (Linear.fromSystemIOU io)
```
This is because most functions in jni follow the pattern:
```
throw :: Linear.MonadIO m => J a #-> m (J a)
throw = Unsafe.toLinear $ \x -> liftPreludeIO (x Prelude.<$ JNI.throw x)
```
Note that
1. the function uses the linear arrow
2. It uses `Unsafe.toLinear` to get rid of the multiplicity check.
3. It uses `liftPreludeIO` to wrap the interesting action in the IO monad.

I'd expect `liftPreludeIO` and `liftPreludeIOU` to deserve a place in `linear-base`. Any thoughts?

src/Control/Monad/Lift/Linear.hs
```diff
@@ -0,0 +1,39 @@
...
+--
+-- > liftU (m >>= f) = liftU m >>= \(Unrestricted a) -> liftU (f a)
+--
+class (Linear.Monad m, NonLinear.Monad (LiftedM m)) => MonadLift m where
+
+  -- | The type of the monad whose computations are lifted
+  type LiftedM m :: * -> *
+
+  -- | Lift a nonlinear monadic computation into a linear monad
+  lift :: LiftedM m a -> m a
```
aspiwack, line 31: `lift` admits an implementation in terms of `liftU`, we ought to provide it as a default.

src/System/IO/Linear.hs
```diff
@@ -7,6 +7,7 @@
 {-# LANGUAGE RebindableSyntax #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE RoleAnnotations #-}
+{-# LANGUAGE TypeFamilies #-}
```
aspiwack, line 10: Looks like a rebase gone a bit awry.

src/Control/Monad/Lift/Linear.hs
```diff
@@ -0,0 +1,39 @@
...
+
+  -- | Like 'lift' but allows to use the result unrestrictedly
+  liftU :: LiftedM m a -> m (Unrestricted a)
```
aspiwack, line 34: Now that I thik of it, `liftU` also admits an implementation in terms of `lift`. So it ought to have a default implementation as well.

facundominguez:
> I'd make all 3 members of the MonadIO typeclass. With default implementations.

I'm fine with that.

> IO.lift.

I think `liftIO` is a better name. The unqualified `IO` refers to `System.IO.Linear.IO` in all of `linear-base`. It also hints strongly that it is related to `Control.Monad.IO.Class.liftIO`, which helps reading it.

> IO.liftSystem, and IO.liftSystemU

... therefore I'd prefer these named `liftSystemIO` and `liftSystemIOU`.

I can drop `MonadLift` in this PR, which defers the question of how to express constraints for a function that must do both logging and linear-monad operations.  I can live with this unresolved.

With `MonadLift` I could write
```
f :: (MonadLogger (LiftedM m), MonadLift m, Linear.Monad m, ClassWithLinearMethods m)  => m ()
```
I could imagine something simpler
```
f :: (MonadLogger m, Linear.Monad m, ClassWithLinearMethods m)  => m ()
```
but this probably will require to provide an instance of monad logger, rather than reusing a existing one.

utdemir (approved):
I am happy with these changes, especially since they're useful in `inline-java`. And they look simple.

@aspiwack , do you still have concerns about it? Is there anything we should do before we merge this in?

aspiwack (approved):
No, it's all fine. It's my entire fault for missing  the latest change. I'm sorry @facundominguez .

Let's merge this.

facundominguez:
Rebased master.

PR #65, opened 2019-08-28
Some basic documentation for arrays

b-mehta:
Add some documentation for destination arrays and polarized arrays.

Also remove some unnecessary pragmas and fix a loop in linear monoids.

### DIFF
--- a/src/Data/Array/Destination.hs
+++ b/src/Data/Array/Destination.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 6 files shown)


src/Data/Array/Polarized/Push.hs
```diff
@@ -19,6 +20,16 @@ import Prelude.Linear
 import qualified Prelude
 import Data.Monoid.Linear
 
+-- XXX: the below isn't really true yet since no friendly way of constructing a
+-- PushArray directly is given yet (see issue #62), but the spirit holds.
+-- XXX: There's also a slight lie in that one might one to consume a PushArray
```
monoidal: might want

src/Data/Array/Destination.hs
```diff
@@ -104,7 +105,9 @@ mirror as f
       fill (f x) dl `lseq` mirror xs f dr
 
 -- TODO: It would be better to implement this directly, instead of in terms of
--- mirror, to avoid allocation.
+-- mirror, to avoid allocation. This would also mean the length argument is
+-- redundant, as it should be.
+-- | Fill a destination array using the given function and length.
```
monoidal: After #66, there's no length argument anymore.

src/Data/Array/Polarized/Push.hs
b-mehta: Agreed, I'll rewrite these to be TODOs - it's not yet clear what a good way of doing this is.

monoidal: I meant "might one" -> "might want"

b-mehta: Ah. My mistake, thanks for this catch.

PR #68, opened 2019-08-31
Transformer instances

b-mehta:
Created `Control.Monad.Linear.Internal`, and moved the majority of `Control.Monad.Linear` there.
This makes it possible to put `<$` in `Control.Monad.Linear`, since there is a different data version which may also be useful:
```haskell
(<$) :: (Functor f, Consumable b) => a -> f b ->. f a
```
for instance to replace every element of a list, one can use `3 Data.<$ [True, False, True]`.
(@facundominguez this may mean your use of `<$` for Control Functors needs to be adjusted to Linear.<$, if you import Control.Monad.Linear as Linear).

Also adds Functor/Monad instances for some standard monad transformers from `transformers`, where appropriate, but the transformers are not re-exported.

### DIFF
--- a/src/Control/Monad/Linear/Internal.hs
+++ b/src/Control/Monad/Linear/Internal.hs
@@ -9,6 +9,8 @@ import Prelude.Linear.Internal.Simple
 import Prelude (String)
 import Data.Functor.Identity
 import qualified Data.Functor.Linear.Internal as Data
+import qualified Control.Monad.Trans.Reader as NonLinear
+import qualified Control.Monad.Trans.State.Strict as Strict
 
 -- $monad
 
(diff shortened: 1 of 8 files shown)


src/Prelude/Linear/Internal/Simple.hs
```diff
@@ -56,3 +56,6 @@ foldr :: (a ->. b ->. b) -> b ->. [a] ->. b
 foldr f z = \case
   [] -> z
   x:xs -> f x (foldr f z xs)
+
+lazyPair :: (x ->. y ->. z) ->. (x,y) ->. z
+lazyPair k = Unsafe.toLinear $ \v -> k (Prelude.fst v) (Prelude.snd v)
```
aspiwack: That looks dangerous. Let's just drop lazy state from our list of supported instances and avoid using `Unsafe.toLinear` gratuitously.

b-mehta: Hmm, okay. My hope was that since both components of the pair are consumed, this should indeed be linear.

aspiwack: Well, it may be considered linear. But it is very subtle: it is because of call-by-need/laziness that we can say that the spine of the pair is forced exactly once, so it's ok, somehow to traverse it twice (remember: there may be linear things in a thunk). Under a call-by-name semantics this is clearly not permitted.

Now, GHC _probably_ never unshares thunk, well, at least not if there is some real computation in them, _probably_. But it might. Which would probably make some things slower in non-linear code, but could possibly segfault in linear code. Do we want to take this risk?

The bottom line is that there is no plan, at the moment, to teach the type system (in particular the linter) about the interaction of sharing with linearity. And in this perspective, it's way to bold to add this `lazyPair` primitive.

b-mehta: Fair enough! Is there a plan to somehow allow lazy consumption of a pair in this sort of way?

aspiwack: Not at the moment. I don't even know how to approach it.

src/Control/Monad/Linear/Internal.hs
```diff
@@ -96,3 +98,38 @@ instance Functor f => Data.Functor (Data f) where
 instance Applicative f => Data.Applicative (Data f) where
   pure x = Data (pure x)
   Data f <*> Data x = Data (f <*> x)
+
+------------------------------------------------
+-- Instances for nonlinear monad transformers --
+------------------------------------------------
+
+-- XXX: Lazy state is not currently supported.
```
aspiwack: I'd not `XXX` this, as we have no plan to in the foreseeable future.

src/Data/Functor/Linear/Internal.hs
```diff
@@ -27,13 +32,20 @@ class Functor f where
...
 -- drops values, which we are not allowed to do in a linear container).
 --
 -- In fact, an applicative functor is precisely a functor equipped with (pure
 -- and) @liftA2 :: (a ->. b ->. c) -> f a ->. f b ->. f c@. In the case where
--- @f = []@, the signature of 'liftA2' specialises to that of 'zipWith'.
+-- @f = []@, the signature of 'liftA2' would specialise to that of 'zipWith'.
+--
+-- Intuitively, Data 'Applicative's can be seen as containers whose "number" of
+-- elements is known at compile-time. This includes vectors of known length
+-- but excludes Maybe, since this may contain either zero or one value.
```
aspiwack: Maybe -> 'Maybe'

PR #69, opened 2019-09-03
Linear num typeclasses

b-mehta:
A more fine grained linear num hierarchy.

### DIFF
--- a/src/Data/Num/Linear.hs
+++ b/src/Data/Num/Linear.hs
@@ -1,0 +1,126 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE StandaloneDeriving #-}
+
+module Data.Num.Linear
+  ( Additive(..)
+  , AddIdentity(..)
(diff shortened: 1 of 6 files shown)


src/Data/Num/Linear.hs
```diff
@@ -0,0 +1,126 @@
...
+-- laws: is an identity for *
+class Multiplicative a => MultIdentity a where
+  one :: a
+
+-- laws: distributivity, annihilation
+-- | Having a linear (*) means we can't short-circuit multiplication by zero
+class (AddIdentity a, MultIdentity a) => Semiring a where
+
+-- no additional laws
+class (Group a, MultIdentity a) => Ring a where
```
aspiwack: I believe you meant (up to the `AdditiveGroup` thing which I put here for consistency with the above suggestion)

```suggestion
class (AdditiveGroup a, SemingRing a) => Ring a where
```

src/Data/Num/Linear.hs
```diff
@@ -0,0 +1,126 @@
...
+-- laws: associative, commutative
+class Additive a where
+  (+) :: a ->. a ->. a
+
+-- laws: is an identity for +
+class Additive a => AddIdentity a where
+  zero :: a
+
+-- usual abelian group laws
+class AddIdentity a => Group a where
```
aspiwack:
```suggestion
class AddIdentity a => AdditiveGroup a where
```

I suppose. `Group` rather belongs to the `Monoid` hierarchy.

PR #70, opened 2019-09-05
Linear monad transformers

b-mehta:
Adds linear reader and state monad transformers, and a monad transformer class.

### DIFF
--- a/src/Control/Monad/Linear.hs
+++ b/src/Control/Monad/Linear.hs
@@ -17,10 +17,88 @@ module Control.Monad.Linear
   , ap
   , Data(..)
   , foldM
+  -- * Monad transformers
+  -- ** ReaderT monad transformer
+  , Reader, reader, runReader, mapReader, withReader
+  , ReaderT(..), runReaderT, mapReaderT, withReaderT
+  , ask, local, asks
+  -- ** StateT monad
+  , State, state, runState, execState, mapState, withState
+  , StateT(..), runStateT, execStateT, mapStateT, withStateT
+  , get, put, modify, gets
+  , MonadTrans(..)
   ) where
 
 import Control.Monad.Linear.Internal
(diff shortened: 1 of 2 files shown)


src/Control/Monad/Linear.hs
```diff
@@ -17,10 +17,90 @@ module Control.Monad.Linear
...
+
+
+get :: (Applicative m, Dupable s) => StateT s m s
+get = state dup
+
+put :: (Applicative m, Consumable s) => s ->. StateT s m ()
+put = void . replace
+
+-- XXX: If multiplicity 0 is added, we can have
+-- gets :: Applicative m => (s # 0 -> a) ->. StateT s m a
```
aspiwack: Though that is an incomparable `gets`. So which one is best? It probably shouldn't be in an `XXX` unless we truly have an intention to change it.

src/Control/Monad/Linear/Internal.hs
```diff
@@ -129,5 +132,85 @@ instance Monad m => Monad (NonLinear.ReaderT r m) where
...
+-- | @replace s@ will replace the current state with the new given state, and
+-- return the old state.
+replace :: Applicative m => s ->. StateT s m s
+replace s = state $ (\s' -> (s', s))
+
+-----------------------------
+-- Monad transformer class --
+-----------------------------
+
+class (forall m. Monad m => Monad (t m)) => MonadTrans t where
```
aspiwack: Yay :slightly_smiling_face: !

src/Control/Monad/Linear.hs
b-mehta: Fair enough, I'll remove the comment for now.

aspiwack:
Remember: approved means that you can merge when you see fit. Unless you want a second round of review.

PR #72, opened 2019-09-09
Some typo fixes

b-mehta:
Fixed some typos, and added that Applicatives should be _non-empty_ fixed size containers.

### DIFF
--- a/src/Data/Functor/Linear/Internal.hs
+++ b/src/Data/Functor/Linear/Internal.hs
@@ -33,19 +33,19 @@ class Functor f where
 
 -- | Data 'Applicative'-s can be seen as containers which can be zipped
 -- together. A prime example of data 'Applicative' are vectors of known length
--- ('ZipLists' would be, if it were not for the fact that zipping them together
+-- ('ZipList's would be, if it were not for the fact that zipping them together
 -- drops values, which we are not allowed to do in a linear container).
 --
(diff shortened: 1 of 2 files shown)


src/Data/Functor/Linear/Internal.hs
```diff
@@ -33,19 +33,21 @@ class Functor f where
...
--- infinitary) container indexed by @r@, while lists do not, since they may
--- contain any number of elements.
+-- Intuitively, the type of 'liftA2' means that 'Applicative's can be seen as
+-- containers whose "number" of elements is known at compile-time. This
+-- includes vectors of known length but excludes 'Maybe', since this may
+-- contain either zero or one value.  Similarly, @((->) r)@ forms a Data
+-- 'Applicative', since this is a (possibly infinitary) container indexed by
+-- @r@, while lists do not, since they may contain any number of elements.
+-- On the other hand, the type of 'pure' enforces that this "number" must be
+-- at least one, and so the container cannot be empty.
```
aspiwack: Pure is an unrestricted function. I'm pretty sure `Const a` is a `Data.Applicative` (if `a` is a (linear) `Monoid`).

b-mehta: Yup, my mistake. I was confusing "`f ()` is non-empty" with the container being non-empty.

PR #73, opened 2019-09-09
Write mirror in terms of fromFunction

b-mehta:
As suggested in #66

### DIFF
--- a/src/Data/Array/Destination.hs
+++ b/src/Data/Array/Destination.hs
@@ -14,7 +14,7 @@ module Data.Array.Destination
   where
 
 import Control.Exception (evaluate)
-import Data.Vector (Vector)
+import Data.Vector (Vector, (!))
 import qualified Data.Vector as Vector
 import Data.Vector.Mutable (MVector)
 import qualified Data.Vector.Mutable as MVector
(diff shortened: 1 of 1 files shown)


src/Data/Array/Destination.hs
```diff
@@ -82,18 +82,8 @@ split n = Unsafe.toLinear unsafeSplit
...
-      -- has no holes to fill so we can safely linearly consume it
-  | otherwise =
-      mirrorHeadAndTail (Vector.head as) (Vector.tail as) . split 1
-  where
-    -- /!\ The first DArray has length 1.
-    mirrorHeadAndTail :: a -> Vector a -> (DArray b, DArray b) ->. ()
-    mirrorHeadAndTail x xs (dl, dr) =
-      fill (f x) dl `lseq` mirror xs f dr
+mirror :: Vector a -> (a ->. b) -> DArray b ->. ()
+mirror v f = fromFunction (\t -> f (v ! t))
```
aspiwack: Just for the records (it's not a suggestion!). Did you know that `\t -> f (v ! t)` can also be written `f . (v !)`?

b-mehta: Yeah! I considered that but I think it's harder to read.

PR #74, opened 2019-09-09
Some cleanups in vectors of known length

b-mehta:
Avoid some uses of `Unsafe.coerce`, and re-organise others.

### DIFF
--- a/src/Data/Vector/Linear.hs
+++ b/src/Data/Vector/Linear.hs
@@ -67,12 +67,12 @@ theLength :: forall n. KnownNat n => Int
 theLength = fromIntegral (natVal' @n (proxy# @_))
 
 instance Data.Functor (V n) where
-  fmap f (V xs) = V $ Unsafe.toLinear ((Unsafe.coerce Vector.map) f) xs
+  fmap f (V xs) = V $ Unsafe.toLinear (Vector.map (\x -> f x)) xs
 
 instance KnownNat n => Data.Applicative (V n) where
   pure a = V $ Vector.replicate (theLength @n) a
   (V fs) <*> (V xs) = V $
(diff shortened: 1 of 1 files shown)


src/Data/Vector/Linear.hs
```diff
@@ -64,15 +63,15 @@ newtype V (n :: Nat) (a :: *) = V (Vector a)
   -- Vectors at the moment.
 
 theLength :: forall n. KnownNat n => Int
-theLength = fromIntegral (natVal' @n (proxy# @_))
+theLength = fromIntegral (natVal @n Proxy)
```
aspiwack: What's the motivation for that? I used proxy to make sure that there was no value at run time.

src/Data/Vector/Linear.hs
```diff
@@ -64,15 +63,15 @@ newtype V (n :: Nat) (a :: *) = V (Vector a)
...
 
 instance Data.Functor (V n) where
-  fmap f (V xs) = V $ Unsafe.toLinear ((Unsafe.coerce Vector.map) f) xs
+  fmap f (V xs) = V $ Unsafe.toLinear (Vector.map (\x -> f x)) xs
 
 instance KnownNat n => Data.Applicative (V n) where
   pure a = V $ Vector.replicate (theLength @n) a
   (V fs) <*> (V xs) = V $
-    Unsafe.toLinear2 (Vector.zipWith @(_ ->. _) (\f x -> f x)) fs xs
+    Unsafe.toLinear2 (Vector.zipWith (\f x -> f $ x)) fs xs
```
aspiwack: Cute :slightly_smiling_face: .

src/Data/Vector/Linear.hs
b-mehta: I thought it was clearer at no real performance cost, but I'll revert this change.

PR #76, opened 2019-09-10
Weaken constraint for lengthOf

b-mehta:
Give a weaker constraint for `lengthOf`.

### DIFF
--- a/src/Control/Optics/Linear/Internal.hs
+++ b/src/Control/Optics/Linear/Internal.hs
@@ -37,7 +37,6 @@ module Control.Optics.Linear.Internal
 import qualified Control.Arrow as NonLinear
 import qualified Data.Bifunctor.Linear as Bifunctor
 import Data.Bifunctor.Linear (SymmetricMonoidal)
-import Data.Monoid
 import Data.Profunctor.Linear
 import Data.Functor.Linear
 import qualified Data.Profunctor.Kleisli.Linear as Linear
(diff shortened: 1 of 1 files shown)


src/Control/Optics/Linear/Internal.hs
```diff
@@ -123,9 +122,8 @@ build (Optical l) x = Linear.runCoKleisli (l (Linear.CoKleisli getConst')) (Cons
 getConst' :: Const a b ->. a
 getConst' (Const x) = x
 
--- TODO: think about if this is the right constraint
-lengthOf :: P.Num r => Optic_ (NonLinear.Kleisli (Const (Sum r))) a b s t -> s -> r
-lengthOf l s = getSum (gets l (const (Sum 1)) s)
+lengthOf :: MultIdentity r => Optic_ (NonLinear.Kleisli (Const (Adding r))) a b s t -> s -> r
```
aspiwack: I'm confused. Shoudn't it be `AddIdentity`? How does it even type?

b-mehta: Surprisingly not, the `MultIdentity` is only necessary so I can say `one`, there's no actual requirement here that `Adding r` forms a `Monoid`. But of course whenever this is used, `AddIdentity r` is also required.

aspiwack: Oh. I see. Because if it's a lens, then you just need `1`. If you are looking at a traversal, then you will have the additional requirement that `NonLinear.Kleisli (Const (Sum r))` is an applicative, which, in turn, requires that `Sum r` is a monoid, and that `r` is `AddIdentity`.

Neat.

b-mehta: Right, pretty much.

PR #79, opened 2019-09-12
Lots of optics

b-mehta:
Draft: 

- It's unclear how efficient Batch is. It's also unclear if it (and its applicative instance) are defined in the correct way: it might allocate less if the order in `:*:` is flipped, and the `<*>` definition adjusted accordingly.
- Some of `set, set', set''`  might not be useful.
- The `withXXX` functions aren't consistent in their return type.
- There are some very bad names, eg `MyFunctor` and `inn`.


b-mehta:
I've modified the `Batch` type, and it is more efficient now.

aspiwack:
I've pushed most of this PR to smaller PRs (#194 , #195, #196 ).

What's left is
- The `set''` primitive, see #197 .
- The traversal story.

I'm leaving the PR open, still, as an inspiration as I'm trying to tackle the traversal story. See also #190 (and soon more).

PR #84, opened 2020-04-02
Merge linear containers

Divesh-Otwani:
Adding stuff from linear-containers except for the examples.

To avoid duplication, there's a new module: Data.Linear.Internal.MutableArray.


src/Data/Linear/Internal/MutableArray.hs
```diff
@@ -0,0 +1,46 @@
...
+
+unsafeNewMutArr :: Int -> a -> MutableArray# RealWorld a
+unsafeNewMutArr (I# size) init =
+  case newArray of
+    (# _, array #) -> array
+  where
+    newArray = runRW# $ \stateRW -> newArray# size init stateRW 
+
+
+-- | writeMutArr should have evaluation forced
```
aspiwack, line 32: I don't understand what this means.

src/Data/Linear/Internal/MutableArray.hs
```diff
@@ -0,0 +1,46 @@
...
+writeMutArr :: MutableArray# RealWorld a -> Int -> a -> ()
+writeMutArr mutArr (I# ix) val =
```
aspiwack, line 34: This is an unsafe function, it should be in the name somehow. A good way would be to have `Unsafe` as the module name. Like `Data.Linear.Array.Unsafe`. So as to avoid repeating it for every function. There is already such an `Unsafe` module in `linear-base`.

src/Data/Linear/Vector.hs
```diff
@@ -0,0 +1,112 @@
...
+-- | Write to an element already written to before.
+-- Note: this will not write to elements beyond the current
+-- length of the array.
+write :: Vector a #-> Int -> a -> Vector a
+write = Linear.toLinear writeUnsafe
+  where
+    writeUnsafe :: Vector a -> Int -> a -> Vector a
+    writeUnsafe v@(Vec (len, size) mutArr) ix val
+      | indexInRange len ix = case writeMutArr mutArr ix val of () -> v
+      | otherwise = error "Index not in range"
```
aspiwack, line 73: I see that you haven't put the `HasCallStack` constraints yet. Let's do that next.

aspiwack:
Aside: I wish you would have made the merge and the refactor in two different PRs. This way it's really hard for me to gauge the impact of the refactor.

PS: remember to clean your history before merge.

Divesh-Otwani:
@aspiwack will certainly do that going forward

src/Unsafe/MutableArray.hs
```diff
@@ -5,29 +5,24 @@
 -- |
 -- Module: MutableArray
 -- Description: Unsafe wrappers around `MutableArray#`s
-
-module Data.Linear.Internal.MutableArray
-  ( unsafeNewMutArr
-  , readMutArr
-  , writeMutArr
+module Unsafe.MutableArray
```
aspiwack, line 8: We may want to mark it as internal nonetheless. But it doesn't need to be done now.

src/Unsafe/MutableArray.hs
```diff
@@ -0,0 +1,38 @@
...
+----------------------------
+
+newMutArr :: Int -> a -> MutableArray# RealWorld a
+newMutArr (I# size) x =
+  case newArray of
+    (# _, array #) -> array
+  where
+    newArray = runRW# $ \stateRW -> newArray# size x stateRW
+
+-- | writeMutArr should have evaluation forced
```
aspiwack, line 27: I still don't understand what this means.

Divesh-Otwani: Forgot to change it. Will when I merge.

PR #87, opened 2020-04-06
Containers/hashtable robinhood

Divesh-Otwani:
Here is an implementation of robin hashmaps. It turned out to be somewhat tricky and I hope I've got it right, especially with maintaining the invariants and the resizing operation.

I absolutely need to write a nice test suite with glass-box and black-box tests covering interesting cases. Also, stress tests would be good.

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -51,8 +51,8 @@ length = Linear.toLinear unsafeLength
     unsafeLength :: Array a -> (Int, Array a)
     unsafeLength v@(Array size _) = (size, v)
 
-write :: HasCallStack => Array a #-> Int -> a #-> Array a
-write = Linear.coerce writeUnsafe
+write :: HasCallStack => Array a #-> Int -> a -> Array a
+write = Linear.toLinear writeUnsafe
   where
     writeUnsafe :: Array a -> Int -> a -> Array a
     writeUnsafe arr@(Array size mutArr) ix val
@@ -61,8 +61,8 @@ write = Linear.coerce writeUnsafe
           () -> arr
       | otherwise = error "Write index out of bounds."
 
-read :: HasCallStack => Array a #-> Int #-> (Array a, a)
-read = Linear.coerce readUnsafe
+read :: HasCallStack => Array a #-> Int -> (Array a, a)
+read = Linear.toLinear readUnsafe
   where
     readUnsafe :: Array a -> Int -> (Array a, a)
     readUnsafe arr@(Array size mutArr) ix
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -50,8 +50,8 @@ length = Linear.toLinear unsafeLength
     unsafeLength :: Array a -> (Int, Array a)
     unsafeLength v@(Array size _) = (size, v)
 
-write :: HasCallStack => Array a #-> Int -> a -> Array a
-write = Linear.toLinear writeUnsafe
+write :: HasCallStack => Array a #-> Int -> a #-> Array a
+write = Linear.coerce writeUnsafe
   where
     writeUnsafe :: Array a -> Int -> a -> Array a
     writeUnsafe arr@(Array size mutArr) ix val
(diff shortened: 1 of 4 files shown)


src/Data/HashMap/Linear.hs
```diff
@@ -1,73 +1,239 @@
...
+    tripleArr arr = case read arr 0 of
+      (arr', (k,v,_)) ->
+        moveHanging size (resizeSeed (size*3) (k,v,-1) arr', 0)
+
+    -- | After copying over to an array trice the original size,
+    -- we need to move all "hanging elements", i.e., those triples (k,v,psl)
+    -- where the psl is greater than the index of the cell.
+    -- Theorem. The hanging elements of a robin array form a prefix.
+    -- The proof is left as an exercise to the reader.
+    -- Arguments: `adjustModedPrefix original-size (arr,ix)`
```
Divesh-Otwani, line 178: This is a typo. The arguments line should use `moveHanging` and not `adjustModedPrefix`

src/Data/Array/Mutable/Linear.hs
```diff
@@ -50,28 +51,33 @@ length = Linear.toLinear unsafeLength
...
     writeUnsafe arr@(Array size mutArr) ix val
       | indexInRange size ix =
         case writeMutArr mutArr ix val of
           () -> arr
-      | otherwise = error "Index not in range"
+      | otherwise = error "Write index out of bounds."
 
-read :: HasCallStack => Array a #-> Int -> (Array a, Maybe a)
-read = Linear.toLinear readUnsafe
+read :: HasCallStack => Array a #-> Int #-> (Array a, a)
```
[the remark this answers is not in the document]
Divesh-Otwani: This has a problem similar to that of `write` -- a linear use of the read index

src/Data/HashMap/Linear.hs
```diff
@@ -1,73 +1,239 @@
...
+-- | A probe sequence length
+newtype PSL = PSL Int
+  deriving (Prelude.Eq, Prelude.Ord, Prelude.Num)
+
+-- | An array of Robin values
+type RobinArr k v = Array (RobinVal k v)
 
+-- | At minimum, we need to store hashable
+-- and identifiable keys
+type EqHashable k = (Eq k, Hashable k)
```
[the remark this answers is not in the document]
Divesh-Otwani: I changed it to `Keyable k` which spares 7 characters per line. Yes, it does add an indirection but my instincts tell me the code looks messy with it expanded. I'll change it on a re-review if you really think it's not worth it.

src/Data/HashMap/Linear.hs
```diff
@@ -1,73 +1,239 @@
...
+  -- | A key, PSL pair for a swap at an index.
+  -- The swapped-out pair will then need to be inserted downstream.
+  IndexToSwap :: (k, PSL) -> Int -> RobinQuery k
 
 -- # Construction and Modification
 --------------------------------------------------
 
-singleton :: Hashable k =>
+-- | Run a computation using a singleton hashmap
+singleton :: Keyable k =>
```
aspiwack, line 92: We should have curried `singleton, but we can do that in another PR.

src/Data/HashMap/Linear.hs
```diff
@@ -1,73 +1,239 @@
...
   (k,v) -> (HashMap k v #-> Unrestricted b) -> Unrestricted b
-singleton = undefined
-
-alter ::  Hashable k => (Maybe a -> Maybe a) -> k -> HashMap k v #-> HashMap k v
-alter = undefined
-
-insert :: Hashable k => HashMap k v #-> k -> v -> HashMap k v
-insert hmap k v = alter (\_ -> Just v) k hmap
+singleton (k :: k, v :: v) (f :: HashMap k v #-> Unrestricted b) =
```
aspiwack, line 94: By the way, since you are not using `ScopedTypeVariables`, these type annotation don't have the meaning you think they have.

Divesh-Otwani: I am using `ScopedTypeVariables`.

aspiwack: I'm leaving it as a puzzle to you to solve, how you are not making use of it enough in this particular case.

It shouldn't prevent you from merging this, of course.

PR #91, opened 2020-04-15
Array test suite

### LINKED ISSUE #90: Mutable Vectors replicate Mutable Arrays
Right now, mutable vectors and mutable arrays both use Unsafe.MutableArray. Most of what Vector does repeats code from mutable arrays.

This should be fixed, probably by implementing mutable vectors in terms of mutable arrays, or just by having one array type with the functionality of both.

Divesh-Otwani:
This is the array & hashmap sub-library with a passing test suite. There were some fun bugs to work out with robin hashing, but it's all over now :)

I didn't add tests for vectors because they just replicate the array tests. I'll add those when I fix 
#90. There are some minor changes to linear-base that I had to make along the way.

(This isn't a small PR, though it's still coherent -- I probably should have broken this up into a PR for array tests and one for the hashmap tests.)


src/Data/Array/Mutable/Linear.hs
```diff
@@ -69,17 +75,52 @@ read = Linear.toLinear readUnsafe
...
+    where
+      showArr n size mutArray
+        | n == 0 =
+          "\n[\n" ++ show (readMutArr mutArray 0) ++ "\n"
+            ++ showArr (n + 1) size mutArray
+        | (n + 1) == size =
+          show (readMutArr mutArray n) ++ "\n]\n"
+        | otherwise =
+          show (readMutArr mutArray n) ++ "\n"
+            ++ showArr (n + 1) size mutArray
```
aspiwack, line 116:
```suggestion
  show (Array size mutArray) = "\n[\n" ++ showArr 0 size mutArray ++ "]\n"
    where
      showArr n size mutArray
        | n == size = ""
        | otherwise = show (readMutArr mutArray n) ++ "\n" ++ showArr (n + 1) size mutArray
```

test/Test/Data/Mutable/Array.hs
```diff
@@ -1,16 +1,112 @@
...
 
-group :: Group
-group = Group "Mutable array tests" $
-  []
+mutArrTests :: TestTree
+mutArrTests = testGroup "Mutable array tests" group
 
+group :: [TestTree]
+group =
+  [ testProperty "[read i (write arr i x) = x]" testWrite,
```
aspiwack, line 33: There are very few tests. We could test that `read i (write arr j x) = read arr i` when `(j <> i)` (there are challenges due to linearity, but we can easily generate the same array twice).

We could test that reading twice at the same index gives the same result.

Maybe we need a test relating `length` and `resize`.

Etc…

test/Test/Data/Mutable/Array.hs
```diff
@@ -1,16 +1,112 @@
...
+  ( mutArrTests,
+  )
+where
 
+import qualified Data.Array.Mutable.Linear as Arr
+import Data.Unrestricted.Linear
 import Hedgehog
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+import qualified Prelude.Linear as Lin
```
aspiwack, line 20:
```suggestion
import qualified Prelude.Linear as Linear
```

test/Test/Data/Mutable/Array.hs
```diff
@@ -1,16 +1,112 @@
...
+-- Tests for mutable arrays.
+--
+-- TODO:
+--  * Test failures for out of bound access
+module Test.Data.Mutable.Array
+  ( mutArrTests,
+  )
+where
 
+import qualified Data.Array.Mutable.Linear as Arr
```
aspiwack, line 15:
```suggestion
import qualified Data.Array.Mutable.Linear as Array
```

test/Test/Data/Mutable/Array.hs
```diff
@@ -1,16 +1,112 @@
...
-group :: Group
-group = Group "Mutable array tests" $
-  []
+mutArrTests :: TestTree
+mutArrTests = testGroup "Mutable array tests" group
 
+group :: [TestTree]
+group =
+  [ testProperty "[read i (write arr i x) = x]" testWrite,
+    testProperty "[alloc creates constant arrays]" testAlloc,
```
aspiwack, line 34: Also these brackets are pure boilerplate. Let's drop them.

test/Test/Data/Mutable/Array.hs
```diff
@@ -1,16 +1,112 @@
...
+makeArray ((ix, val) : xs) arr = makeArray xs Lin.$ Arr.write arr ix val
+
+-- | Given an array size, create random values to insert
+randomArrVals :: Int -> Gen [(Int, Int)]
+randomArrVals size = do
+  numInsert <- Gen.int $ Range.constant 150 500
+  let numToInsert = Range.singleton numInsert
+  let genVal = Gen.int $ Range.constant 1 300
+  let genIx = Gen.int $ Range.constant 0 (size -1)
+  ixs <- Gen.list numToInsert genIx
```
aspiwack, line 55: This is probably not a good idea, just use `[0..size-1]`: what you want is a random value per cell, that's all.

test/Test/Data/Mutable/Array.hs
```diff
@@ -1,16 +1,112 @@
...
+makeArray :: [(Int, Int)] -> Arr.Array Int #-> Arr.Array Int
+makeArray [] arr = arr
+makeArray ((ix, val) : xs) arr = makeArray xs Lin.$ Arr.write arr ix val
+
+-- | Given an array size, create random values to insert
+randomArrVals :: Int -> Gen [(Int, Int)]
+randomArrVals size = do
+  numInsert <- Gen.int $ Range.constant 150 500
+  let numToInsert = Range.singleton numInsert
+  let genVal = Gen.int $ Range.constant 1 300
```
aspiwack, line 53: Probably prefer
```suggestion
  let genVal = Gen.int $ Range.linear 1 300
```

test/Test/Data/Mutable/Array.hs
```diff
@@ -1,16 +1,112 @@
...
+--------------------------------------------------------------------------------
+
+-- The read of a cell we wrote to should be what we wrote.
+testWrite :: Property
+testWrite = property $ do
+  arrSize <- forAll $ Gen.int (Range.constant 1 100)
+  ix <- forAll $ Gen.int $ Range.constant 0 (arrSize - 1)
+  val <- forAll $ Gen.int $ Range.constant 0 100
+  randInserts <- forAll $ randomArrVals arrSize
+  let test = testWriteLin ix val #. makeArray randInserts
```
aspiwack, line 69: Give yourself a `fromList` function with the same sort of signature as `alloc`. It will make your life simpler. Probably in the `Array` module.

test/Test/Data/Mutable/Array.hs
```diff
@@ -1,16 +1,112 @@
...
+group =
+  [ testProperty "[read i (write arr i x) = x]" testWrite,
+    testProperty "[alloc creates constant arrays]" testAlloc,
+    testProperty "[resize works correctly]" testResize
+  ]
 
+-- # Internal Library
+--------------------------------------------------------------------------------
 
+type ArrayTester a = Arr.Array a #-> Unrestricted Bool
```
aspiwack, line 41: If you were to stick to bool, then you wouldn't need `Unrestricted`, since `Bool` is movable.

However, it's probably not the best idea, better to move to `Unrestricted (m ())` to make it possible for you to use `===`, which will give you better error message in case the test fails.

Quite frankly, I don't think this type synonym is particularly useful. You should probably just inline it.

test/Test/Data/Mutable/HashMap.hs
```diff
@@ -1,16 +1,228 @@
...
 
-group :: Group
-group = Group "Mutable hashmap tests" $
-  []
+mutHMTests :: TestTree
+mutHMTests = testGroup "Mutable hashmap tests" group
 
+group :: [TestTree]
+group =
+  [ testProperty "[singletons work correctly]" testSingleton,
```
aspiwack, line 33: Again, what does « correctly » means?

test/Test/Data/Mutable/HashMap.hs
```diff
@@ -1,16 +1,228 @@
...
+    testProperty "[lookup after insert]" insertLookup,
```
aspiwack, line 34: There are two possible tests here: lookup of what you've inserted, and lookup of what you haven't inserted. Both tests are useful.

test/Test/Data/Mutable/HashMap.hs
```diff
@@ -1,16 +1,228 @@
...
+    testProperty "[insert following delete]" insertAfterDelete,
```
aspiwack, line 35: I don't know what this tests.

test/Test/Data/Mutable/HashMap.hs
```diff
@@ -1,16 +1,228 @@
...
+    testProperty "[insert ensures key membership]" insertMember,
+    testProperty "[insert new key increments size]" insertSize,
+    testProperty "[delete following insert]" deleteAfterInsert,
```
aspiwack, line 38: I don't know what this tests.

test/Test/Data/Mutable/Vector.hs
```diff
@@ -1,16 +1,78 @@
...
-mutVecTests = checkParallel group
+mutVecTests :: TestTree
+mutVecTests = testGroup "Mutable vector tests" group
 
-group :: Group
-group = Group "Mutable vector tests" $
-  []
+group :: [TestTree]
+group =
+  [testProperty "[snoc on vecs]" testSnoc]
```
aspiwack, line 32: Another test that I don't know what it tests.

test/Test/Util.hs
```diff
@@ -0,0 +1,16 @@
+{-# LANGUAGE LinearTypes #-}
+
+module Test.Util where
```
aspiwack, line 3: We frown upon `Util` modules. Let's not.

test/Test/Util.hs
```diff
@@ -0,0 +1,16 @@
...
+
+import Data.Unrestricted.Linear
+import qualified Prelude.Linear as Lin
+
+-- XXX: This should be in linear base
+-- This is also a terrible name
+getSnd :: (Consumable a, Movable b) => (a, b) #-> Unrestricted b
+getSnd (a, b) = lseq a (move b)
```
aspiwack, line 11: This is too bespoke to be anywhere. The main reason why you need it is because the `read` functions currently omits to return an `Unrestricted` value. Also because we are lacking proper support for linear cases, I'm afraid, but that's not on you.

test/Test/Util.hs
```diff
@@ -0,0 +1,16 @@
...
+
+(#.) :: (b #-> c) #-> (a #-> b) #-> (a #-> c)
+(#.) = (Lin..)
+
+infixr 5 #.
```
aspiwack, line 16: This already exists as `(.)` in the linear prelude.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -11,39 +13,43 @@
...
+import GHC.Exts hiding (toList)
 import GHC.Stack
 import qualified Unsafe.Linear as Linear
-import Prelude hiding (length, read)
 import Unsafe.MutableArray
+import Prelude hiding (length, read)
 
 -- # Data types
-----------------------------
+-------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
Divesh-Otwani: @aspiwack In general, if I have some trivial formatting I want to do, how should I go about making those changes?

src/Data/Array/Mutable/Linear.hs
```diff
@@ -69,17 +75,52 @@ read = Linear.toLinear readUnsafe
...
+            ++ showArr (n + 1) size mutArray
+        | (n + 1) == size =
+          show (readMutArr mutArray n) ++ "\n]\n"
+        | otherwise =
+          show (readMutArr mutArray n) ++ "\n"
+            ++ showArr (n + 1) size mutArray
+
+instance Consumable (Array a) where
+  consume :: Array a #-> ()
+  consume = Linear.toLinear (const ())
```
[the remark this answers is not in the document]
Divesh-Otwani: Thanks -- it's important that we get that right and that I learn this.

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -95,11 +98,17 @@ read = Linear.toLinear readUnsafe
...
 -- # Internal library
--------------------------------------------------------
+-------------------------------------------------------------------------------
 defaultSize :: Int
 defaultSize = 16
 
 -- | Argument order: indexInRange len ix
 indexInRange :: Int -> Int -> Bool
-indexInRange size ix = 0 <= ix && ix < size
+indexInRange size ix = (ix < size) && (0 <= ix)
```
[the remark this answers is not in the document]
Divesh-Otwani: Yes, this was a pointless change.

src/Data/HashMap/Linear.hs
```diff
@@ -113,87 +119,89 @@ alter = Unsafe.toLinear unsafeAlter
...
 -- | If the load is too high resize by tripling the array.
+-- XXX: This is ugly, but algorithmically simple and correct.
 maybeResize :: Keyed k => HashMap k v #-> HashMap k v
-maybeResize (HashMap (size, count) arr)
-   | count*3 < size = HashMap (size, count) arr
-   | otherwise = HashMap (size*3, count) (Unsafe.toLinear tripleArr arr)
+maybeResize (HashMap (!size, !count) arr)
+  | count * 3 < size = HashMap (size, count) arr
+  | otherwise =
+    (Unsafe.toLinear withAssocList) $
```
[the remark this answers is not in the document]
Divesh-Otwani: It's on a todo list at the top. I'll make this an issue in Github.

test/Test/Data/Mutable/Array.hs
Divesh-Otwani: Yes, I'll add a lot more tests and a thorough reasoned scheme to covering everything important.

test/Test/Data/Mutable/Array.hs
Divesh-Otwani: I think that adds verbosity -- I think Lin is clear and succinct.

test/Test/Data/Mutable/Array.hs
Divesh-Otwani: I think Arr is a common and clear abbreviation.

src/Data/Array/Mutable/Linear.hs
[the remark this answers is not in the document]
aspiwack: Make a separate PR, as much as possible.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -11,39 +13,43 @@
...
 -- # Creation
-----------------------------
+-------------------------------------------------------------------------------
 
 -- | Allocate a constant array given a size and an initial value
 alloc :: Int -> a -> (Array a #-> Unrestricted b) -> Unrestricted b
-alloc size x f = f (Array size (newMutArr size x))
+alloc size x f
+  | size > 0 = f (Array size (newMutArr size x))
+  | otherwise = error $ "Trying to allocate an array of size " ++ show size
```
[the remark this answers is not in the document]
aspiwack: Cf the blog post now that it's out :slightly_smiling_face:

test/Test/Data/Mutable/Array.hs
aspiwack: To be clear: I'm not offering this one up for negotiation.

test/Test/Data/Mutable/Array.hs
aspiwack: Nor this one.

test/Test/Data/Mutable/Array.hs
Divesh-Otwani: In the future, I'll implement all suggested changes in a review of one of my PRs :)

Also, reconsidering it, shorter qualified names can get terrible if there are a lot of them.

Divesh-Otwani:
Hi @aspiwack --

- I made all the changes above. 
- I changed `ArrayTester` and `VectorTester` to a type that I think makes sense for a type synonym: `Array a #-> Unrestricted (TestT IO ())`. I know `Bool` is moveable, but this seems like something I could abstract. Should I inline it? On that note, should I inline `HMTest`?
- The HashMap tests could be made as elegant as the array ones, which I will do in response to #93. I'll have the continuations be of type `HashMap k v #-> Unrestricted (TestT IO ())` .
- I tried to make the approach of testing these much more systematic and kind-of based on rewrite rules that would roughly define the semantics of these structures.
- Though I squashed the commits, there are trivial merge conflicts. I'll eventually figure out a way to do merges that don't have this issue.
- Along the way, I had to make some tiny changes to the data structures.
- I haven't run things through ormolu because I want to do that in a separate PR to keep things clean.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -11,75 +15,123 @@
...
 
--- Resize with a seed value for the newly allocated elements
+-- | Resize to a new constant array given a seed value
 resizeSeed :: HasCallStack => Int -> a -> Array a #-> Array a
-resizeSeed newSize seed (Array _ mutArr) =
-  Array newSize (resizeMutArrSeed mutArr seed newSize)
+resizeSeed newSize seed (Array _ _) =
+  Array newSize (Unsafe.newMutArr newSize seed)
+
+-- XXX: Replace with toVec
```
aspiwack, line 110: I mean, it's a fine function to have. Having an `O(1)` freeze would be nice too, of course.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -11,75 +15,123 @@
...
+
+-- # Instances
+-------------------------------------------------------------------------------
+
+instance Show a => Show (Array a) where
+  show = show . snd . (\x -> toList x)
+
+instance Consumable (Array a) where
+  consume :: Array a #-> ()
+  consume (!arr) = (Unsafe.toLinear (\_ -> ())) arr
```
aspiwack, line 131: In fact, you can:
```suggestion
  consume (Array _ _) = ()
```
It's probably better.

src/Data/HashMap/Linear.hs
```diff
@@ -205,29 +213,36 @@ lookup hmap k = (Unsafe.toLinear lookupFromIx) $ queryIndex hmap k
...
-      _ ->
-        walkDownArr (arr, (ix + 1) `mod` size) (key, psl + PSL 1) size
+      _ -> walkDownArr (arr, (ix + 1) `mod` size) (key, psl + PSL 1) size
+
+-- # Instances
+--------------------------------------------------
+
+instance Consumable (HashMap k v) where
+  consume :: HashMap k v #-> ()
+  consume (HashMap !_ !arr) = lseq arr ()
```
aspiwack, line 241: You don't need the `!` patterns on the subcomponents:
- The first one is not linear
- The array is linear, but you will consume it anyway, nothing unsafe, so you can follow the types.

```suggestion
  consume (HashMap _ arr) = consume arr
```

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -14,91 +15,108 @@
...
     readUnsafe v@(Vec (len, _) mutArr) ix
-      | indexInRange len ix = (v, readMutArr mutArr ix)
+      | indexInRange len ix = (v, Unsafe.readMutArr mutArr ix)
       | otherwise = error "Read index not in range."
 
+-- # Instances
+-------------------------------------------------------------------------------
+
+instance Consumable (Vector a) where
+  consume (!vec) = Unsafe.toLinear (\_ -> ()) vec
```
aspiwack, line 111:
```suggestion
  consume (Vec _ _) = ()
```

test/Test/Data/Mutable/Array.hs
```diff
@@ -0,0 +1,230 @@
...
+
+randomSize :: Gen Int
+randomSize =  Gen.int (Range.constant 100 1000)
+
+randomVal :: Gen Int
+randomVal = Gen.int (Range.linear 1 1000)
+
+-- # I like this name better
+unLinear :: Unrestricted a #-> a
+unLinear = unUnrestricted
```
aspiwack, line 88: But if you do, it needs to be changed globally not in a test file. Please don't do this.

test/Test/Data/Mutable/Array.hs
```diff
@@ -0,0 +1,230 @@
...
+type ArrayTester = Array.Array Int #-> Unrestricted (TestT IO ())
+
+-- | Given a size, make a random list of that lenght
+randomList :: Int -> Gen [Int]
+randomList size = do
+  let size' = Range.singleton size
+  Gen.list size' $ Gen.int $ Range.linearFrom 0 (-1000) 1000
+
+randomSize :: Gen Int
+randomSize =  Gen.int (Range.constant 100 1000)
```
aspiwack, line 81: Remove this. Just use a range in your random list (also… minimum 100? This means your counter example will be unnecessarily large most of the time).

Make the range `linear`, so that Hedgehog can fiddle with the sizes for you. So `Range.linear 0 1000` or something.

When you need the size later in the property, you can always define

```haskell
let size = length xs
```

test/Test/Data/Mutable/Array.hs
```diff
@@ -0,0 +1,230 @@
...
+  ]
+
+-- # Internal Library
+--------------------------------------------------------------------------------
+
+type ArrayTester = Array.Array Int #-> Unrestricted (TestT IO ())
+
+-- | Given a size, make a random list of that lenght
+randomList :: Int -> Gen [Int]
+randomList size = do
```
aspiwack, line 76: Prefer

```suggestion
list :: Int -> Gen [Int]
list size = do
```

Why? because, then you will write

```haskell
x <- forAll list
```

The point is that the property is not about randomness, or random lists, it's about lists. All of them. It happens to be checked using random examples, but that's all.

test/Test/Data/Mutable/Array.hs
```diff
@@ -0,0 +1,230 @@
...
+  let size' = Range.singleton size
+  Gen.list size' $ Gen.int $ Range.linearFrom 0 (-1000) 1000
+
+randomSize :: Gen Int
+randomSize =  Gen.int (Range.constant 100 1000)
+
+randomVal :: Gen Int
+randomVal = Gen.int (Range.linear 1 1000)
```
aspiwack, line 84: Likewise

```suggestion
val :: Gen Int
val = Gen.int (Range.linear 1 1000)
```

test/Test/Data/Mutable/HashMap.hs
```diff
@@ -0,0 +1,249 @@
...
+  Gen.string strSize Gen.alpha
+
+-- | Random pairs
+randKeyVals :: Gen [(Int, String)]
+randKeyVals = do
+  randSize <- Gen.int $ Range.constantFrom 600 400 800
+  let randSizeGen = Range.singleton randSize
+  keys <- Gen.list randSizeGen randKey
+  vals <- Gen.list randSizeGen randVal
+  return $ zip keys vals
```
aspiwack, line 136: Same as above.

test/Test/Data/Mutable/Vector.hs
```diff
@@ -0,0 +1,203 @@
...
+
+randomSize :: Gen Int
+randomSize =  Gen.int (Range.constant 100 1000)
+
+randomVal :: Gen Int
+randomVal = Gen.int (Range.linear 1 1000)
+
+-- # I like this name better
+unLinear :: Unrestricted a #-> a
+unLinear = unUnrestricted
```
aspiwack, line 67: All this like above :slightly_smiling_face:

test/Test/Data/Mutable/Array.hs
Divesh-Otwani: Oh -- yes, of course. That's brilliant and makes a lot of sense :)

test/Test/Data/Mutable/Array.hs
Divesh-Otwani: Yes, this was an unwise decision!

Divesh-Otwani:
Made all requested changes above :smiley:

src/Prelude/Linear.hs
```diff
@@ -21,7 +21,7 @@ module Prelude.Linear
     -- * Unrestricted
     -- $ unrestricted
   , Unrestricted(..)
-  , unUnrestricted
+  , unLinear
```
aspiwack, line 24: Do this in another PR, use `unUnrestricted` for now, and we change after. We should take the opportunity to bikeshed the name of the type and the data constructor, as well.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -11,75 +15,123 @@
...
 -- # Creation
-----------------------------
+-------------------------------------------------------------------------------
 
 -- | Allocate a constant array given a size and an initial value
-alloc :: Int -> a -> (Array a #-> Unrestricted b) -> Unrestricted b
-alloc size x f = f (Array size (newMutArr size x))
+-- Note that this does throw an exception and that there's a reasonable
+-- coherent view and method to using exceptions in Haskell!
+-- See https://www.tweag.io/posts/2020-04-16-exceptions-in-haskell.html.
```
aspiwack, line 53: No for this PR, but this is not very helpful. You won't be repeating this on every partial function. Simply specify your precondition, instead.

test/Test/Data/Mutable/Array.hs
```diff
@@ -0,0 +1,225 @@
...
+--------------------------------------------------------------------------------
+
+type ArrayTester = Array.Array Int #-> Unrestricted (TestT IO ())
+
+-- | A random list
+list :: Gen [Int]
+list = do
+  size <- Gen.int $ Range.linearFrom 1 1 1000
+  let size' = Range.singleton size
+  Gen.list size' $ Gen.int $ Range.linearFrom 0 (-1000) 1000
```
aspiwack, line 79:
```suggestion
list =
  let size = Range.linearFrom 1 1 1000 in
  Gen.list size value
```

Todo in a future PR:
- This one should be renamed `non_empty_list`
- There should be a `list` which can also generate empty lists. There appears to be few tests which make sense on empty lists. But when they do, it is risky not to test for it :slightly_smiling_face:

test/Test/Data/Mutable/Array.hs
```diff
@@ -0,0 +1,225 @@
...
+
+-- | A random value
+value :: Gen Int
+value = Gen.int (Range.linear 1 1000)
```
aspiwack, line 83:
```suggestion
value = Gen.int (Range.linear (-1000) 1000)
```

test/Test/Data/Mutable/Array.hs
```diff
@@ -0,0 +1,225 @@
...
+
+readWrite1Test :: Int -> Int -> ArrayTester
+readWrite1Test ix val arr =
+  compInts (move val) (getSnd Linear.$ Array.read (Array.write arr ix val) ix)
+
+readWrite2 :: Property
+readWrite2 = property $ do
+  size <- forAll $ Gen.int $ Range.linearFrom 2 2 1000
+  let size' = Range.singleton size
+  let list = Gen.list size' $ Gen.int $ Range.linearFrom 0 (-1000) 1000
```
aspiwack, line 142:
```suggestion
  size <- forAll $ Gen.int $ Range.linearFrom 2 2 1000
  let size' = Range.singleton size
  let list = Gen.list size' $ Gen.int $ Range.linearFrom 0 (-1000) 1000
```
```suggestion
  let list = Gen.list (Range.linearFrom 2 2 1000) value
  let size = length list
```

test/Test/Data/Mutable/Vector.hs
```diff
@@ -0,0 +1,197 @@
...
+
+-- | Given a size, make a random list of that lenght
+list :: Gen [Int]
+list = do
+  size <- Gen.int $ Range.linearFrom 1 1 1000
+  let size' = Range.singleton size
+  Gen.list size' $ Gen.int $ Range.linearFrom 0 (-1000) 1000
+
+val :: Gen Int
+val = Gen.int (Range.linear 1 1000)
```
aspiwack, line 61: Same changes as in the other file.

PR #99, opened 2020-04-27
Added mutable sets & axiomatic property tests

Divesh-Otwani:
I added mutable sets in the style of the arrays, vectors and hashmaps.


src/Data/Set/Mutable/Linear.hs
```diff
@@ -0,0 +1,82 @@
...
+import qualified Data.Coerce as Coerce
+import qualified Prelude.Linear as Linear
+import Unsafe.Linear as Unsafe
+import GHC.Stack
+
+
+-- # Data Definitions
+-------------------------------------------------------------------------------
+
+newtype Set a = Set (Linear.HashMap a ())
```
aspiwack, line 35: Maybe let's leave a comment to possibly improve on this representation later.

PR #105, opened 2020-04-29
First Pass at Module-Level Documentation

Divesh-Otwani:
I've gotten through a first pass at module level documentation.

Things I haven't gotten to yet (partly due to time, partly due to me needing to learn a few things to figure out what's going on):
(Italics means it's a public module.)

 * Control
    * *Optics.Linear.Internal.hs*
    * Monad
       * *Linear.hs*
       * Linear
          * Internal.hs
 * Data
    * Profunctor
       * *Linear.hs*
       * Kleisli
          * *Linear.hs*
    * Array:
       * *Polarized.hs*
          * *Push.hs*
          * *Pull.hs*
       * *Destination.hs*

### DIFF
--- a/src/Control/Monad/Linear.hs
+++ b/src/Control/Monad/Linear.hs
@@ -1,6 +1,20 @@
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 
+
+-- | This module defines the linear monad hierarchy and classic monad instances.
+--
+-- It's critical to know that the linear monad hierarchy is different from
+-- the "Data.Control.Functor" hierarchy.
+-- To understand the differences see this [blog post](https://www.tweag.io/posts/2020-01-16-data-vs-control.html).
+--
+-- == New to monads?
+--
+-- If you are unfamiliar with monads, we recommend [these](https://mmhaskell.com/monads/)
+-- tutorials on functors, monads, reader and writer monads, state monads and
+-- monad transformers for what's featured in this module. (Of course, there are
+-- a plethora of resources on monads in the Haskell community; find what works
+-- best for you!)
 module Control.Monad.Linear
   ( -- * Linear monad hierarchy
     -- $monad
@@ -11,18 +25,20 @@ module Control.Monad.Linear
   , Applicative(..)
   , dataPureDefault
   , Monad(..)
-  , MonadFail(..)
   , return
   , join
   , ap
-  , Data(..)
   , foldM
+  , MonadFail(..)
+  , Data(..)
   -- * Monad transformers
   -- ** ReaderT monad transformer
+  -- $readerT
   , Reader, reader, runReader, mapReader, withReader
   , ReaderT(..), runReaderT, mapReaderT, withReaderT
   , ask, local, asks
   -- ** StateT monad
+  -- $stateT
   , State, state, runState, execState, mapState, withState
   , StateT(..), runStateT, execStateT, mapStateT, withStateT
   , get, put, modify, gets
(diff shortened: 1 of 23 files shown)


aspiwack:
Is it still a draft? Does it need to be?

Remember: the shorter a PR is, the shorter it will be to review, and to merge. Two small PRs can be merged independently, where a big PR needs an overall review.

src/Data/Functor/Linear/Internal.hs
```diff
@@ -25,6 +25,9 @@ import qualified Control.Monad.Trans.Maybe as NonLinear
 import qualified Control.Monad.Trans.Except as NonLinear
 import qualified Control.Monad.Trans.State.Strict as Strict
 
+-- | Linear Data Functors should be thought of as containers of some type @a@
```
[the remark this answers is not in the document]
Divesh-Otwani: Yes -- this was sloppy of me. I'll try to see where it occurs and remove it. There shouldn't be a confusion between types and values.

Divesh-Otwani:
This is no longer a draft. For some reason in my head I thought this was a small PR. That's not the case. I'll put all the remaining documentation in smaller PRs.

src/Control/Monad/Linear.hs
```diff
@@ -1,6 +1,27 @@
...
+-- the "Data.Control.Functor" hierarchy.
+-- To understand the differences see this [blog post](https://www.tweag.io/posts/2020-01-16-data-vs-control.html).
+--
+-- You should use
+--
+-- * control functors for an effectful computation holding a single @a@ in a @f a@.
+-- * control applicative functors when you want a control functor that you can apply
+-- * linear monads when you want to sequence /linear/ functions with a context, i.e.,
+--   when you want to sequence functions of the form @a #-> m b@.
+--
```
aspiwack, line 17: I'm not sure these sentences carry any information unfortunately. Skip this for now. It's alright, linear-base is for fairly advanced users as it stands.
```suggestion
```

src/Control/Monad/Linear.hs
```diff
@@ -35,9 +58,46 @@ import Data.Unrestricted.Linear
 import Prelude.Linear.Internal.Simple ((.), ($))
 import qualified Data.Functor.Linear.Internal as Data
 
+
+-- $stateT
```
aspiwack, line 62: I'd put this paragraph close to `StateT` in the file. But, otherwise: ok.

src/Control/Monad/Linear/Internal.hs
```diff
@@ -30,37 +31,59 @@ import qualified Control.Monad.Trans.State.Strict as NonLinear
 
 -- TODO: make the laws explicit
 
--- | Enriched linear functors.
+-- | Enriched linear functors. These are control functors. The functor of type
```
aspiwack, line 34: Remove any reference to enriched functors (prefer control functors) from the documentation. It's just an artifact of the fact that I didn't have names for them, then, and I used the mathematical terminology.

I'll make a short reference to enriched functors in a mathematical note in the future. For reference, but it is not relevant for most of the documentation, as I don't want to use this terminology anymore.

src/Data/Bifunctor/Linear.hs
```diff
@@ -4,23 +4,45 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE TypeOperators #-}
 
-module Data.Bifunctor.Linear where
+-- | This module provides Bifunctor and related classes.
+--
+-- == 'Bifunctor'
+--
+-- Use a bifunctor instance to map functions over structures that hold
+-- two data types and could have a functor instance over either one.
```
aspiwack, line 12: I think I've written this before, but “hold two data types” is not very meaningful terminology, we need to find something better.

src/Data/Functor/Linear.hs
```diff
@@ -3,17 +3,22 @@
...
+-- which represent containers, and control functors which represent effects.
+-- Please read this
+-- [blog post](https://www.tweag.io/posts/2020-01-16-data-vs-control.html).
+-- For more details, see "Control.Monad.Linear".
+--
+-- * Linear data functors should be thought of as containers of data.
+-- * Linear data applicative functors should be thought of as containers
+-- that can be zipped.
+-- * Linear data traversible functors should be thought of as
+-- containers which store a finite number of values.
```
aspiwack, line 17: Contrary to the equivalent bullet-point list in `Control.Monad`, I quite like this one!

src/Unsafe/MutableArray.hs
```diff
@@ -23,42 +31,57 @@ import GHC.Exts
...
 readMutArr mutArr (I# ix) =
   case doRead of (# _, a #) -> a
   where
     doRead = runRW# $ \stateRW -> readArray# mutArr ix stateRW
 
--- | Resize a mutable array, using the first value to fill in the new cells
+-- | Grow a mutable array. This is given an array, a given larger size,
+-- and it returns a larger array of the given size using the first value
+-- to fill in the new cells and copying over all the unchanged cells. This
+-- fails for a size smaller than the size of the given array.
```
aspiwack, line 69: It's ok. And so are the other functions in this module, I'm just picking one randomly. But I'm a big fan of writing documentation like this:

```
-- | @'resizeMutArr' arr size@, where @size@ is larger than the length of @arr@, returns a new array with length @size@ […]
```
That is: give name to things, and refer to them precisely, rather than using semi-precise English phrases to work around the fact that you didn't give yourself names. So, in the future, I want to see more documentation like this.

It's particularly easy/reasonable for data structures like this, since everything tends to have a super clear semantics, so you can give a pretty formal description of the function as its docstring.

src/Control/Monad/Linear.hs
Divesh-Otwani: StateT is defined in another file that's internal so I can't migrate it.

PR #106, opened 2020-04-30
A few simple analogs from base

Divesh-Otwani:
I'm getting a head start on some of the analogs from base that we need.

I made a few minor design choices that I hope are straightforward :)

### DIFF
--- a/src/Data/Bool/Linear.hs
+++ b/src/Data/Bool/Linear.hs
@@ -1,0 +1,39 @@
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+-- | This module provides linear functions on the standard 'Bool' type.
+module Data.Bool.Linear
+  ( -- * The Boolean type
+    Bool(..)
+    -- * Operators
+  , (&&)
+  , (||)
+  , not
+  )
+  where
+
+import Prelude (Bool(..))
+
+-- | @True@ iff both are @True@.
+-- __NOTE:__ this is strict and not lazy!
+(&&) :: Bool #-> Bool #-> Bool
+False && False = False
+False && True = False
+True && x = x
+
+infixr 3 &&
+
+-- | @True@ iff either is @True@
+-- __NOTE:__ this is strict and not lazy!
+(||) :: Bool #-> Bool #-> Bool
+True || False = True
+True || True = True
+False || x = x
+
+infixr 2 ||
+
+-- | @not b@ is @True@ iff b is @False@
+-- __NOTE:__ this is strict and not lazy!
+not :: Bool #-> Bool
+not False = True
+not True = False
--- a/src/Data/Ord/Linear.hs
+++ b/src/Data/Ord/Linear.hs
@@ -1,0 +1,102 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE MonoLocalBinds #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Data.Ord.Linear
+  ( Ord(..)
+  , min
+  , max
+  , compare
+  )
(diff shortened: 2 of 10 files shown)


src/Data/Bool/Linear.hs
```diff
@@ -0,0 +1,35 @@
...
+  , (&&)
+  , (||)
+  , not
+  )
+  where
+
+import Prelude (Bool(..))
+
+-- | True iff both are True.
+(&&) :: Bool #-> Bool #-> Bool
```
aspiwack, line 18: I'm wondering out loud whether we should take the opportunity to make a boolean algebra type class to abstract over these? Maybe not, it many not be useful often enough to be featured in a standard library.

src/Data/Ord/Linear.hs
```diff
@@ -0,0 +1,62 @@
...
+class Eq a => Ord a where
+  {-# MINIMAL compare | (<=) #-}
+  compare :: a #-> a #-> Ordering
+  compare = Unsafe.toLinear2 compare'
+    where
+      compare' :: Ord a => a -> a -> Ordering
+      compare' x y
+        | x == y = EQ
+        | x <= y = LT
+        | True = GT
```
aspiwack, line 28: If you need `Unsafe.toLinear`, then it's probably unsafe :wink: . 

The conclusion is that there is no linear definition of `compare` based on just `(==)` and `(<=)`, and that's fine.
```suggestion
```

src/Data/Ord/Linear.hs
```diff
@@ -0,0 +1,62 @@
...
+  where
+
+import Data.Eq.Linear
+import qualified Prelude as Unrestricted
+import Prelude (Bool(..))
+import Data.Ord (Ordering(..))
+import qualified Unsafe.Linear as Unsafe
+
+class Eq a => Ord a where
+  {-# MINIMAL compare | (<=) #-}
```
aspiwack, line 20: As a consequence:

```suggestion
  {-# MINIMAL (<=) #-}
```

src/Data/Ord/Linear.hs
```diff
@@ -0,0 +1,62 @@
...
+        GT -> x
+        _ -> y
+
+  min :: a #-> a #-> a
+  min = Unsafe.toLinear2 min'
+    where
+      min' :: Ord a => a -> a -> a
+      min' x y = case compare x y of
+        LT -> x
+        _ -> y
```
aspiwack, line 59: You can define these without `Unsafe.toLinear`, at the cost of one extra binding in `where`. Please do.

src/Data/Ord/Linear.hs
```diff
@@ -0,0 +1,62 @@
...
+
+instance Unrestricted.Ord a => Ord a where
+  compare = Unsafe.toLinear2 Unrestricted.compare
```
aspiwack, line 62: This is very wrong. We could give some (unsafe!) deriving-via combinator to that effect. But let's do that later: add an issue instead.

```suggestion
```

src/Data/Bool/Linear.hs
facundominguez: Do this functions deserve a word about being strict? (unlike their non-linear variants)

aspiwack: This is a good point. Well-spotted.

Divesh-Otwani: Yes linearity forces them to be strict. That's interesting -- so in some sense giving linear types to functions in `base` that are linear does impact performance. (So a non-linear function using the linear `&&` non-linearly might have a different performance. So linear types don't impact the performance of a function that's already written, but they do force constraints on the implementation which might impact performance. This is an interesting observation.)

src/Data/Ord/Linear.hs
Divesh-Otwani: I'm probably missing something but I don't think that's possible.

If you don't know `a` is `Dupable` as soon as you use it in `<` or `<=` you can't return it.

PR #108, opened 2020-05-12
Doc/lenses

Divesh-Otwani:
I finally finished documenting optics.

- I make a lot of claims and I think they are all sound
- I tried to make it as friendly as possible without repeating things from elsewhere

### DIFF
--- a/src/Control/Optics/Linear/Internal.hs
+++ b/src/Control/Optics/Linear/Internal.hs
@@ -6,6 +6,18 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE TypeOperators #-}
 
+-- | This module provides linear optics.
+--
+-- == Learning about optics
+--
+-- == Examples of 'Lens'-es
+--
+-- == Examples of 'Prism's
+--
+-- == Examples of 'Iso'morphisms
+--
+-- == Examples of 'Traversal's
+--
 module Control.Optics.Linear.Internal
   ( -- * Types
     Optic_(..)
@@ -44,8 +56,18 @@ import Data.Void
 import Prelude.Linear
 import qualified Prelude as P
 
+
+-- | A wrapper for an Optic that uses a "Data.Profunctor.Linear" @arr@
 newtype Optic_ arr a b s t = Optical (a `arr` b -> s `arr` t)
 
+-- | @Optic c a b s t@ is essentially
+-- > forall arr. c arr => (a `arr` b ) -> (s `arr` t)
+-- for some @arr@ that is an instance of "Data.Profunctor.Linear".
+--
+-- An optic is a generic way of lifting a computation on values of type @a@ to
+-- values of type @s@. You have some value of type @s@, from which you could
+-- access value(s) of type @a@. After performing computations of type
+-- @a `arr` b@, you change that value of type @s@ into a value of type @t@.
 type Optic c a b s t =
   forall arr. c arr => Optic_ arr a b s t
 
--- a/src/Control/Optics/Linear/Internal.hs
+++ b/src/Control/Optics/Linear/Internal.hs
@@ -67,7 +67,7 @@
 -- === How can each optic be used? How do I think about them?
 --
 -- * A @Traversal a b s t@ is roughly equivalent to a
---  @s #-> (Tree a, Tree b #-> t)@.
+--  @s \#-> (Tree a, Tree b \#-> t)@.
 --  It provides a means of accessing several @a@s organized in some
 --  structural way in an @s@, and a means of changing them to @b@s to create a
 --  @t@. In very ordinary language, it's like walking or traversing the data
@@ -75,22 +75,22 @@
 --  language, it's like selecting some specific @a@s by looking at each
 --  constructor of a data definition and recursing on each non-base type.
 --
--- * A @Lens a b s t@ is equivalent to a @(s #-> (a,b #-> t)@.  It is a way to
+-- * A @Lens a b s t@ is equivalent to a @(s \#-> (a,b \#-> t)@.  It is a way to
 --  cut up an instance of a /product type/ @s@ into an @a@ and a way to take a
 --  @b@ to fill the place of the @a@ in @s@ which yields a @t@. When @a=b@ and
---  @s=t@, this type is much more intuitive: @(s #-> (a,a #-> s))@.  This is a
+--  @s=t@, this type is much more intuitive: @(s \#-> (a,a \#-> s))@.  This is a
 --  traversal on exactly one @a@ in a @s@.
 --
--- * A @Prism a b s t@ is equivalent to @(s #-> Either a t, b #-> t)@ for some
+-- * A @Prism a b s t@ is equivalent to @(s \#-> Either a t, b \#-> t)@ for some
 --  /sum type/ @s@. In the non-polymorphic version, this is a 
---  @(s #-> Either a s, a #-> s)@ which represent taking one case of a sum type
+--  @(s \#-> Either a s, a \#-> s)@ which represent taking one case of a sum type
 --  and a way to build the sum-type given that one case.  A prism is a
 --  traversal focusing on one branch or case that a sum type could be.
 --
--- * An @Iso a b s t@ is equivalent to a @(s #-> a, b #-> t)@.  In the simple
+-- * An @Iso a b s t@ is equivalent to a @(s \#-> a, b \#-> t)@.  In the simple
 --  case of an @Iso' a s@, this is equivalent to inverse functions
---  @(s #-> a, a #-> s)@.  In the general case an @Iso a b s t@ means if you
---  have the isomorphisms @(a #-> b, b #-> a)@ and @(s #-> t, t #-> s)@, then
+--  @(s \#-> a, a \#-> s)@.  In the general case an @Iso a b s t@ means if you
+--  have the isomorphisms @(a \#-> b, b \#-> a)@ and @(s \#-> t, t \#-> s)@, then
 --  you can form isomorphisms between @s@, @t@, @a@ and @b@.
 --
 -- === A bird's eye view of the types
@@ -102,8 +102,8 @@
 --
 -- > type Traversal s t a b =
 -- >   forall f. Applicative f => (a -> f b) -> s -> f t
--- > type Lens s a =
--- >   forall f. Functor f => (a -> f a) -> (s -> f t)
+-- > type Lens s t a b =
+-- >   forall f. Functor f => (a -> f b) -> (s -> f t)
 -- > type Prism s t a b =
 -- >   forall p f. (Choice p, Applicative f) => p a (f b) -> p s (f t)
 -- > type Iso s t a b =
@@ -134,7 +134,7 @@
 -- +=================+============+===============+====================+===========+
 -- |     @(->)@      |     X      |       X       |         X          |           |
 -- +-----------------+------------+---------------+--------------------+-----------+
--- |    @(#->)@      |     X      |       X       |         X          |           |
+-- |    @(\#->)@     |     X      |       X       |         X          |           |
 -- +-----------------+------------+---------------+--------------------+-----------+
 -- |    (Prelude)    |            |               |                    |           |
 -- |  @Functor f@    |            |               |                    |           |
--- a/src/Control/Optics/Linear/Internal.hs
+++ b/src/Control/Optics/Linear/Internal.hs
@@ -16,23 +16,28 @@
 --
 -- This documentation provides
 --
---  * a high-level idea of what linear optics are and how to think about them
+--  * a high-level idea of
+--     * what linear optics are
+--     * how to think about linear optics
+--     * how these relate to the standard optic definitions in @lens@
 --  * examples of using these optics
 --
--- == How do I think about /linear/ optics?
+-- == Conceptualizing and using optics
+--
+-- === What are /linear/ optics?
 --
 -- **Optics can be conceptualized as a first class object with which you can
 -- view and map functions over sub-structure(s) within a larger structure.**
 --
--- /Linear/ optics are optics where the \"viewing\" and \"mapping\" are done
+-- __/Linear/ optics are optics where the \"viewing\" and \"mapping\" are done
 -- with linear functions (and any corresponding structures hold values
--- linearly, i.e., with constructors that use linear arrows).
+-- linearly, i.e., with constructors that use linear arrows).__
 --
 -- In types: a (linear) optic of type @Optic a b s t@ is a way of viewing the
 -- sub-structures of type @a@ in the structure of type @s@ and mapping a
 -- function from an @a@ to a @b@ on those sub-structures in @s@ which change an
 -- @s@ to a @t@. The non-polymorphic version of the optic is specialized
--- to the types @Optic a a s s@ and is usually defined with a tick mark, 
+-- to the types @Optic a a s s@ and is usually defined with a tick mark,
 -- e.g., the non-polymorphic @Lens@ is @Lens'@.
 --
 -- There are four basic optics: traversals, lenses, prisms and isomorphisms.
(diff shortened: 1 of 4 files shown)


b-mehta:
(This was originally meant to be a two-line review but turned into the email I promised @Divesh-Otwani some time ago, putting it here anyway so @aspiwack can comment as well. It consists of hopefully coherent ramblings on my lens thoughts, starting from the concrete finished ideas to the half-baked (and sometimes raw) ones.) 

For explaining (linear) lenses, I personally think it's better to avoid the "some type `x`" description and instead look at `(s #-> (a, b #-> t))`. The intuition here (starting with `Lens'` for now) is that we can cut up `s` into `a` and `a #-> s`, that is we take out the `a` component and we're left with a "slot for" a new `a` which would produce a new `s` - the key example here should be tuples. For `Lens` proper, we cut up `s` into `a` and a slot for a new type `b`: pairing this with the "rest of" the original `s` we get something of type `t`. (I thought I made a constructor for `Lens` of the form:
```haskell
lens :: (s #-> (a, b #-> t)) -> Lens a b s t
```
not sure if I got around to it though. Similarly it should be possible to have the inverse of this by defining the right Profunctor. Again I thought I'd done this in some branch - it should look something like `withPrism` or `withIso` below. (EDIT: https://github.com/tweag/linear-base/blob/lots-of-optics/src/Control/Optics/Linear/Internal.hs#L81)
On that note, I put in some weirdly named profunctors (eg market and exchange) - these were lifted virtually without modification from eg `http://hackage.haskell.org/package/lens-4.19.2/docs/Control-Lens-Internal-Prism.html` - they weren't intended to be external.
The reason I like this intuition for lenses is that it works just as well for linear and non-linear lenses: I'm not much of a fan of the coend-like notion suggested by the "some type `x`", for instance it's not entirely clear (to me at least) whether the category we should be considering is the self-enriched category of linear types or the (non-enriched) category of linear types, or the usual category of types enriched in linear types, or just the usual category of types with an extra closed structure. In particular it's not clear which of these is the right way to think about linear lenses.

Similar ideas work nicely for Prism and Iso: expressed in `prism` and `iso` and `withPrism` and `withIso`. Now that I look at it I don't entirely remember why `withPrism` has that type instead of `Optic_ (Market a b) a b s t -> (b #-> t, s #-> Either t a)` though I do remember there being a good reason.

You might like to add something about why `newtype`s are used instead of `type`s for lenses: there's the downside of introducing a new operator `.>` so it's probably worth justifying (I'm sure aspiwack can talk much more than me on the upsides). Cf https://www.well-typed.com/blog/2019/09/announcing-the-optics-library/

Another useful idea for your intuition or more probably for documentation somewhere would be a table of classes and profunctors (at least, I found this helpful). On one side would be a list of the profunctors (eg `Nonlinear.Kleisli f`) and on the other would be the classes they could satisfy (eg `Strong (,) ()`), and in each cell would be the conditions needed for that profunctor to satisfy that class: [in this case `Prelude.Functor f`](https://github.com/tweag/linear-base/blob/f8b2c7686975377aab60da5f40ca14538ea4f1b0/src/Data/Profunctor/Linear.hs#L97). Of course some would just be a yes/no, eg for the profunctor `Market`. I'd give you the one I'd made but I have no idea where it is, sorry.

Another thing I thought about for some time was how linear we can get with lenses. The usual lens introduction story (eg [here](https://skillsmatter.com/skillscasts/4251-lenses-compositional-data-access-and-manipulation)) starts by talking about `get` and `set` so it's tempting to want linear versions of these. It's pretty easy to see that this doesn't work in the naive way, so in `master` there are basic non-linear `get` and `set`. I explored how linear we can make these in other branches, eg `https://github.com/tweag/linear-base/blob/dad3c1019279a8f80ad04f2434b366689a447c4f/src/Control/Optics/Linear/Internal.hs#L163`. In particular, `get'` will get the value `a` but also some other stuff we don't want: linearity doesn't let us discard it so I chucked it into `Top` (I think you called `Top` something else, `Waste` maybe?). The advantage of this is that it's easy to get something of the form `s -> a` from something of the form `s #-> (Top, a)`, since the latter is basically a special case of the former. One could also compose arrows of the form `s #-> (Top, a)`: they're obviously Kleisli arrows for `Top * -`. The idea of this sort of arrow is that it's somewhere between a nonlinear and a linear arrow: you can't duplicate but you can discard. Note it's different from `s & 1 #-> a`, since the latter needs you to use all or none of `s` but the former lets you use a portion of it. This is related to affine types.
The potential disadvantage of `get'` compared to `get` is evident from looking at the type signatures:
```haskell
get :: Optic_ (NonLinear.Kleisli (Const a)) a b s t -> s -> a
get l = gets l P.id

get' :: Optic_ (Linear.Kleisli (Const (Top, a))) a b s t -> s ->. (Top, a)
get' l = gets' l id
```
`get'` needs something different from the optic it's given. I say different here rather than stronger or weaker - I don't recall which typeclasses are satisfied respectively by `(NonLinear.Kleisli (Const a))` and `(Linear.Kleisli (Const (Top, a)))` but if I remember correctly it's not a simple relationship (the table I mentioned above would clear up what this relationship should be) - I half remember something about nonlinear vs linear monoids floating around here. I don't have any concrete answers here, but it's something to consider.
There's similar stuff happening with `set'` and `set''`:
```
set' :: Optic_ (Linear.Kleisli (MyFunctor a b)) a b s t -> s ->. b ->. (a, t)
set' (Optical l) s = runMyFunctor (Linear.runKleisli (l (Linear.Kleisli (\a -> MyFunctor (\b -> (a,b))))) s)

set'' :: Optic_ (NonLinear.Kleisli (Reader b)) a b s t -> b ->. s -> t
set'' (Optical l) b s = runReader (NonLinear.runKleisli (l (NonLinear.Kleisli (const (Reader id)))) s) b
```
In `set'` we combine an `s` and `b` to produce an `a, t` - in effect swapping out the `a` inside the `s` for `b` to produce a `t`. It might be nicer to think of this as `(s,b) #-> (a,t)`. (Side note: compare to Iso which looks like `(s #-> a, b #-> t)`). `set''` below is a slightly more non-linear version of this, just discarding the `a` (maybe it could land in a Top instead). I'm pretty sure I wrote this before realising that `s #-> (a, b #-> t)` is the most sensible version of a linear lens, but the comparisons between the versions are interesting: in this we get the `a` just by supplying the `s`, but in `set'` we only get the `a` after supplying `b` as well.
Again no conclusions here, but explaining some of the ideas surrounding this stuff (you can see just how half-baked this was from the super helpful name `MyFunctor`!).
Similarly there's [this](https://github.com/tweag/linear-base/blob/two-traversals/src/Control/Optics/Linear/Internal.hs#L183) - a less linear version of `match` with a different (I think weaker, but not sure) constraint. On this note, it looks like here I used `Kleisli (Either _)` instead of `Market` - if they turns out to be the same probably the Kleisli version is best to keep.

And now we come to Traversals. The story here is a mess, so I'll leave it for later. Briefly though, the `Batch` type seems like the right type to represent linear Traversals in the same way that `s #-> (a, b #-> t)` represents linear lenses, but it's not clear what efficiency considerations there are. There's also a bunch of questions about what linear `Traversables` should look like (eg should the applicative be prelude, data or control?). There's a paper which suggests every law-abiding (Prelude) traversable enforces linearity, but I didn't see a proof anywhere, though I had a suspicion the `Batch` type would help with it. Then should every Prelude Traversable be Linear Traversable? In any case, even if this were provable, it would probably still be hard to convince ghc.  I never really got my head around this stuff so it's even less coherent than the previous. I would suggest looking at (some variants of) the `both` optic to take examples to think about - it's not prototypical in the same way that `Either` is for `Prism` and tuples are for `lens` but I found it informative. I _think_ the most coherent branch here is `one-traversal` but I'm not sure - apologies. `PTraversal` was meant to indicate Prelude traversal, `DTraversal` Data traversal and `BTraversal` Batch traversal (i.e. using the Batch type). 

Hope some of this was useful!

Divesh-Otwani:
Thanks for this ^. It makes a lot of sense and is very useful. A few thoughts:

- I think the formulation of `lens` is very sensible
- I'll absolutely add a table of type classes and key instances with their laws; that seems really useful.
- I think `get'` makes sense but I'd have to look at it closer. I haven't examined or played around with Top. I like the idea.
- I agree with the intuitions about set' but I'm not so sure about the type.
- I see the thing about traversals; it's seems tricky to determine which applicative constraint we want for them.

In general, to address developing out our optics, I'm going to take an example-based approach. I'll make some longer-form real-ish examples that use lenses, prisms, traversals and isomorphisms and try out different designs on them. I think it's very difficult to try to determine the sensible definitions of linear optics without a lot of motivation (whereas it's obvious for simpler things like "linear boolean facilities").

Divesh-Otwani:
Making this a TODO: I should use doctest.

aspiwack:
@utdemir I'll leave it to you to review this PR.

src/Control/Optics/Linear/Internal.hs
```diff
@@ -6,31 +6,442 @@
...
+--
+--  * X is a specialization of Y
+--  * X is a strict subset of Y
+--  * You can (basically) implement @f :: X -> Y@ with @f = id@
+--  but you can't implement @f :: Y -> X@.
+--
+-- === How can each optic be used? How do I think about them?
+--
+-- * A @Traversal a b s t@ is roughly equivalent to a
+--  @s \#-> (Tree a, Tree b \#-> t)@.
```
utdemir, line 70: Why is there a `Tree` here?

src/Control/Optics/Linear/Internal.hs
```diff
@@ -6,31 +6,442 @@
...
+-- >
+-- > buildUp :: Maybe Int #-> IntStructure
+-- > buildUp Nothing = SomeInts []
+-- > buildUp (Just i) = OneInt i
+-- >
+-- > -- A traversal that only looks at one int, in two different cases
+-- >
+-- > intStructT :: Traversal' Int IntStructure
+-- > intStructT = Optical $ \int_arr_int ->
+-- >   Linear.dimap extract buildUp (Linear.wander int_arr_int)
```
utdemir, line 257: Is this a lawful traversal? I might be wrong here, but `extract` loses some elements from the `IntStructure` which I find a bit strange. The `lens` documentation states that [http://hackage.haskell.org/package/lens-4.19.2/docs/Control-Lens-Traversal.html](page):

> a Traversal needs to leave the same number of elements as a candidate for subsequent Traversal that it started with.

src/Control/Optics/Linear/Internal.hs
```diff
@@ -6,31 +6,442 @@
...
+-- > fromTuple :: (a,b) #-> Pair a b
+-- > fromTuple (a,b) = Pair a b
+-- >
+-- > pairTupleIso :: Iso' (a,b) (Pair a b)
+-- > pairTupleIso = iso toTuple fromTuple
+-- >
+-- > pair1 :: Pair Int Int
+-- > pair1 = Pair 2 3
+-- >
+-- > setPairViaTuple :: (a,b) -> Pair a b -> Pair a b
```
utdemir, line 335: Isn't the second arrow has to be linear in this signature?

src/Control/Optics/Linear/Internal.hs
```diff
@@ -6,31 +6,442 @@
...
+-- about abelian groups and in fact, the theorems that hold on abelian groups
+-- are a superset of the theorems that hold on just groups.
+--
+-- __Please take care to understand the principle at play:__
+--
+-- /If @D@ is a stricter constraint than @C@ (i.e., any @D@ instance has a @C@/
+-- /instance), then the rank-2 type @forall x. C x => f@ is a sub-type of/
+-- /@forall x. D x => f@/
+--
+-- === What do these optic types /mean/?
```
utdemir, line 397: I found this section a bit out of place. I think it spends too much space to say "the types are complicated and very general", which is not a very useful insight IMO.

src/Control/Optics/Linear/Internal.hs
```diff
@@ -44,71 +455,157 @@ import Data.Void
 import Prelude.Linear
 import qualified Prelude as P
 
+
+-- # Data Type Definitions
+-------------------------------------------------------------------------------
+
+-- | A wrapper for an Optic that uses a "Data.Profunctor.Linear" @arr@
```
utdemir, line 462: This line sounds like it's specialisation of a more generic data type, and my instinct when I read "a wrapper" is just skipping it and looking for the actual type, which doesn't exist in this case. I'd probably just paste the source as the haddock.

src/Control/Optics/Linear/Internal.hs
```diff
@@ -123,19 +620,36 @@ lengthOf l s = getAdded (gets l (const (Adding one)) s)
 
 -- XXX: the below two functions will be made redundant with multiplicity
 -- polymorphism on over and traverseOf'
+
+-- | The same as 'over' but for normal non-linear arrows @(->)@
```
utdemir, line 624: The word 'normal' sounds redundant here.

src/Control/Optics/Linear/Internal.hs
```diff
@@ -123,19 +620,36 @@ lengthOf l s = getAdded (gets l (const (Adding one)) s)
...
+iso f g = Optical (dimap f g)
+
+-- | @prism builder takeCase@ constructs a prism given a @builder@
+-- which goes from a @b@ (produced by applying a function of type
+-- @a -> b@ to the @a@ case of the sum type @s@) to a @t@ and a
+-- @takeCase@ which breaks @s@ down into either the @a@ case or the
+-- transformed structure of type @t@.
+prism :: (b #-> t) -> (s #-> Either t a) -> Prism a b s t
+prism b s = Optical $ \f -> dimap s (either id id) (second (rmap b f))
+
```
utdemir, line 655: I think the `lens` function on the examples would be a nice addition here.

src/Data/Profunctor/Linear.hs
```diff
@@ -56,13 +102,28 @@ class (SymmetricMonoidal m u, Profunctor arr) => Strong m u arr where
   second arr = dimap swap swap (first arr)
   {-# INLINE second #-}
 
+-- | This typeclass asserts the constraint on the profunctor @arr@ that
+-- function-like things of type @a `arr` b@ with certain properties
+-- can be extended to work on traversible structures. These are the properties
+-- these function-like things need: they need @Strong (,) ()@ and
```
utdemir, line 108: These four lines just repeats what the first line of the instance definition. I think we should just remove them.

src/Data/Unrestricted/Linear.hs
```diff
@@ -10,15 +10,70 @@
...
+-- >     withDuplicate :: (Int, Int) #-> Int #-> Bool
+-- >     withDuplicate (ix,ix') size = (0 <= ix) && (ix < size)
+-- >     (<) :: Int #-> Int #-> Bool
+-- >     (<) = ...
+-- >
+-- >     (<=) :: Int #-> Int #-> Bool
+-- >     (<=) = ...
+-- >
+-- >     (&&) :: Bool #-> Bool #-> Bool
+-- >     (&&) = ...
```
utdemir, line 57: I'd probably instead implement a simpler function like `square x = x * x`.

src/Data/Unrestricted/Linear.hs
```diff
@@ -39,14 +93,27 @@ import Data.Monoid.Linear
...
   Unrestricted :: a -> Unrestricted a
 
--- | Project an @a@ out of an @Unrestricted a@. If the @Unrestricted a@ is
--- linear, then we get only a linear value out.
+-- | Get an @a@ out of an @Unrestricted a@. If you call this function on a
+-- linearly bound @Unrestricted a@, then the @a@ you get out has to be used
+-- linearly, for example:
+--
+-- > restricted :: Unrestricted a #-> b
+-- > restricted x = f (unUnrestricted x)
```
utdemir, line 112: This name does not match the actual name of the function.

src/Control/Optics/Linear/Internal.hs
Divesh-Otwani: I'm cutting this section almost entirely. Not ready to write it yet.

Divesh-Otwani:
Closing and being replaced by #247.

PR #110, opened 2020-05-15
Doc/dps arrays

Divesh-Otwani:
I documented DPS arrays (and did some background reading on them).

(By the way, I don't see how this implementation reduces the number of allocations; it seems like it doesn't actually help -- though I'm probably missing something. The example I made is doable with exactly one allocation of a temporary array of size n in C. This is not the case here.)

### DIFF
--- a/src/Data/Array/Destination.hs
+++ b/src/Data/Array/Destination.hs
@@ -2,9 +2,102 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
+-- | This module provides destination arrays
+--
+-- We can use linear types to release a safe API for destination arrays
+-- and DPS; with this it's much easier to write memory efficient programs
+-- that can often be as fast as idomatic C code (with the simplicity of
+-- Haskell and FP, of course).
+--
+-- == What are destination arrays and DPS?
+--
+-- Destination arrays are arrays that are meant to written to inside a
+-- function. These arrays are used for DPS (destination passing style
+-- programming), a style of programming that uses memory
+-- efficiently.
+--
+-- In this style of programming, certain functions are given
+-- an array to write into instead of returning a result.
+--
+-- This style is common place in C programming. One might use it to write this
+-- function, in which @sum@ is a destiniation array.
+--
+-- @
+-- void add(int size, int *a, int *b, int *sum){
+--   for (int i=0; i<size;++i){sum[i]=a[i]+b[i];}
+-- }
+-- @
+--
+-- C programs using DPS are much more efficient.
+-- [TODO why and how!? ... explain!!]
+--
+-- For a deeper background, see
+-- [this paper](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/11/dps-fhpc17.pdf)
+--
+-- == Why do we need linear types?
+--
+-- Because of linear types, any function that uses a destination array
+-- must take that 'DArray' as a linear argument. This is because the only
+-- way to create a 'DArray' is through 'alloc' which takes a linear
+-- continutation of type @DArray a #-> ()@.
+--
+-- Since 'DArray' doesn't have a 'Consumable' instance, the only way to
+-- consume it is with the given API (e.g., with 'fill' or perhaps
+-- 'fromFunction'). 
+--
+-- Hence, /linearity enforces that any function using a 'DArray' consumes it
+-- and does so by writing to each cell of the array/.
+--
+-- Another property of the API is that we guarentee efficient stack allocation
+-- of destination arrays -- i.e., unlike how C programmers must be careful to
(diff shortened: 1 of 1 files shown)


aspiwack:
I don't know what your question actually is. But if what you are trying to do is compare mutable arrays and destinations, my suggestion is to see how the interaction of split and freeze (the latter being part of the allocation primitive in destinations) is different.

src/Data/Array/Destination.hs
```diff
@@ -2,9 +2,86 @@
...
+-- >   sendVectorToTwoServers vecToSend
+-- >   where
+-- >
+-- >     sendVectorToTwoServers :: Vector Int -> IO ()
+-- >     sendVectorToTwoServers _ = {- fake implementation -}
+-- >       return ()
+-- >
+-- >     inputVector :: IO (Vector Int)
+-- >     inputVector =
+-- >       return (fromList (map (\x -> (7 * (x+3)) `div` 11) [1..100]))
```
utdemir, line 78: Instead of describing an IO action, I think this would be a shorter example if we were to just describe a `Vector Int -> Vector Int` function. eg.

```
computeDiff :: Vector Int -> DPS.DArray Int #-> ()
computeDiff = undefined

vectorDiff :: Vector Int -> Vector Int
vectorDiff vec =
  let diffSize = (Vector.length vec) - 1
  in  DPS.alloc diffSize (computeDiff vec)
```

src/Data/Array/Destination.hs
```diff
@@ -2,9 +2,86 @@
...
+--
+-- Because of linear types, any function that uses a destination array
+-- must take that 'DArray' as a linear argument. This is because the only
+-- way to create a 'DArray' is through 'alloc' which takes a linear
+-- continutation of type @DArray a #-> ()@ and any function inside that
+-- continuation that uses a @DArray@ must use it linearly.
+--
+-- Since 'DArray' doesn't have a 'Consumable' instance, the only way to
+-- consume it is with the given API (e.g., with 'fill' or perhaps
+-- 'fromFunction') which fills the destiniation array completely
```
utdemir, line 45: Typo here on `destination`.

src/Data/Array/Destination.hs
```diff
@@ -2,9 +2,86 @@
...
+--   for (int i=0; i<size;++i){sum[i]=a[i]+b[i];}
+-- }
+-- @
+--
+-- == Linear Types Enforce Destination Arrays are Filled
+--
+-- Linear types are used to ensure that the destination array
+-- is always written to. Why? Well:
+--
+-- Because of linear types, any function that uses a destination array
```
utdemir, line 37: "because of linear types" sounds a bit hand-wavy. The explanation after this sentence is clear, so maybe we should just start with that ("The only way to create ...").

aspiwack:
@utdemir I believe, on the contrary, that it makes you the best person to review: the documentation is supposed to make you understand why you want to use this.

I'm available for specific questions though.

utdemir:
@aspiwack I think I am missing the main point, because I still don't understand how this can be more efficient than just calling something like [`Vector.fromListN`](https://hackage.haskell.org/package/vector-0.12.1.2/docs/Data-Vector.html#v:fromListN) inside the function.

The only thing I can think of is that creating a `DArray` once, splitting them and filling using two different functions would be faster than combining two vectors after the fact. But the module gives the impression that it is not the only point.

aspiwack:
Sorry, sorry, I forgot to reply to this one. I promise that I'm catching up with my email and stuff.

“More efficient” is a bit vague. The general way that I think about destinations is that taking a destination instead of returning an array lets you ask a new question, namely: whose responsibility is it to allocate the array. When you return an array, you have no choice: it is you who will allocate the array. Where as when you take a destination as an argument, it can be someone else.

At this point, exercise: show that from a dps function `f :: a -> DArray b #-> ()`, you can define an array-returning function `f :: a -> Array b`. So DPS is indeed the more general form.

There can be many reasons for doing such a thing: destinations can be split. So I can give one part to a thread, and another part to another thread. If I used array-returning functions, both of these thread would allocate an array, and I would have to copy them into a new array on my way out. Vector depends on array-fusion to avoid such extra-allocation, but it's really not reliable (if these are really threads (_e.g._ `async`), vector fusion will not trigger). Another case where fusion will most certainly not work, is when you want to fill in a memory-mapped buffers, since it is not a normal vector thing.

Further reading:
- Using DPS for memory-efficient code generation: https://www.microsoft.com/en-us/research/publication/using-destination-passing-style-compile-functional-language-efficient-low-level-code/
- My Haskell Exchange 2017 talk speaks about destination quite a bit: https://skillsmatter.com/skillscasts/10637-distributed-programming-with-linear-types

Divesh-Otwani:
Closing in favor of #238.

PR #111, opened 2020-05-18
Doc/push pull arrays

Divesh-Otwani:
Finished documenting push-pull arrays.

This closes out the module documentation for linear base as it stands. I'll make sure that added stuff that I finish comes with documentation so we don't acquire a technical debt.

* I'll write a really minimal user guide because I'd rather just have it be tutorial based.
* I'm going to defer getting haddock with linear arrows working (since it requires me to build GHC which might end up being a time-sink) until we approach the deadline for releasing linear base (which I'd assume we're doing right alongside merging into GHC?)

### DIFF
--- a/src/Data/Array/Polarized.hs
+++ b/src/Data/Array/Polarized.hs
@@ -1,6 +1,103 @@
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 
+-- | This module provides converters and documentation for polarized arrays.
+--
+-- There are two polarities or \"directions\": push arrays and pull arrays.
+--
+-- You should use push an pull arrays if you want to explicity control
+-- when GHC allocates space in the garbage-collector heap for arrays.
+-- The point of doing this is to reduce the amount of memory used
+-- and improve the performance of your program.
+--
+-- The performance of your program is improved because linear types enable the
+-- API which does not allow operations on push or pull arrays that cannot be
+-- fused. That is, for some subtle and complicated reasons, we can be sure that
+-- only one function allocates memory (and it's called 'alloc') and all other
+-- compositions do not allocate memory.
+--
+-- == How to use push and pull arrays
+--
+-- __The trifecta of controlling allocation:__
+--
+-- \[ \texttt{Vector a} \]
+-- \[ \text{fromVector}~~ { \Huge \swarrow  ~~~~ \nwarrow } ~~~ \text{alloc} \]
+-- \[ \texttt{Pull a} {\Huge ~~~ \longrightarrow ~~~} \texttt{Push a}\]
+-- \[ \text{transfer}\]
+--
+-- Pull arrays are good as arguments to functions, because they are easy to
+-- \"pull\" from, map over and zip with.
+--
+-- Push arrays are good as values returned from functions, and are easy to
+-- \"write\" to and modify.
+--
+-- A typical application starts with concrete lists as @Vector@s (which are
+-- inputs), goes to @Pull@ arrays, then to @Push@ arrays and ends with
+-- @Vector@s (which are outputs).  All the computations inbetween the
+-- "fromVector" and "alloc" are fused and no memory from the GC heap is
+-- allocated when they are executed.
+--
+-- Such a combination of functions that controls fusion is called 
+-- a __fusion pipeline__.
+--
+-- == A toy example
+--
+-- > import Data.Array.Polarized
+-- > import qualified Data.Array.Polarized.Push as Push
+-- > import qualified Data.Array.Polarized.Pull as Pull
+-- > import qualified Data.Functor.Linear as Linear
+-- > import Data.Vector (Vector, (!), fromList)
+-- > import Prelude
+-- >
+-- > type Pull a = Pull.Array a
+-- > type Push a = Push.Array a
+-- >
+-- > pipeline :: Vector Int -> Vector Int -> Vector Int
+-- > pipeline coeff =
+-- >   pushToVector .
+-- >   diff .
+-- >   addCoeffVector .
+-- >   multThree .
+-- >   fromVector'
+-- >   where
+-- >     pushToVector :: Pull Int -> Vector Int
+-- >     pushToVector arr = Push.alloc (transfer arr)
+-- >
+-- >     fromVector' :: Vector Int -> Pull Int
+-- >     fromVector' arr = Pull.fromVector arr
+-- >
+-- >     addCoeffVector :: Pull Int -> Pull Int
+-- >     addCoeffVector arr = addCoeff (Pull.fromVector coeff) arr
+-- >
+-- > multThree :: Pull Int -> Pull Int
+-- > multThree arr = Linear.fmap (* 3) arr
+-- >
+-- > addCoeff :: Pull Int -> Pull Int %1-> Pull Int
+-- > addCoeff arr = Pull.zipWith (+) arr
+-- >
+-- > -- | Takes the difference between elements
+-- > -- which is common in differentials
+-- > diff :: Pull Int -> Pull Int
+-- > diff arr = withLen (Pull.findLength arr)
+-- >   where
+-- >     withLen :: (Int, Pull Int) -> Pull Int
+-- >     withLen (len, arr) =
+-- >       case (Pull.split 1 arr, Pull.split (len - 1) arr) of
+-- >         ((_, tail), (top,_)) -> Pull.zipWith (-) tail top
+-- >
+--
+--
+-- == Background for the interested
+--
+-- To understand a bit more about why and how these work, see these:
+--
+-- * http://www.cse.chalmers.se/~josefs/talks/LinArrays.pdf
+-- * https://www.sciencedirect.com/science/article/pii/030439759090147A
+-- * http://jyp.github.io/posts/controlled-fusion.html
+-- * http://lopezjuan.com/limestone/vectorcomp.pdf -- find this on an
+-- internet archive like https://archive.org/web/
+--
+--
 module Data.Array.Polarized
   ( transfer
   , walk
@@ -15,13 +112,10 @@ import Prelude.Linear
 
 import Data.Vector (Vector)
 
--- See:
+
+-- DEVELOPER NOTE:
 --
--- - http://lopezjuan.com/limestone/vectorcomp.pdf (available on the wayback
---   machine)
--- - http://jyp.github.io/posts/controlled-fusion.html
---
--- | The general spirit is: `Push.Array` are those arrays which are friendly in
+-- The general spirit is: `Push.Array` are those arrays which are friendly in
 -- returned-value position. And `Pull.Array` are those arrays which are friendly
 -- in argument position. If you have more than one array in an unfriendly
 -- position, you need to allocate (allocated arrays are friendly in all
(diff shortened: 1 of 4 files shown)


src/Data/Array/Polarized.hs
```diff
@@ -1,6 +1,103 @@
...
+--
+-- > import Data.Array.Polarized
+-- > import qualified Data.Array.Polarized.Push as Push
+-- > import qualified Data.Array.Polarized.Pull as Pull
+-- > import qualified Data.Functor.Linear as Linear
+-- > import Data.Vector (Vector, (!), fromList)
+-- > import Prelude
+-- >
+-- > type Pull a = Pull.Array a
+-- > type Push a = Push.Array a
```
aspiwack, line 53: Why not just inline these?

src/Data/Array/Polarized.hs
```diff
@@ -1,6 +1,103 @@
...
+--
+-- == Background for the interested
+--
+-- To understand a bit more about why and how these work, see these:
+--
+-- * http://www.cse.chalmers.se/~josefs/talks/LinArrays.pdf
+-- * https://www.sciencedirect.com/science/article/pii/030439759090147A
+-- * http://jyp.github.io/posts/controlled-fusion.html
+-- * http://lopezjuan.com/limestone/vectorcomp.pdf -- find this on an
+-- internet archive like https://archive.org/web/
```
aspiwack, line 98: Then give a wayback machine link.

src/Data/Array/Polarized.hs
```diff
@@ -1,6 +1,103 @@
...
+-- \[ \texttt{Vector a} \]
+-- \[ \text{fromVector}~~ { \Huge \swarrow  ~~~~ \nwarrow } ~~~ \text{alloc} \]
+-- \[ \texttt{Pull a} {\Huge ~~~ \longrightarrow ~~~} \texttt{Push a}\]
+-- \[ \text{transfer}\]
+--
+-- Pull arrays are good as arguments to functions, because they are easy to
+-- \"pull\" from, map over and zip with.
+--
+-- Push arrays are good as values returned from functions, and are easy to
+-- \"write\" to and modify.
```
aspiwack, line 32: We don't write to them, we construct them. They are not mutable arrays.

src/Data/Array/Polarized.hs
```diff
@@ -1,6 +1,103 @@
...
+-- > import qualified Data.Array.Polarized.Pull as Pull
+-- > import qualified Data.Functor.Linear as Linear
+-- > import Data.Vector (Vector, (!), fromList)
+-- > import Prelude
+-- >
+-- > type Pull a = Pull.Array a
+-- > type Push a = Push.Array a
+-- >
+-- > pipeline :: Vector Int -> Vector Int -> Vector Int
+-- > pipeline coeff =
```
aspiwack, line 56: Did you try to typecheck this example? It doesn't look correct to me.

src/Data/Array/Polarized/Pull.hs
```diff
@@ -1,21 +1,85 @@
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 
--- | This module is intended to be imported qualified, e.g.
+-- | This module provides pull arrays.
+--
+-- Use pull arrays to force fusion (i.e., force GHC to not allocate arrays in
+-- the garbage collector memory) in a computation that only uses lists
```
aspiwack, line 7: What does it have to do with lists?

src/Data/Array/Polarized/Pull.hs
```diff
@@ -1,21 +1,85 @@
...
+-- as /intermediate/ results without operations that \"write\" to a list.
+-- You should think about pull arrays as arrays you can \"pull\" or read from,
+-- zip with and map over. These are arrays that work nicely as arguments to
+-- functions and poorly as arrays to write to.
+--
+-- Pull arrays fit within a larger framework for controlling when
+-- garbage collector memory is allocated; please see "Data.Array.Polarized".
+--
+-- Import this module qualified for clarity and to avoid
+-- name clashes.
```
aspiwack, line 17: What's wrong with the prosaic, and less patronising, “This module is intented to be imported qualified.”?

src/Data/Array/Polarized/Pull.hs
```diff
@@ -1,21 +1,85 @@
...
 
--- | This module is intended to be imported qualified, e.g.
+-- | This module provides pull arrays.
+--
+-- Use pull arrays to force fusion (i.e., force GHC to not allocate arrays in
+-- the garbage collector memory) in a computation that only uses lists
+-- as /intermediate/ results without operations that \"write\" to a list.
+-- You should think about pull arrays as arrays you can \"pull\" or read from,
+-- zip with and map over. These are arrays that work nicely as arguments to
+-- functions and poorly as arrays to write to.
```
aspiwack, line 11: It's not that they work “poorly” as returned values, they are just less flexible there. If you have a pull array argument, you have quite a bit of the power of an allocated array. Whereas if you need to return one, you have quite a bit of restrictions.

src/Data/Array/Polarized/Pull.hs
```diff
@@ -1,21 +1,85 @@
...
+-- >   let nsum = normSumExample x_pull y_pull z_pull
+-- >   putStrLn P.$ "norm sum is " ++ show nsum
+-- >
+-- > normSumExample :: Pull Int -> Pull Int -> Pull Int -> Double
+-- > normSumExample x y z = norm (sumV (sumV x y) z)
+-- >
+-- > sumV :: Pull Int #-> Pull Int #-> Pull Int
+-- > sumV = Pull.zipWith (Linear.+)
+-- >
+-- > norm :: Pull Int -> Double
```
aspiwack, line 46: You should really avoid function which require unrestricted access to pull or push arrays. They are not suited for this. It's not that it can't work (though, maybe we should make it harder to get such unrestricted access), but if you indeed share a pull array, you are duplicating computation, so it's a rather bad idea.

Example should not encourage bad practices.

src/Data/Array/Polarized/Pull.hs
```diff
@@ -42,11 +106,12 @@ import qualified Unsafe.Linear as Unsafe
...
 asList = foldr (\x xs -> x:xs) []
 
--- | /!\ Partial! Only works if both arrays have the same length.
+-- | @zipWith f [x1,x2,...] [y1,y2,...] == [f x1 y1, f x2 y2, ...]@
+-- __Partial:__ Only works if both arrays have the same length.
 zipWith :: (a #-> b #-> c) -> Array a #-> Array b #-> Array c
 zipWith f x y = Data.fmap (uncurry f) (zip x y)
 
--- | Fold a pull array into a monoid.
+-- | Fold a pull array using a monoid.
```
aspiwack, line 114: Maybe:
```suggestion
-- | Fold over a pull array using a monoid.
```
What do you think?

src/Data/Array/Polarized/Pull.hs
```diff
@@ -42,11 +106,12 @@ import qualified Unsafe.Linear as Unsafe
 asList :: Array a #-> [a]
 asList = foldr (\x xs -> x:xs) []
 
--- | /!\ Partial! Only works if both arrays have the same length.
+-- | @zipWith f [x1,x2,...] [y1,y2,...] == [f x1 y1, f x2 y2, ...]@
+-- __Partial:__ Only works if both arrays have the same length.
```
aspiwack, line 110:
```suggestion
-- | @zipWith f [x1,x2,...,xn] [y1,y2,...,yn] = [f x1 y1, f x2 y2, ..., f xn yn]@
-- __Partial:__ `zipWith f [x1,x2,...,xn] [y1,y2,...,yp]` is an error if @n ≠ p@
```
(avoid non-sentences, also, `==` is the symbol for equality test, we are interested here in true equality. For instance this equation still works if the `xi` are functions, whereas it wouldn't make sense to call `==`, so use the symbol `=` in its mathematical sense)

src/Data/Array/Polarized/Pull/Internal.hs
```diff
@@ -31,17 +31,19 @@ data Array a where
...
+
+
+-- | Produce a pull array of lenght 1 consisting of solely the given element.
 singleton :: a #-> Array a
 singleton = Unsafe.toLinear (\x -> fromFunction (\_ -> x) 1)
 
--- | /!\ Partial! Only works if both arrays have the same length.
--- Zip both pull arrays together.
+-- | @zip [x1, x2, ...] [y1, y2, ...] == [(x1,y1), (x2,y2), ...]@
+-- __Partial__ Only works if both arrays have the same length.
```
aspiwack, line 46: See above.

src/Data/Array/Polarized/Pull/Internal.hs
```diff
@@ -61,41 +63,42 @@ make x n = fromFunction (const x) n
...
 
--- | Extract the length of an array, and give back the original array. This
--- is possible since getting the length of an array doesn't count as consuming
--- it.
+-- | Extract the length of an array, and give back the original array.
 findLength :: Array a #-> (Int, Array a)
 findLength (Array f n) = (n, Array f n)
 
--- | A constructor for pull arrays from a function and specified length.
+-- | @fromFunction arrIndexer len@ constructs a pull array given a function
```
aspiwack, line 78:
```suggestion
-- | @'fromFunction' arrIndexer len@ constructs a pull array given a function
```

Adding the quotes makes the name a link. It's not a link which goes far, but I like this convention :slightly_smiling_face: .

src/Data/Array/Polarized.hs
curiousleo: @Divesh-Otwani I can recommend doctests for this sort of thing. The setup is fiddly, but when it works, it's nice.

Example introduction of doctests into a codebase: https://github.com/idontgetoutmuch/random/pull/34/files

The doctest syntax is described here: https://github.com/sol/doctest

Divesh-Otwani: Thanks -- I did type check it but random stuff does go wrong so a doctest is really useful.

src/Data/Array/Polarized/Pull.hs
Divesh-Otwani: Yes, It doesn't need to be non-linear so I'll fix it.

src/Data/Array/Polarized/Pull.hs
Divesh-Otwani: Oh nothing -- Beth Aitman (in her talk about documentation) recommended directness and interestingly there was some discussion in the audience about whether imperatives like "do this" v.s. "this is a good idea" sounded impolite. It seemed to vary across cultures with the British and Asian at the most polite end and the Americans at the other. I'll lean towards a more European style going forward. (To my culturally-American ears "Import this module like this ..." sounds completely normal and not-patronizing.)

aspiwack: Actually, I don't think I am opposed to “Import this module qualified”. But the “for clarity and to avoid name clashes” sounds a bit petty and redundant, I don't know. Like assuming that the reader doesn't know what qualifying modules is for.

Or “Prefer importing this module qualified”. All this is fine.

As a side note, whether the intent for a module is to be imported qualified or unqualified will affect the design of its function names. This is the reason for the “intended to be imported qualified” phrasing that I had.

aspiwack:
@utdemir I'll leave it to you to review this PR, as well. Probably after @Divesh-Otwani has addressed the points above.

utdemir (approved):

PR #114, opened 2020-06-15
Zurihac work

Divesh-Otwani:
A few things from the zurihac talk.

### DIFF
--- a/examples/Simple/TopSort.hs
+++ b/examples/Simple/TopSort.hs
@@ -1,0 +1,118 @@
+{-# OPTIONS_GHC -Wno-name-shadowing #-}
+{-# OPTIONS_GHC -Wno-unused-matches #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
+
+
+module Simple.TopSort where
+
+import qualified Prelude.Linear as Linear
+import Prelude.Linear ((&))
+import Data.Unrestricted.Linear
+import qualified Data.HashMap.Linear as HMap
+import Data.HashMap.Linear (HashMap)
+import Data.Bifunctor.Linear (second)
+import Data.Maybe.Linear (catMaybes)
+import qualified Data.Functor.Linear as Data
+import Test.HUnit hiding (Node)
+
+
+-- # All Tests
+-------------------------------------------------------------------------------
+
+test1 :: Test
+test1 = topsort [(1,[2,3]), (2, [4]), (3,[4]), (4,[])] ~=?
+  [1,2,3,4]
+
+test2 :: Test
+test2 = topsort [(5,[2,0]), (4,[0,1]), (0,[]), (2,[3]), (3,[1]), (1,[])] ~=?
+  [5,2,3,4,0,1]
+
+test3 :: Test
+test3 = topsort
+  [ (1,[2]), (2,[4,5]), (3,[9,7]), (4,[7,8,10]), (5,[10]), (6,[10])
+  , (7,[]),(8,[]),(9,[]),(10,[])
+  ] ~=?
+    [1,2,4,8,5,3,9,7,6,10]
+
+topsortTests :: IO Counts
+topsortTests = runTestTT  $ TestList [test1, test2, test3]
+
+-- # The topological sort of a DAG
+-------------------------------------------------------------------------------
+
+type Node = Int
+type InDegGraph = HashMap Node ([Node], Int)
+
+topsort :: [(Node, [Node])] -> [Node]
+topsort = reverse . postOrder . fmap (  \(n,nbrs) -> (n,(nbrs,0))  )
+  where
+    postOrder :: [(Node, ([Node], Int))] -> [Node]
+    postOrder [] = []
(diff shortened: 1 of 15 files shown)


examples/Simple/TopSort.hs
```diff
@@ -0,0 +1,115 @@
...
+test2 = topsort
+  [(5,[2,0]), (4,[0,1]), (0,[]), (2,[3]), (3,[1]), (1,[])]
+-- [5,2,3,4,0,1]
+
+test3 :: [Int]
+test3 = topsort
+  [ (1,[2]), (2,[4,5]), (3,[9,7]), (4,[7,8,10]), (5,[10]), (6,[10])
+  , (7,[]),(8,[]),(9,[]),(10,[])
+  ]
+-- [1,2,4,8,5,3,9,7,6,10]
```
aspiwack, line 37: Maybe we should find a way to check the answers in tests?

src/Data/HashMap/Linear.hs
```diff
@@ -100,11 +100,13 @@ data RobinQuery k where
 -- # Construction and Modification
 --------------------------------------------------
 
+
+
 -- | Run a computation using a singleton hashmap
 singleton :: Keyed k =>
-  (k, v) -> (HashMap k v #-> Unrestricted b) -> Unrestricted b
+  (k, v) -> (HashMap k v #-> Unrestricted b) -> b
```
aspiwack, line 107: That's a bad idea. You may want to have a `singleton'` or what have you, which returns the non-unrestricted version. But you need `Unrestricted` in all generality, otherwise, you wouldn't be able to create a hashmap in the scope of another hashmap.

src/Data/HashMap/Linear.hs
```diff
@@ -224,14 +226,15 @@ member hmap k = memberFromQuery (queryIndex hmap k)
     memberFromQuery (h, IndexToSwap _ _) = (h, False)
     memberFromQuery (h, IndexToUpdate _ _) = (h, True)
 
-lookup :: Keyed k => HashMap k v #-> k -> (HashMap k v, Maybe v)
+lookup :: Keyed k => HashMap k v #-> k -> (HashMap k v, Unrestricted (Maybe v))
```
aspiwack, line 229: Asking the tough questions: do we want `Unrestricted (Maybe v)` or `Maybe (Unrestriced v)`, or, even, some maybe variant which has an unrestricted field?

src/Data/Set/Mutable/Linear.hs
```diff
@@ -40,14 +40,14 @@ type Keyed a = Linear.Keyed a
 -- # Constructors and Mutators
 -------------------------------------------------------------------------------
 
-singleton :: Keyed a => a -> (Set a #-> Unrestricted b) -> Unrestricted b
+singleton :: Keyed a => a -> (Set a #-> Unrestricted b) -> b
```
aspiwack, line 43: Same as above: let's not.

src/Data/Set/Mutable/Linear.hs
```diff
@@ -40,14 +40,14 @@ type Keyed a = Linear.Keyed a
...
 singleton a (f :: Set a #-> Unrestricted b) = Linear.singleton (a,()) f'
   where
     f' :: Linear.HashMap a () #-> Unrestricted b
     f' hmap = f (Set hmap)
 
 fromList :: HasCallStack =>
-  Keyed a => [a] -> (Set a #-> Unrestricted b) -> Unrestricted b
+  Keyed a => [a] -> (Set a #-> Unrestricted b) -> b
```
aspiwack, line 50: Same again.

src/Data/HashMap/Linear.hs
Divesh-Otwani: This is a tough question ... hmm

src/Data/HashMap/Linear.hs
Divesh-Otwani: Yes, I see.

examples/Simple/TopSort.hs
curiousleo: Doctests? _[Leo out]_

Divesh-Otwani: I wasn't able to set up doctests with linear base last time -- too many issues came up and it became a time sink. I just set up HUnit since it's simple enough.

aspiwack: Maybe later, then.

src/Data/HashMap/Linear.hs
aspiwack: Or maybe it should be the opposite, and `singleton` should be the `b` one, and `singleton'` the `Unrestricted b` one. Good bikeshedding problem.

My point is that we should not take away the more general version from the API.

src/Data/HashMap/Linear.hs
Divesh-Otwani: I'll make this an issue and defer the problem.

src/Data/HashMap/Linear.hs
Divesh-Otwani: Not sure which one should be which. Just defaulted to what was there before for now.

PR #119, opened 2020-06-26
fix: deprecated '-->.' notation

nbacquey:
Use 'FUN' from GHC.Exts
Instead of 'a -->.(p) b' we now have 'FUN p a b'


monoidal:
CI uses an older version of linear types, which does not export FUN yet.

facundominguez:
I don't have a clue of how to fix the current CI failure. Probably we should try GHC head and see if it occurs with it, before investigating further.

facundominguez:
I'm trying with GHC HEAD now.

Divesh-Otwani:
Hey, @facundominguez  just wanted to say thank you for this and the update to GHC HEAD!

facundominguez:
Kudos to @nbacquey for starting the PR.

Divesh-Otwani:
Thanks @nbacquey :)

PR #124, opened 2020-07-21
Rewriting hashmap implementation for correctness

### LINKED ISSUE #118: Add Data.Hashmap.withEmpty
The only way to create a HashMap is `singleton`, but it's awkward to use because in many settings you don't have the first value to initially populate your HashMap with. E.g. often one needs to make a HashMap out of a list, and the list might be empty.

Add `withEmpty :: Keyed k => (HashMap k v #-> Unrestricted b) -> Unrestricted b`. `with` in the name of the function is chosen to make it obvious to the reader that `withEmpty` expects a modifier. E.g. `withEmpty (fillFromList xs >>> runAlgorithm)`  as opposed to `empty (fillFromList xs >>> runAlgorithm)`. Similarly `singleton` should be renamed to `withSingleton`.

**Describe alternatives you've considered**
Dropping `singleton` altogether. Rejected because it doesn't seem necessary.

**Additional context**
N/A

Divesh-Otwani:
Rewrote hashmaps in a much better style with an empty constructor and fixed several deep and confusing bugs.

Fixes #118

### DIFF
--- a/src/Data/HashMap/Linear.hs
+++ b/src/Data/HashMap/Linear.hs
@@ -3,10 +3,13 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE MagicHash #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StrictData #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE UnliftedNewtypes #-}
 {-# OPTIONS_GHC -Wno-name-shadowing #-}
 {-# OPTIONS_GHC -Wno-incomplete-patterns #-}
 
@@ -14,27 +17,25 @@
 -- This module provides mutable hashmaps with a linear interface.
 --
 -- To use mutable hashmaps, create a linear computation of type
--- @HashMap k v #-> Unrestricted b@ and feed it to 'singleton' with
--- an initial key-value pair @(k,v)@.
+-- @HashMap k v #-> Unrestricted b@ and feed it to 'empty'.
 --
 -- This hashmap is implemented with robin hood hashing which has good average
 -- case performance.
 module Data.HashMap.Linear
   ( -- * A mutable hashmap
     HashMap,
+    Keyed,
+    Size(..),
     -- * Run a computation using a 'HashMap'
-    singleton,
-    singleton',
+    empty,
     -- * Modifiers and Constructors
-    alter,
     insert,
     delete,
     insertAll,
     -- * Accessors
     size,
-    member,
     lookup,
-    Keyed,
+    member
   )
 where
 
@@ -42,9 +43,10 @@ import Data.Array.Mutable.Linear
 import Data.Hashable
 import Data.Unrestricted.Linear
 import Prelude.Linear hiding ((+), lookup, read)
-import qualified Unsafe.Linear as Unsafe
 import Prelude ((+))
 import qualified Prelude
+import qualified Unsafe.Linear as Unsafe
+import GHC.Stack
 
 -- # Implementation Notes
 -- This is a simple implementatation of robin hood hashing.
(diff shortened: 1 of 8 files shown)


facundominguez:
Is there any account of what the deep and confusing bugs are and how this PR fixes them?

Divesh-Otwani:
@facundominguez A little bit is recorded here https://github.com/tweag/linear-base/issues/118#issuecomment-659490266

I could write more if needed but the basic idea is that I was running into situations where turning off and on -XStrict changed whether some tests passed. It seemed reasonable to conclude that the issue was bad uses of `toLinear`. I couldn't figure out where this was happening and it was easier to re-write the whole thing so I did that and those specific tests passed. There were a few bugs (which I solved, thankfully due to the ability to print things nicely) but these didn't involve some complex situation about the order of state changing behind some web of unsafe `toLinear` calls.

aspiwack:
A note: apparently “resolves” is not a magic keyword parsed by Github. I know that at least “fixes” and “closes” are.

src/Data/HashMap/Linear.hs
```diff
@@ -59,218 +58,185 @@ import qualified Prelude
...
-type Keyed k = (Eq k, Hashable k)
-
 -- | A mutable hashmap with a linear API
 data HashMap k v where
   -- | HashMap (size,count) array-of-robin-values
   -- The size is the amount of memory the array takes up.
   -- The count is the number of stored mappings.
   -- Our sparseness invariant: count*3 <= size.
-  HashMap :: (Int, Int) -> RobinArr k v #-> HashMap k v
+  HashMap :: Size -> Count -> RobinArr k v #-> HashMap k v
```
aspiwack, line 67: A note for later, but you probably want the size and count to be unpacked. It's likely to be better for performance.

src/Data/HashMap/Linear.hs
```diff
@@ -59,218 +58,185 @@ import qualified Prelude
...
-  Keyed k =>
-  HashMap k v #-> (Maybe v -> Maybe v) ->
-  k ->
-  HashMap k v
-alter = Unsafe.toLinear unsafeAlter
+-- | Run a computation given an empty hashmap
+empty :: forall k v b.
+  Keyed k => Size -> (HashMap k v #-> Unrestricted b) -> Unrestricted b
+empty sz@(Size s) scopeFn =
+  alloc s (-1, (error "key err val", error "val err val")) robinArrayScopeFn
```
aspiwack, line 116: There error messages are really cryptic. Just put a real sentence there. It becomes self-documenting. Better for the two error messages to be different indeed, as they will be easier to tell apart.

src/Data/HashMap/Linear.hs
```diff
@@ -59,218 +58,185 @@ import qualified Prelude
...
-alter ::
-  Keyed k =>
-  HashMap k v #-> (Maybe v -> Maybe v) ->
-  k ->
-  HashMap k v
-alter = Unsafe.toLinear unsafeAlter
+-- | Run a computation given an empty hashmap
+empty :: forall k v b.
+  Keyed k => Size -> (HashMap k v #-> Unrestricted b) -> Unrestricted b
+empty sz@(Size s) scopeFn =
```
aspiwack, line 115: Why `scopeFn` if `scope` would mean exactly the same thing?

src/Data/HashMap/Linear.hs
```diff
@@ -59,218 +58,185 @@ import qualified Prelude
...
-      (Int, Int) ->
-      Int ->
-      RobinVal k v ->
-      HashMap k v
-    swapFromRead (arr', (k', v', _)) sizes ix (k, v, psl) =
-      insert (HashMap sizes (write arr' ix (k, v, psl))) k' v'
-
--- | This deletes the key-value pair if it is present, and otherwise does
--- nothing.
+    robinArrayScopeFn :: RobinArr k v #-> Unrestricted b
```
aspiwack, line 118: Same question.

src/Data/HashMap/Linear.hs
```diff
@@ -59,218 +58,185 @@ import qualified Prelude
...
+    robinArrayScopeFn arr = scopeFn $ HashMap sz 0 arr
+
+-- | If the key is present, this update the value.
+-- If not, if there's enough space, insert a new pair.
+-- If there isn't enough space, error.
```
aspiwack, line 123: I'm confused. Why are we erroring when the hashmap is full? Instead of resizing?

src/Data/HashMap/Linear.hs
```diff
@@ -59,218 +58,185 @@ import qualified Prelude
...
+  (HashMap (Size s) ct arr, IndexToSwap (k',psl') ix') -> read arr ix' & \case
+    (arr'', Unrestricted (p_old, (k_old, v_old))) ->
+      tryInsertAtIndex
+        (HashMap (Size s) ct (write arr'' ix' (psl', (k',v))))
+        ((ix' + 1) `mod` s)
+        (k_old, p_old + 1)
+        v_old
+
+-- | This deletes the key-value pair if it is present.
+-- If it's not present, this does nothing.
```
aspiwack, line 147: When we are done with the review. I'll ask you do do a pass on the Haddock comments to remove all these “this”, and, generally, make the doc strings more precise and explicit. This can be a separate PR.

src/Data/HashMap/Linear.hs
Divesh-Otwani: No reason -- just my strange idiosyncrasies.

Divesh-Otwani:
All changes made -- re-requesting review :smile: @aspiwack

src/Data/HashMap/Linear.hs
```diff
@@ -59,218 +61,219 @@ import qualified Prelude
...
+    scope' :: RobinArr k v #-> Unrestricted b
+    scope' arr = scope $ HashMap sz (Count 0) arr
+
+-- | If the key is present, this update the value.
+-- If not, if there's enough space, insert a new pair.
+-- If there isn't enough space, resize and insert
+insert :: (Keyed k, HasCallStack) => HashMap k v #-> k -> v -> HashMap k v
+insert (HashMap sz@(Size s) ct@(Count c) arr) k v = case c < s of
+  False -> resizeMap (HashMap sz ct arr) & \case
+    !hmap' -> insert hmap' k v
```
aspiwack, line 138: It's not entirely clear from the context why you are using a `!` pattern here. But it looks like it may be very important. So add a comment explaining why you are forcing your result here. For the benefit of the future generations.

test/Test/Data/Mutable/Set.hs
```diff
@@ -29,13 +29,10 @@ mutSetTests = testGroup "Mutable set tests" group
...
 group =
   -- Tests of the form [accessor (mutator)]
-  [ testProperty "∀ x. member (sing x) x = True" memberSing1
-  , testProperty "∀ x,y/=x. member (sing x) y = False" memberSing2
-  , testProperty "∀ x. member (insert s x) x = True" memberInsert1
+  [ testProperty "∀ x. member (insert s x) x = True" memberInsert1
   , testProperty "∀ x,y/=x. member (insert s x) y = member s y" memberInsert2
   , testProperty "∀ x. member (delete s x) x = False" memberDelete1
   , testProperty "∀ x,y/=x. member (delete s x) y = member s y" memberDelete2
-  , testProperty "∀ x. size (sing x) = 1" sizeSing
```
aspiwack, line 38: This could easily be replaced by `size empty = 0, couldn't it?

src/Data/HashMap/Linear.hs
Divesh-Otwani: Whoops, left that in there -- there's no reason, just made it easier to put in print statements for debugging.

test/Test/Data/Mutable/Set.hs
Divesh-Otwani: Yes!

PR #126, opened 2020-07-29
Implement Data.{List,Foldable}.Linear

utdemir:
This PR implements the linear versions of `Data.List` and `Data.Foldable` modules. I am not a fan of the API yet, so this PR is just to open the discussion. Here are some notes:

* I found that there are usually a few ways to handle leftovers. Let's think about the function `any`:

  ```
  # Applies the predicate to all elements. After the first `True`, the rest is only to consume.
  any :: Foldable t => (a #-> Bool) -> t a #-> Bool
  # Applies the predicate until the first `True`, then uses the `Consumable` instance.
  any :: Consumable a => (a #-> Bool) -> t a #-> Bool
  # Applies the predicate until the first `True`, and return the leftovers.
  any :: (a #-> Bool) -> t a #-> (Bool, [a])
  ```

  From those three, the last one is the most generic, but unfortunately also the most confusing one. Especially when `t` is `[]`. I think it would be better to go with the generic one, but somehow make it clear that the return value is the leftovers. I am not sure how to do that, but maybe an explicit data type might work.

  I think I had this issue because I tried to replicate `Data.Foldable` function-by-function. But most functions there are not very useful in a linear context, such as `null`, `length` or `elem`. So, maybe it would be better to only implement some subset of functions on `Data.Foldable` and change `Data.List` functions type signatures to be more useful in a linear context.

* Especially `Data.List` comes with a lot of rewrite rules and annotations to make sure that the operations fuse. Replicating those might end up being a lot of work, especially since I do not completely understand how they work. In some cases we can't even `Unsafe.toLinear` them since our semantics are a bit different (we need to call `lseq` to forget the unused elements). So, this PR just implements the simplest versions.

* I do not think the linear `Foldable` can be derived from linear `Traversable`; as opposed to their base counterparts. I think the main reason is that `Foldable` gives you a way to get rid of the container, while `Traversable` doesn't. So, I could not implement `foldMapDefault` using only `Data.Traversable`. The closest I got was this:

  ```
  foldMapDefault :: (Traversable t, Monoid m, Consumable (t ())) => (a #-> m) -> t a #-> m
  foldMapDefault f xs = sequence ((\a -> (f a, ())) <$> xs)
    & \case (m, t) -> t `lseq` m
  ```

  which requires that ugly extra `Consumable (t ())` constraint. I think we can still make `Foldable` a superclass of `Traversable`, but I hesitated doing it without a discussion.

Do you think this is the right direction?

### DIFF
--- a/src/Data/Foldable/Linear.hs
+++ b/src/Data/Foldable/Linear.hs
@@ -3,8 +3,17 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
 
-module Data.Foldable.Linear where
+module Data.Foldable.Linear
+  ( Foldable (..)
+  , all
+  , and
+  , or
+  , traverse_
+  , sequence_
+  , foldMapDefault
+  ) where
 
 import qualified Prelude as Prelude
 import Prelude (Maybe(..), Either(..), Int, Ordering (..))
@@ -18,7 +27,6 @@ import Data.Maybe.Linear
 import Data.List.NonEmpty (NonEmpty (..))
 import Data.Functor.Linear
 import Data.Monoid.Linear
-import Data.Ord (Down(Down))
 
 class Foldable t where
     {-# MINIMAL foldMap | foldr #-}
@@ -43,10 +51,10 @@ class Foldable t where
         f' k x z = k $! f x z
 
     foldl :: (b #-> a #-> b) -> b #-> t a #-> b
-    foldl f z t = appEndo (getDual (foldMap (Dual . Endo . flip f) t)) z
+    foldl f z t = appEndo (getDual' (foldMap (Dual . Endo . flip f) t)) z
       where
-        getDual :: Dual a #-> a
-        getDual (Dual a) = a
+        getDual' :: Dual a #-> a
+        getDual' (Dual a) = a
 
     foldl' :: forall a b. (b #-> a #-> b) -> b #-> t a #-> b
     foldl' f z0 xs = foldr f' id xs z0
@@ -108,7 +116,7 @@ instance Foldable NonEmpty where
   foldMap f (x:|xs) = f x <> foldMap f xs
 
 instance Consumable a => Foldable (Either a) where
-  foldMap f (Left l) = l `lseq` mempty
+  foldMap _ (Left l) = l `lseq` mempty
   foldMap f (Right r) = f r
 
 maximumBy :: (Foldable t, Dupable a) => (a #-> a #-> Ordering) -> t a #-> a
@@ -155,6 +163,15 @@ sequence_ :: (Foldable t, Applicative f, Consumable a)
           => t (f a) #-> f ()
 sequence_ = traverse_ id
 
+
+-- | This function can be used as a value for 'foldMap' given a 'Traversable'
+-- and a 'Consumable' instance.
+--
+-- Note: Unlike their base counterparts, 'Data.Foldable.Linear' is not a
+-- superclass of 'Data.Traversable.Linear'; since there are 'Traversable's
+-- which are not 'Foldable', for example `(a,)` or `Const a`. This type
+-- signature also witnesses that by requring an extra `Consumable (t ())`
+-- constraint.
 foldMapDefault :: (Traversable t, Monoid m, Consumable (t ())) => (a #-> m) -> t a #-> m
 foldMapDefault f xs = sequence ((\a -> (f a, ())) <$> xs) & \case
   (m, t) -> t `lseq` m
--- a/src/Data/Foldable/Linear.hs
+++ b/src/Data/Foldable/Linear.hs
@@ -1,0 +1,160 @@
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Data.Foldable.Linear where
+
+import qualified Prelude as Prelude
+import Prelude (Maybe(..), Either(..), Int, Ordering (..))
+import Prelude.Linear.Internal.Simple
+import Data.Unrestricted.Linear
+import Data.Eq.Linear
+import Data.Ord.Linear
+import Data.Bool.Linear
+import Data.Num.Linear
+import Data.Maybe.Linear
+import Data.List.NonEmpty (NonEmpty (..))
+import Data.Functor.Linear
+import Data.Monoid.Linear
+import Data.Ord (Down(Down))
+
+class Foldable t where
+    {-# MINIMAL foldMap | foldr #-}
+
+    fold :: Monoid m => t m #-> m
+    fold = foldMap id
+
+    foldMap' :: Monoid m => (a #-> m) -> t a #-> m
+    foldMap' f = foldl' (\acc a -> acc <> f a) mempty
+
+    foldMap :: Monoid m => (a #-> m) -> t a #-> m
+    {-# INLINE foldMap #-}
+    foldMap f = foldr ((<>) . f) mempty
+
+    foldr :: (a #-> b #-> b) -> b #-> t a #-> b
+    foldr f z t = appEndo (foldMap (Endo . f) t) z
+
+    foldr' :: forall a b. (a #-> b #-> b) -> b #-> t a -> b
+    foldr' f z0 xs = foldl f' id xs z0
+      where
+        f' :: (b #-> b) #-> a #-> b #-> b
+        f' k x z = k $! f x z
+
+    foldl :: (b #-> a #-> b) -> b #-> t a #-> b
+    foldl f z t = appEndo (getDual (foldMap (Dual . Endo . flip f) t)) z
+      where
+        getDual :: Dual a #-> a
+        getDual (Dual a) = a
+
+    foldl' :: forall a b. (b #-> a #-> b) -> b #-> t a #-> b
+    foldl' f z0 xs = foldr f' id xs z0
+       where
+         f' :: a #-> (b #-> b) #-> b #-> b
+         f' x k z = k $! f z x
+
+    foldr1 :: forall a. (a #-> a #-> a) -> t a #-> a
+    foldr1 f xs = fromMaybe (Prelude.errorWithoutStackTrace "foldr1: empty structure")
+                    (foldr mf Nothing xs)
+      where
+        mf :: a #-> Maybe a #-> Maybe a
+        mf x Nothing = Just x
+        mf x (Just y) = Just (f x y)
+
+    foldl1 :: forall a. (a #-> a #-> a) -> t a #-> a
+    foldl1 f xs = fromMaybe (Prelude.errorWithoutStackTrace "foldl1: empty structure")
+                    (foldl mf Nothing xs)
+      where
+        mf :: Maybe a #-> a #-> Maybe a
+        mf Nothing y = Just y
+        mf (Just x) y = Just (f x y)
+
+    toList :: t a #-> [a]
+    toList = foldr (:) []
+
+    -- | This is likely not very useful and only here to preserve the symmetry
+    -- between 'Prelude.Foldable'.
+    null :: Consumable a => t a #-> Bool
+    null = foldr (\x a -> (x, a) `lseq` False) True
+
+    length :: Consumable a => t a #-> Int
+    length = foldl' (\c a -> a `lseq` c + 1) 0
+
+    -- | This is likely not very useful and only here to preserve the symmetry
+    -- between 'Prelude.Foldable'.
+    elem :: Prelude.Eq a => a -> t a #-> Bool
+    elem x = any (== x)
+
+    minimum :: (Dupable a, Prelude.Ord a) => t a #-> a
+    minimum = minimumBy compare
+
+    maximum :: (Dupable a, Prelude.Ord a) => t a #-> a
+    maximum = maximumBy compare
+
+instance {-# OVERLAPPABLE #-} (Foldable t, Consumable a) => Consumable (t a) where
+  consume = foldl' (flip lseq) ()
+
+instance Foldable [] where
+  foldr f z = \case
+    [] -> z
+    x:xs -> f x (foldr f z xs)
+
+instance Foldable Maybe where
+  foldMap _ Nothing = mempty
+  foldMap f (Just s) = f s
+
+instance Foldable NonEmpty where
+  foldMap f (x:|xs) = f x <> foldMap f xs
+
+instance Consumable a => Foldable (Either a) where
+  foldMap f (Left l) = l `lseq` mempty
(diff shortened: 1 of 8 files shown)


facundominguez:
It is looking to me that it would be easier to decide what to include if we had a few applications working with linear lists or similar. In the abstract, it is difficult to compare the alternatives.

aspiwack:
All good points.

I believe `any` should be your first suggestion (no leftover, no `Consumable`). It's the unfortunate consequence of the linear `(||)` being strict. Or at the very least, it's a similar phenomenon.

I can't think of a `Traversable` which wouldn't have `Consumable (t ())`, but there may be. It is certainly true that there is no control applicative which would let us drop the container to implement some fold. `Const m` is a data applicative. At some point, we considered using data applicatives in the definition of `Traversable`, but we didn't go for it.

@b-mehta do you remember why not? Do you have thoughts on any of the above?

b-mehta:
I don't recall completely, but we had a quick discussion here: https://github.com/tweag/linear-base/issues/48, which could be useful. As for a `Traversable` without `Consumable (t ())`, doesn't `(a,)` where `a` isn't consumable work? I could be wrong though, it's been a while since I've thought about these things.

utdemir:
> As for a `Traversable` without `Consumable (t ())`, doesn't `(a,)` where `a` isn't consumable work? 

Sorry @b-mehta, I do not understand what you mean.

As far as I can understand when implementing `foldMap`, we can effectfully map the values inside the container, but then we are still left with the container itself that we somehow need to consume. As @aspiwack mentioned `Const` would be a solution to that, but it doesn't type check becuse it can not be a `Control` functor.

@aspiwack , makes sense. I will make `Foldable` a superclass of `Traversable`.

aspiwack:
> > As for a `Traversable` without `Consumable (t ())`, doesn't `(a,)` where `a` isn't consumable work?
> 
> Sorry @b-mehta, I do not understand what you mean.


He says that

```haskell
instance Traversable ((,) a)
```

(even linearly)

But that you _don't_ have `instance Consumable (a,())` in general. This is very true. Thanks @b-mehta . In fact, this works with `Const a` as well.

And the argument for `Traversable` to use control applicative and not data applicative, is in fact similar.

This means, by the way, that `Foldable` oughtn't be a superclass of `Traversable`. Since `Const a` is `Traversable`, but not `Foldable` (in general). And so is `(,) a`. Quite fascinating all this.

src/Data/List/Linear.hs
```diff
@@ -0,0 +1,81 @@
...
+import Data.Functor.Linear
+import Data.List.NonEmpty (NonEmpty ((:|)))
+
+map :: (a #-> b) -> [a] #-> [b]
+map = fmap
+
+-- | @filter p xs@ returns a list with elements satisfying the predicate.
+--
+-- See 'Data.Maybe.Linear.mapMaybe' if you do not want the 'Dupable' constraint.
+filter :: Dupable a => (a #-> Bool) -> [a] #-> [a]
```
Divesh-Otwani, line 24: I think it's a good idea to have one with a consumable constraint instead -- using `a #-> (Bool, a)`. Some things might be easy to consume but hard to duplicate, like mutable arrays.

src/Data/Foldable/Linear.hs
```diff
@@ -0,0 +1,160 @@
...
+
+traverse_ :: (Foldable t, Applicative f, Consumable b)
+          => (a #-> f b) -> t a #-> f ()
+traverse_ f = foldr (\x k -> f x *> k) (pure ())
+
+sequence_ :: (Foldable t, Applicative f, Consumable a)
+          => t (f a) #-> f ()
+sequence_ = traverse_ id
+
+foldMapDefault :: (Traversable t, Monoid m, Consumable (t ())) => (a #-> m) -> t a #-> m
```
Divesh-Otwani, line 158: I think documenting a bit of the discussion of `Consumable t ()` is good here. Like how not all `Traversible` things are `Foldable` and this is in part seen by the fact that not all `Traversible` things `t` have a `Consumable (t ())` instance.

src/Data/Foldable/Linear.hs
```diff
@@ -0,0 +1,160 @@
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Data.Foldable.Linear where
```
Divesh-Otwani, line 7: Could you add explicit export lists?

src/Data/List/Linear.hs
```diff
@@ -0,0 +1,81 @@
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Data.List.Linear
```
Divesh-Otwani, line 6: Could you add explicit export lists? Also, maybe make an issue to copy/paste important haddoc from base itself.

src/Prelude/Linear/Internal/Simple.hs
```diff
@@ -53,15 +61,15 @@ uncurry f (x,y) = f x y
...
-foldr :: (a #-> b #-> b) -> b #-> [a] #-> b
-foldr f z = \case
-  [] -> z
-  x:xs -> f x (foldr f z xs)
-
-foldl :: (b #-> a -> b) -> b #-> [a] -> b
-foldl _ b [] = b
-foldl f b (x:xs) = foldl f (f b x) xs
-
+-- XXX: temporary: with multiplicity polymorphism functions expecting a
```
Divesh-Otwani, line 64: Thank you.

src/Data/List/Linear.hs
```diff
@@ -0,0 +1,81 @@
+{-# LANGUAGE LinearTypes #-}
```
Divesh-Otwani, line 1: Data.List can get long. Could you break this up into sections like in the base counterpart. Maybe look at the format of using `-- # Title` and a line of 90 `-------` characters like in [HashMap.hs]( https://github.com/tweag/linear-base/blob/master/src/Data/HashMap/Linear.hs).

src/Data/List/Linear.hs
utdemir: You are right, that is actually the reason I mentioned `Data.Maybe.Linear.mapMaybe` on the Haddock. Its type signature(`(a #-> Maybe b) -> [a] #-> [b]`) is similar to yours, but I think it's a bit nicer because with returning `Maybe a` we don't need to get `(False, a)` just to immediately forget the `a`. But if you think it's still necessary I am happy to add that function too.

Divesh-Otwani: Actually nevermind. Perhaps just mentioning that for consumable things, you can easily use `mapMaybe` in the haddock would be sufficient.

aspiwack:
Before merging this. I have a thought (it can be addressed in a future PR, but this thread is the perfect venue for this discussion). It occurred to me today. @b-mehta I'd like your opinion on this as well.

We've established that being “traversable” by `Control.Applicative`-s is a common occurrence, but doesn't imply `Foldable`.

However, being “traversable” by `Data.Applicative`-s, a less common occurrence, _does_ imply `Foldable` (indeed `Const m` is a `Data.Applicative` (for `m` a monoid)).

So, here is the thought. Why don't we give a name to the equivalent of `traverse`, but with `Data.Applicative` and add it to the `Foldable` type class (I'd call it `fold` but it's already taken, let's find another name)?

It gives a stark distinction between `Foldable` and `Traversable` in terms of `Data` and `Control`. It sort of makes sense, doesn't it? But maybe there are objections there that I'm not currently seeing. Probably even. So I'd like everybody's thought on the subject.

utdemir:
@aspiwack , 

I played around with a `Traversable` using `Data.Applicative`.

As you mentioned, it is a less common occurence; I couldn't implement instances for:  `(,) a`, `Const a` or `Either a`, I now understand that the reason is `Data.Applicative`'s possibility of duplicating those `a`'s.

It is quite interesting how big of a difference linearity can make. After spending a few hours thinking about this, it does make sense. On the other hand, it feels a bit daunting at times; having two different `Data.Functor.Traversable` typeclass each differing by the `Applicative` instance. It is confusing to me because I think practically they are very similar to each other. So, I personally am not keen to introduce the additional complexity, and my prefence would be to only have `Traversable` using `Control.Applicative`'s, and having a separate `Foldable` instance.

But I would also be happy to implement the new typeclass if you think it is worth it.

aspiwack:
The heart of my question was, rather, do we lose anything if we add this data-traversal as an additional method in foldable:

```haskell
class Functor t => Traversable t where
  …
  foldButReallyTraverseWellThisNeedABetterNameButWhatever :: Data.Applicative f => (a -> f b) -> t a -> f (t b)
```

Is there any functor of notice which supports `foldMap` (say) but not `foldButReallyTraverseWellThisNeedABetterNameButWhatever`?

Because I have this idea that this would be a good place for it.

aspiwack:
[above comment edited after @utdemir pointed out I wrote in several mistakes. Disregard the email and go read the comment on github directly]

utdemir:
We found a few `Foldable` instances which does not support

```
f :: Data.Applicative f => (a #-> f b) -> t a #-> f (t b)
```

eg. `Consumable a => (,) a`, `Consumable a => Either a`. They can only work if the constraint is `Movable` instead of `Consumable`, since they end up calling `Data.Applicative.pure :: a -> f a` (notice the unrestricted arrow).

src/Data/Foldable/Linear.hs
```diff
@@ -0,0 +1,177 @@
...
+    elem x = any (== x)
+
+    minimum :: (Dupable a, Prelude.Ord a) => t a #-> a
+    minimum = minimumBy compare
+
+    maximum :: (Dupable a, Prelude.Ord a) => t a #-> a
+    maximum = maximumBy compare
+
+instance {-# OVERLAPPABLE #-} (Foldable t, Consumable a) => Consumable (t a) where
+  consume = foldl' (flip lseq) ()
```
utdemir, line 104: Do you think this orphan instance is a good idea? I just observed that it causes confusing error messages when something is missing a `Consumable` instance, where you get something along the lines of `X is not a Foldable`. I think we can expose a function or a newtype wrapper instead.

aspiwack: No. It's probably not a good idea. I hadn't noticed it before.

Better is to make a deriving-via combinator.

utdemir:
I just noticed another interesting difference/. It looks like the `foldl` definition here and on the new `Data.Foldable` is a bit different:

```haskell
-- Prelude.Linear.Internal.Simple.foldl (removed with this PR)
foldl :: (b #-> a -> b) -> b #-> [a] -> b

-- Data.Foldable.Linear.foldl
foldl :: (b #-> a #-> b) -> b #-> t a #-> b 
```

It looks like the old one traverses a non-linear container with a linear accumulator, which might be useful on some cases, but the new one still makes sense to me.

Maybe we should have both. Unfortunately that requires us to find different names for them.

sjoerdvisscher:
For me `foldMap` is the central method of `Foldable`, all other methods derive from it using different `Monoid`s. So if `foldMap` is linear on the container, all other methods should be too.

utdemir:
> For me `foldMap` is the central method of `Foldable`, all other methods derive from it using different `Monoid`s. So if `foldMap` is linear on the container, all other methods should be too.

I agree. However, the old signature also does make sense by itself, in fact I just had a case where I required exactly that. I guess it's derivable from the second one by mapping the list elements to `Unrestricted`, but it's a hassle.

utdemir:
@sjoerdvisscher Thanks, yes, it looks like it is a job for `alaf`! Of course, that is given we implemented a linear version of it.

I now think since the new implementation already covers the old one with some wrapping and unwrapping around it, I am happy with only including the new one for now.

This PR is starting to diverge from `master`, so I want to try again to get it in. @aspiwack, it seems like there is no more discussion going on. We can ping some more people to look at it if you want to, or if you have any other comments I'd be happy to address them before we can merge this one.

aspiwack:
To be frank, I lost track of this discussion. What are the conclusions in your opinion?

utdemir:
@aspiwack,

In my opinion, the biggest open question here is linear `Foldable` not being a superclass of linear `Traversable`. The reason for that is `Traversable` using a `Control.Applicative` instead of `Data.Applicative` (https://github.com/tweag/linear-base/pull/126#issuecomment-668042694).

* We do not want reverting `Traversable` to use `Data.Applicative` instead, since it is a less common occurence. 
* We explored adding a method using `Data.Applicative`'s to `Foldable`, but then we are losing a few useful `Foldable` instances (https://github.com/tweag/linear-base/pull/126#issuecomment-669072136)

In short, the current state is that we have `Foldable` and `Traversable` classes but neither imply the other.

---

There are a few other minor discussions, mainly about choosing between alternative signatures; but we made some decisions there, and as @facundominguez said before we have some use cases it's difficult to choose between them (https://github.com/tweag/linear-base/pull/126#issuecomment-665668829).

One of my concerns we haven't addressed is this:

> Data.List comes with a lot of rewrite rules and annotations to make sure that the operations fuse. Replicating those might end up being a lot of work, especially since I do not completely understand how they work. In some cases we can't even Unsafe.toLinear them since our semantics are a bit different (we need to call lseq to forget the unused elements). So, this PR just implements the simplest versions.

If all above is good, I think this PR is good to merge. We can always revisit these decisions.

facundominguez:
> We do not want reverting Traversable to use Data.Applicative instead, since it is a less common occurence.

I found yesterday that this would have been useful to define
```Haskell
instance Dupable a => Dupable (Array a) where
  dupV = traverse dupV
```

aspiwack:
> We explored adding a method using `Data.Applicative`'s to `Foldable`, but then we are losing a few useful `Foldable` instances ([#126 (comment)](https://github.com/tweag/linear-base/pull/126#issuecomment-669072136))

We are losing instances, yes, but I'm not sure that we concluded that they were useful.

Maybe we just need the data-applicative-wise traversable in addition to all this. But, to be honest, I've never regarded `Foldable` as more than a historical type class with not a lot of meaning. Whereas data-applicable-wise traversable would make more sense to me. @mboes do you have an opinion on this matter?

utdemir:
> Maybe we just need the data-applicative-wise traversable in addition to all this. But, to be honest, I've never regarded Foldable as more than a historical type class with not a lot of meaning. Whereas data-applicable-wise traversable would make more sense to me.

I agree with this. So here's an idea. Let's side-step this discussion by removing `Foldable` from the equation entirely. I can modify this PR to just work on lists, putting everything in `Data.List.Linear`. This would already ease the pain of not having those utility functions, since most of our collections expose a `toList` method (of course, `Foldable` is just a glorified `toList` function). With the added advantage that we can merge this sooner. 

I am convinced that we will end up having a variant of Traversable-with-data-applicatives, so lets make a separate issue to discuss that (or if we indeed want `Foldable`).

aspiwack:
> I am convinced that we will end up having a variant of Traversable-with-data-applicatives

Me too. It does seem increasingly unavoidable.

Ok for making this a list-only issue and merging fast.

utdemir:
I'm closing this PR in favour of #190 , we can continue the discussion there.

Issue #127, opened 2020-08-01
The Big Bikeshedding Bonanza


Divesh-Otwani:
With the release of GHC 9.0, which will feature linear types, approaching; we are also preparing a release of linear-base.

And in order for this to happen, it is time to pay attention to some things which we had so far neglected. That is, stuff scoring high in the [Wadler's law](https://wiki.haskell.org/Wadler's_Law) scale. By virtue of them taking time, we have consistently put off thinking about naming too much, and ploughed forward with temporary names for many things, so that we could make progress on implementation.

But now is the time. We have the [bikeshedding label](https://github.com/tweag/linear-base/labels/Bikeshedding) to group these issues. So if you are into this sort of things, join us there and contribute your opinion.

utdemir:
Here are some, about the module structure:

* When I import a `*.Linear` module, I expect to only get linear functions. and the data types those functions work on. It is surprising  to me that I can import `Prelude.Linear` and it contains non-linear functions.
* In some cases using `*.Linear` modules require some data types from `base` to be imported (eg. `Data.Maybe.Linear` does not export `Maybe (..)`, or `Data.Ord.Linear` does not export `Ordering`), which is a bit annoying.
* `Data.Functor.Linear` vs. `Control.Monad.Linear` is a bit confusing when I need to import a `Functor`. I'd expect both `{Data,Control}.Functor.Linear` to exist. `Monad` can either be inside of `Control.Functor.Linear`, or we can have a separate `Control.Monad.Linear`.
* Some modules does not have `Linear` suffix: `System.IO.Resource`, `Foreign.Marshall.Pure`.

aspiwack:
@utdemir please make separate issues instead.

Divesh-Otwani:
> Here are some, about the module structure:
> 
>     * When I import a `*.Linear` module, I expect to only get linear functions. and the data types those functions work on. It is surprising  to me that I can import `Prelude.Linear` and it contains non-linear functions.
> 
>     * In some cases using `*.Linear` modules require some data types from `base` to be imported (eg. `Data.Maybe.Linear` does not export `Maybe (..)`, or `Data.Ord.Linear` does not export `Ordering`), which is a bit annoying.
> 
[...]

Issue #131, opened 2020-08-06
The name of Unrestricted


aspiwack:
This is something which has bothered everyone for months. It resurfaced recently in our internal discussion channels. We still don't have a clear solution.

The problem: `Unrestricted`, while a perfectly descriptive name, is very long. And considering how often this type is referred to in type signatures (and its constructor in function definition), we really want a shorter name. It's for everybody's sanity.

This naming issue can also be accompanied  by a reflection of how to name of the accompanying, and commonly useful

```haskell
data X s a where
  X :: s #-> a -> X s a
```
(the fields may be in the reverse order, I don't care much either way)

Here is a list of all the proposals that have been mooted that I can remember

- `Many`: but since it's already used for the name of the multiplicity, it invites many conflicts which are likely to be unpleasant.
- `Bang` after the fact that `Unrestricted a` is write `!a` in traditional linear logic literature. But it's fairly meaningless, if you don't know that notation.
- `Poly`: after _polynomial_ (linear: I must consume once, quadratic: I must consume twice, … polynomial: either of these). It does makes sense from a point of view of the mathematical analogy. And `Poly` means something like “several” in Greek, so it makes sense.
- `U`: first letter of `Unrestricted`. Easy to use, but a one-letter name may be consider a bit too rare and precious for this use.
- `This`: short, but ultimately hard to convey an intuition. It sounds a bit like it ought to mean the identity, which I don't think is a very good intuition. It could work if we have a story for how this makes sense. It does have the advantage that it comes with a reasonable name for `X`: ``s `AndThis` a``.
- `Lax`: mostly a synonymous of `Unrestricted`. But nobody's really convinced by it
- `Freely`, `Surely`: adverbs in the tradition of using such adverbs to pronounce modality (sometime `!a` is pronounced “certainly `a`”, see also “there merely exists” in homotopy type theory).
  - It could be a corresponding adjective, though `Free` is already taken, and `Sure` sounds weird. (like is “p is a mere proposition”)
  - Or an adverb of the same kind as `Maybe`
- `Box`: very generic
- `With`: this one has a bit of a twist, we would have 
   ```haskell
   type With :: Multiplicity -> Type -> Type
   data With p a where
     With :: a #p-> With p a
   ```
   So that `With 'Many a` is `Unrestricted a`. But since this last case is so common we would make a shorthand, such as `With'`.
- `More`: A bit like `Many`. Though it may sound a bit excessive, I guess.

[...]

PR #133, opened 2020-08-12
Make sure that *.Linear modules only export linear functions

utdemir:
I created this PR after my comment on #127:

> When I import a *.Linear module, I expect to only get linear functions. and the data types those functions work on. It is surprising to me that I can import Prelude.Linear and it contains non-linear functions.
> In some cases using *.Linear modules require some data types from base to be imported (eg. Data.Maybe.Linear does not export Maybe (..), or Data.Ord.Linear does not export Ordering), which is a bit annoying.

The main changes are

* `Prelude` re-export from `Prelude.Linear` is removed. Instead I only exported common data types like `Int` and `Bool`.
* Now `Data.Maybe.Linear`, `Data.Either.Linear` and `Data.Ord.Linear` exports `Maybe(..)`, `Either(..)` and `Ordering(..)` respectively.
* `Data.Semigroup.Linear` used to export non-linear destructors for newtype wrappers like `Sum`. I implemented linear versions of them manually.

The rest is pretty much just fixing the import lists and qualifying some names. I tried to not make any controversial changes, so let me know if you prefer anything different. However, this PR does cause the API to change quite a bit, sometimes in unexpected ways (eg. now `Prelude.Linear.Eq` is `Data.Eq.Linear.Eq` instead of `Prelude.Eq`); but I do not think we guarantee backwards compatibility yet.

### DIFF
--- a/examples/Foreign/List.hs
+++ b/examples/Foreign/List.hs
@@ -12,7 +12,7 @@ module Foreign.List where
 import qualified Data.List as List
 import Foreign.Marshal.Pure (Pool, Box)
 import qualified Foreign.Marshal.Pure as Manual
-import Prelude.Linear hiding (map, foldl, foldr)
+import Prelude.Linear hiding (foldl, foldr)
 
 -- XXX: we keep the last Cons in Memory here. A better approach would be to
 -- always keep a Box instead.
--- a/examples/Spec.hs
+++ b/examples/Spec.hs
@@ -17,13 +17,13 @@ import Foreign.List (List)
 import qualified Foreign.List as List
 import Foreign.Marshal.Pure (Pool)
 import qualified Foreign.Marshal.Pure as Manual
-import Prelude (return)
+import Prelude (return, IO, Show)
 import qualified Prelude as P
 import Prelude.Linear
 import Test.Hspec
 import Test.QuickCheck
 
-eqList :: forall a. (Manual.Representable a, Movable a, Eq a) => List a #-> List a #-> Unrestricted Bool
+eqList :: forall a. (Manual.Representable a, Movable a, P.Eq a) => List a #-> List a #-> Unrestricted Bool
 eqList l1 l2 =
     eqUL (move (List.toList l1)) (move (List.toList l2))
   where
@@ -64,7 +64,7 @@ main = hspec P.$ do
     describe "exceptions" P.$ do
       it "doesn't corrupt memory" P.$ do
         property (\(l :: [Int]) -> do
-          let l' = l ++ (throw InjectedError)
+          let l' = l P.++ (throw InjectedError)
           catch @InjectedError
             (P.void P.$ evaluate
                (Manual.withPool $ \pool ->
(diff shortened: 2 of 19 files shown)


aspiwack:
A few comments on spirit of all this, before I review:

- `Prelude.Linear` is meant to be a replacement for `Prelude`. More or less drop-in. So we should have a similar content. It's true that `Prelude` probably exports too many things. But we should make conscious decisions for what goes in and what goes out. That was the rationale for re-exporting everything and hiding things I didn't want. But a more principled approach would be to have a list of what is in `Prelude` and not in `Prelude.Linear`, and decide what to exclude and what to port.
-  It's alright for a non-linear function to be exported by a linear module. There is at least one (the constructor for `Unrestricted` (or whatever the name will end up being as per #131 ). When a linear variant of the function makes sense, it should be the exported function. But a function not having a linear arrow in it is not ground for exclusion.

utdemir:
My reasoning is only based on my experience developing on a few small linear modules, so not a very informed opinion; but here is what I think:

1. With the current state of the library, we can not use linear-functions as non-linear functions. This might change if we were to write all our functions in a multiplicty-polymorphic way.
2. An average application uses both linear functions and non-linear functions. This can be observed on most of our medium-sized modules on `linear-base` where we import both `Prelude` and `Prelude.Linear`. And I might actually even claim that in an average application there'll be much more non-linear functions than linear functions.

So, for now the reader needs to be aware of which functions are linear and which are not. One way to make it easier is importing `Prelude.Linear` qualified, then the `Linear.` (or `L.`) prefix clearly distinguishes linear functions. 

> Prelude.Linear is meant to be a replacement for Prelude. 

I think before most of our functions are linearity polymorphic, we can not reasonably expect people to not use `Prelude` at all. 

---

So, in short; I'd agree with re-exporting non-linear functions from linear-base once we're sure that our `Prelude.Linear` can provide the same functionality as `linear-base`. But even in that case, I'd be happier with an explicit export list.

mboes:
> With the current state of the library, we can not use linear-functions as non-linear functions. This might change if we were to write all our functions in a multiplicty-polymorphic way.

This is only true of higher-order functions, right?

utdemir:
> This is only true of higher-order functions, right?

To be honest, I did not know that, but it seems so :). In my defense, I was misled by the existence of `forget :: (a #-> b) #-> a -> b` function.

My point is now a bit weaker, but still stands since we are also exporting higher order functions.

aspiwack:
Mostly the intent is that you would import `Prelude.Linear` unqualified, and qualify base's `Prelude` when needed (though it may be the opposite in a module with very little linear types I suppose). But yes, things will be better when we can make things polymorphic.

But it doesn't deter the argument: linear modules export functions which are relevant to the stuff at hand. The `Unrestricted` data type comes with a decidedly unrestricted function: `Unrestricted :: a -> Unrestricted a`, as I was saying. Any module exporting a record will come with its projections, which are unrestricted as well. You shouldn't have to import a second module to import a function just because it doesn't have linear arrows in it.

However, the spirit is that _when a function can have linear arrows_ it should. In linear-base, I mean.

utdemir:
Okay. Then I'll modify this PR by re-adding the `Prelude` re-export but with an explicit export list, and while doing that I can note the functions we should implement with linear arrows.

utdemir:
We decided that we want to implement something like this after the other PR's are cleaned up, and with using small PR's with easier discussions. So I am closing this one.

PR #136, opened 2020-08-22
Edit pass on documentation

mboes:
- replaced the word "tools", which has little to no precedent, with
  "definitions" or "abstractions",
- use relative links where possible,
- remove sections with low information,
- fix license link,
- consistent indentation,
- consistent headings and capitalization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-Linear base
-===========
+# Linear base
 
 [![License MIT](https://img.shields.io/badge/license-MIT-brightgreen.svg)](https://github.com/tweag/linear-base/blob/master/LICENSE)
 [![Build status](https://badge.buildkite.com/5b60ab93dadba234a95e04e6568985918552dcc9e7685ede0d.svg?branch=master)](https://buildkite.com/tweag-1/linear-base)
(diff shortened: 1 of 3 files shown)


docs/DESIGN.md
```diff
@@ -1,107 +1,71 @@
...
+     (E.g., we decide there should be different laws for some type
+     class already in `base`.)
 
-  * Qualified imports that are meaningful and not abbreviations. So, for
-    instance, import `Data.Functor.Linear` as `Linear` and not as `F` for
-    functor.
-  * An export list for public modules.
-  * Module level documentation and haddock on key data definitions and top
-    level functions.
+## Conventions
```
Divesh-Otwani, line 61: I like this section!!!

docs/DESIGN.md
```diff
@@ -1,107 +1,71 @@
...
-General advice:
+We have established the following conventions in this project:
 
- * Sticking with the feel of the code already in linear base; make your work blend in.
- * Organize your code into simple sections.
- * Write tests for anything using unsafe functions (and tests in general).
+* use full words for Qualified imports, not abbreviations. For
+  instance, import `Data.Functor.Linear` as `Linear` and not as `F`
+  for functor.
+* All public modules have an export list.
```
Divesh-Otwani, line 68: :smile:

README.md
```diff
@@ -11,80 +10,54 @@ package that ships with GHC.
...
+
+## Getting started
 
-4. Add `linear-base` to your cabal file's `build-depends:` for the appropriate
-   library or executable.
+`-XLinearTypes` is not yet part of any GHC release. We recommend using
+Stack and Nix together to pull in an experimental version of GHC into
+your project. Use [this `stack.yaml`](./stack.yaml) as a starting
+point for your project. To learn more about Stack+Nix integration, see
+[here](https://docs.haskellstack.org/en/stable/nix_integration/).
```
Divesh-Otwani, line 35: @mboes It seems to me this is not friendly to new users. If I want to try playing with linear types and linear base, I shouldn't have to learn about using Stack and Nix together. I should be able to follow some simple instructions, install nix and stack if I have to and then just start coding. If putting a lengthy getting started section here isn't a good idea, I could make a template repository with minimal instructions there and link to it from here.

utdemir (approved):
LGTM. I think we can merge it as-is.

README.md
utdemir: Somebody recently reached out to me about how to use `LinearTypes` without using Nix. Other than compiling GHC manually, it seems like using `ghcup` with "custom bindists" (https://gitlab.haskell.org/haskell/ghcup-hs#installing-custom-bindists) is another option. So I was wondering if we should mention other methods here as well.

However, we are supposed to get a proper GHC release within a month or so (https://gitlab.haskell.org/ghc/ghc/-/wikis/status/ghc-9.0.1); so I personally don't worry much about making this section as beginner friendly as possible.

aspiwack: It's really not worth pursuing: we provide one preferred avenue that we know works (because we are testing it). Using an unreleased version of GHC is a power user thing, anyway, so we can assume that users will figure it out.

aspiwack (approved):

README.md
mboes: > I shouldn't have to learn about using Stack and Nix together

As @aspiwack says above, yes you should, if to start with you're willing to use an unreleased library that depends on an unreleased compiler featuring an experimental extension. As maintainers, we're not on the hook for providing multiple ways to build the library, with the risk of each system going out of sync with other because we're not testing or we are testing but the CI state space explodes. We're only on the hook for making sure *some* simple command exists that will reliably build the whole thing and execute the tests, and for making really sure that command does work.

Once GHC 9.0 ships, we should strive to have the library included in Stackage. From that point onwards, releases can be built like any other Cabal package. Releases should be frequent. And it will still be fair to support only one command for building and testing the `master` branch, for power users.

Divesh-Otwani: Hmm, I see why that puts too much on maintainers. In that case, I'll agree that it's important to put it on Stackage. It's much easier to make linear types mainstream if there's a low/non-existant barrier to entry.

PR #138, opened 2020-08-23
Move Prelude.Linear.Internal.Simple to Prelude.Linear.Internal

mboes:
Adjust the imports in other modules as well. However, this begs the
question: why are so many other modules importing the internal
version, instead of `Prelude.Linear`?


aspiwack:
I haven't reviewed, yet. But I can answer the question: `Prelude.Linear` is a prelude for the users of linear-base. It's public facing, and is allowed to depend on anything in linear-base. `Prelude.Linear.Internal.{,.Simple}` is the prelude of the library.

As it turns out, I believe that the `Prelude.Linear` is not as late in the dependency hierarchy as its namesake in base. But it's still pretty late.

mboes:
Makes sense. This should be documented.

aspiwack (approved):

PR #139, opened 2020-08-23
Shorten Prelude.Linear module documentation

mboes:


src/Prelude/Linear.hs
```diff
@@ -1,16 +1,13 @@
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 
--- |
--- This module defines all the facilities you need when writing linear code. As
--- an analog to the original 'Prelude', it has the basic tools any haskell
--- program using @-XLinearHaskell@ will probably need.
---
--- This module is designed to be imported unqualifed.
+-- | This module redefines 'Prelude' data types, functions and type classes
```
aspiwack, line 4: It doesn't “redfine data types”, so the sentence, as it stands, is a bit misleading. Doc is hard, I'll come back to it soon, and attempt a fix.

Divesh-Otwani:
Hope this works nicely ^

utdemir (approved):
LGTM. @aspiwack if you suggest an improvement over this, we can always send a separate PR.

aspiwack:
Just, we don't provide linear versions of the data types: they are the same data types (though we add a handful of new ones which only make sense in linear contexts).

PR #143, opened 2020-08-27
Make {Array,Vector}.Mutable.Linear lazier

### LINKED ISSUE #142: Data.Array.Mutable.Linear.Array is unnecessarily strict
**Describe the bug**

As a result of #135, `Data.Array.Mutable.Linear.Array` is stricter than necessary; `read`'s force the elements to WHNF. This is counter-intutitive since most other Haskell collections are lazy on the values.

**To Reproduce**

```haskell
{-# LANGUAGE LinearTypes #-}
{-# OPTIONS_GHC -fno-warn-name-shadowing #-}

module Main where

import Data.Function ((&))
import qualified Prelude.Linear as L
import Prelude.Linear (Unrestricted(..))

processArr :: Array.Array () #-> Unrestricted ()
processArr arr =
  Array.write arr 0 (error "this should not be evaluated") L.& \arr ->
  Array.read arr 0 L.& \(arr, Unrestricted _) ->
    arr `L.lseq` Unrestricted ()

main :: IO ()
main =
   Array.fromList [()] processArr
     & \(Unrestricted a) -> print a
```

**Expected behavior**

We expect the above program to print `()`, however it currently fails.

We should also have a test case for this.

utdemir:
Closes #142.

This PR changes wraps the result of `Unsafe.MutableArray.read` with an
unboxed tuple, so the array lookup can be evaluated without evaluating
the result.

Because of that, now `Data.{Array,Vector}.Mutable.Linear.read` functions
does not unnecessarily force the return value.

I added a test case about strictness issue just for `Array`'s. Because
Vector's `read` function returns the result linearly as opposed to
`Array` which wraps it with an `Unrestricted`. I think this is another
error, so I'll fix that with a separate PR:

```
read :: HasCallStack => Vector a #-> Int -> (Vector a, a)
read :: HasCallStack => Array a #-> Int -> (Array a, Unrestricted a)
```

---

While implementing this, I also noticed that the issue fixed in #135 also applies
to vectors, so this PR also fixes that while adding a test case.

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -127,7 +127,7 @@ read = Unsafe.toLinear readUnsafe
 -- | Using first element as seed, resize to a constant array
 resize :: HasCallStack => Int -> Array a #-> Array a
 resize newSize (Array _ mutArr) =
-  let (# a #) = Unsafe.readMutArr mutArr 0
+  let !(# a #) = Unsafe.readMutArr mutArr 0
   in  Array newSize (Unsafe.newMutArr newSize a)
 
 -- | Resize to a new constant array given a seed value
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -120,14 +120,15 @@ read = Unsafe.toLinear readUnsafe
     readUnsafe :: Array a -> Int -> (Array a, Unrestricted a)
     readUnsafe arr@(Array size mutArr) ix
       | indexInRange size ix =
-          let !val = Unsafe.readMutArr mutArr ix
-          in  (arr, Unrestricted val)
+          let !(# a #) = Unsafe.readMutArr mutArr ix
+          in  (arr, Unrestricted a)
       | otherwise = error "Read index out of bounds."
 
 -- | Using first element as seed, resize to a constant array
 resize :: HasCallStack => Int -> Array a #-> Array a
 resize newSize (Array _ mutArr) =
-  Array newSize (Unsafe.newMutArr newSize (Unsafe.readMutArr mutArr 0))
+  let (# a #) = Unsafe.readMutArr mutArr 0
+  in  Array newSize (Unsafe.newMutArr newSize a)
 
 -- | Resize to a new constant array given a seed value
 resizeSeed :: HasCallStack => Int -> a -> Array a #-> Array a
--- a/src/Unsafe/MutableArray.hs
+++ b/src/Unsafe/MutableArray.hs
@@ -57,9 +57,13 @@ writeMutArr mutArr (I# ix) val =
 
 -- | Read from a given mutable array arr, given an index
 -- in [0, size(arr)-1]
-readMutArr :: MutArr# a -> Int -> a
+--
+-- This returns an unboxed tuple to give the callee the
+-- ability to evaluate the function call without evaluating
+-- the final result.
+readMutArr :: MutArr# a -> Int -> (# a #)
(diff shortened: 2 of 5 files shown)


src/Data/Array/Mutable/Linear.hs
```diff
@@ -120,14 +120,15 @@ read = Unsafe.toLinear readUnsafe
     readUnsafe :: Array a -> Int -> (Array a, Unrestricted a)
     readUnsafe arr@(Array size mutArr) ix
       | indexInRange size ix =
-          let !val = Unsafe.readMutArr mutArr ix
-          in  (arr, Unrestricted val)
+          let !(# a #) = Unsafe.readMutArr mutArr ix
```
utdemir, line 123: I do not completely understand why we still need the bang pattern here, I thought pattern matching on the result would evaluate it even without the bang pattern. Without it, the read/write reordering issue reappears.

aspiwack: That's actually fairly surprising to me too. It does sound wrong. A `let` at an unboxed value should always be evaluated. I could understand if `let (# a #) = …` was illegal syntax, I suppose (we can make a case that there shouldn't be a lazy pattern-matching on unboxed tuple). But I wouldn't have expected it to fail to evaluate its right hand side.

I did find the explanation, though, it's in the last paragraph of the [relevant manual section](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-UnboxedTuples). Note how the desugaring breaks the strictness assumption by adding an extra thunk around the computation.

It may be best to avoid `let (#…` and replace them with explicit `case` to avoid surprises. But `let !p = …` is, of course, equivalent to a `case`.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -120,14 +120,15 @@ read = Unsafe.toLinear readUnsafe
...
+          in  (arr, Unrestricted a)
       | otherwise = error "Read index out of bounds."
 
 -- | Using first element as seed, resize to a constant array
 resize :: HasCallStack => Int -> Array a #-> Array a
 resize newSize (Array _ mutArr) =
-  Array newSize (Unsafe.newMutArr newSize (Unsafe.readMutArr mutArr 0))
+  let (# a #) = Unsafe.readMutArr mutArr 0
```
aspiwack, line 130: So we need
```suggestion
  let !(# a #) = Unsafe.readMutArr mutArr 0
```

src/Unsafe/MutableArray.hs
```diff
@@ -69,7 +73,8 @@ readMutArr mutArr (I# ix) =
 -- fails for a size smaller than the size of the given array.
 resizeMutArr :: MutArr# a -> Int -> MutArr# a
 resizeMutArr mutArr newSize =
-  resizeMutArrSeed mutArr (readMutArr mutArr 0) newSize
+  let (# a #) = readMutArr mutArr 0
```
aspiwack, line 76: And
```suggestion
  let !(# a #) = readMutArr mutArr 0
```

aspiwack (approved):

src/Data/Array/Mutable/Linear.hs
utdemir: To be honest, after I replaced the case with a `let` initially, I think the let binding looks clearer. However, I'm happy to replace it with an explicit `case` if that's what you prefer.

aspiwack: It doesn't matter for today. Let's see how often we get owned by this, before deciding for a solution.

PR #145, opened 2020-08-30
Update nixpkgs

### LINKED ISSUE #123: stack ghci panics
As of right now, calling `stack ghci` will panic when loading `Data.Unsafe.MutableArray`. It appears to be a regression related to `runRW#`.

The following code works in current ghci, and panics in the ghci version which is pinned in linear-base nix-shell:
```haskell
> :set -XMagicHash
> import GHC.Exts
> runRW# $ \s -> 0
```
As long as this bug persists, we can't use `ghcid` with linear-base, which is rather no bueno. (though I'm not sure it's the only hurdle for this particular race, currently)

This is probably [upstream's 18397](https://gitlab.haskell.org/ghc/ghc/-/issues/18397). It's being fixed. So this is a reminder to track said issue, and update our head as soon as the bug is fixed.

utdemir:
This PR updates nixpkgs to the latest revision which includes a newer
GHC including the [fix][] to #123.

Closes #123.

[fix]: https://gitlab.haskell.org/ghc/ghc/-/commit/2f0bae734e2dc8737fbbb8465de7ded89c1121b6

### DIFF
--- a/nixpkgs.nix
+++ b/nixpkgs.nix
@@ -1,7 +1,7 @@
 let
-  # 2020-08-22 master
-  rev = "d8e671676d91fa9d3ef64b69ba7680d07d79f7a7";
-  sha256 = "0bpf3qncb1qxg7dx4rgmwbhbi74lia6jdzjvmg8wvalc738p6wgw";
+  # 2020-08-31 master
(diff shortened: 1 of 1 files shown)


Divesh-Otwani:
Yay! :smiley:

aspiwack (approved):

Issue #144, opened 2020-08-27
Implement mutable Vector's using mutable Array's


utdemir:
Currently, `Data.Vector.Mutable.Linear` and `Data.Vector.Mutable.Array` have separate, but similar implementations. We can remove some duplication via implementing the `Vector` in terms of `Array`.

* Arrays should be ordinary arrays with `read` and `write` functions working with indexes and with an explicit `resize` functionality.
* Vectors should growable arrays supporting `push` and `pop` operations efficiently by maintaining some spare capacity.

Issue #130, opened 2020-08-06
Sources of uniqueness: the daring design debate


aspiwack:
[because alliterative titles are always the best titles]

This is something which has been at the back of my mind for a while. And I'm not exactly sure how to approach it.

Here is the problem: when creating some piece of data which needs to be unique (either for resource handling of or mutation or what have you) there are actually many ways to create such a piece of data.

In linear-base, we use two different ones:
- Delimit the data in a scope `withThingy :: (Thingy #-> Unrestricted a) #-> Unrestricted a`
- Use a specific control monad `newThingy :: M Thingy`
  - In this latter case, there are some restriction on the `run` function of the monad, if the `Thingy` is to stay unique
  - Typical examples are the `IO` monad, and its cousin the `RIO` monad.

These are not completely independent ways to create a unique thingy, since `withThingy` can be turned into the monadic style with the continuation monad. But they are not the same either, if only because `M` need not be a continuation monad.

Sometimes one of this styles is required and the other one doesn't really work. But sometimes you may want both. Maybe I can create a linear mutable array with the pure interface, but also with `IO`. Note how neither is an instance of the other. So I'd basically need two distinct creation functions. Meh…

But, more importantly, there are other ways to create unique thingy. For instance, if I have any guaranteed unique piece of data, I can create a new linear mutable array, and it will remain unique. And so there starts to be a bit of an `n*m` problem here. Where each type of unique pieces of data need many creation functions. That's unpleasant, to say the least.

So I've been thinking that data should be able to carry some kind of token establishing that they are unique. I've been calling them `Source` in my head (we can arrange it so that `Source` is 0-width data, so that they are mostly free).

The rest is a bit hazy in my head, still. But I imagine something like the following: we have a number of ways to build uniqueness sources
```haskell
newSource :: IO Source
withSource :: (Source #-> Unrestricted a) #-> Unrestricted a
```
Moreover sources would be `Dupable`, I assume (which means that they don't guarantee uniqueness, but I think that it is unavoidable).

Then we would have
```haskell
fromSource :: Source #-> Thingy
```
(and maybe `withThingy` and `newThingy` as convenient shorthands for the appropriate composition)

To be able to create a thingy out of, say, another thingy, without creating a new scope, we could use the following
```haskell
produceSource :: Thing #-> (Source, Thingy)
```
Implementing this is what requires `Source` to be `Dupable` (this, and whatever consume or freeze function that we would have on the type, though these only require `Consumable`).

[...]

PR #146, opened 2020-09-02
Implement mutable vectors using arrays

utdemir:
This PR refactors `Data.Vector.Mutable.Linear` to use `Data.Array.Mutable.Linear` instead of `Unsafe.MutableArray`. I also did a few refactors around the Array code.

In summary:

* I exposed `unsafeRead` and `unsafeWrite` functions from vectors and array to reduce unnecessary bounds-checking in vector code.
* I changed the type signatures so that the linear collection is the last parameter as opposed to the first. This might be controversial , but IMO it makes it nicer to compose the functions, eg:

```haskell
Vector.fromList l
  $ \vec ->
    vec
      Linear.& Vector.length
      Linear.& \(vec, Unrestricted oldLength) -> Vector.snoc 42 vec
      Linear.& Vector.length
      Linear.& getSnd
      Linear.& \(Unrestricted newLength) -> Unrestricted $ newLength === oldLength + 1
```

* `Array` had a few resize functions with counter-intutive semantics. I replaced them with three functions: `allocBeside`, `growBy` and `shrinkTo`. Let me know if you prefer something else.
* `MutableArray`, `Array` and `Vector` used to require at least one element inside. I removed this constraint. It seems like `MutableArray` can have 0 length, and therefore there is no reason `Array` and `Vector` to have that restriction.

So I'm sorry if this PR is a bit all around the place. I suggest you review the modules in this order:

* `Unsafe.MutableArray`
* `Data.Array.Mutable.Linear`
* `Data.Vector.Mutable.Linear.`
* tests
* `Data.HashMap.Linear` (only contains changes because of the parameter order change)

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -1,5 +1,5 @@
 {-# LANGUAGE BangPatterns #-}
-{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GADTSyntax #-}
 {-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE LambdaCase #-}
@@ -43,14 +43,18 @@ module Data.Array.Mutable.Linear
     Array,
     -- * Performing Computations with Arrays
     alloc,
+    allocBeside,
+    empty,
     fromList,
     -- * Mutators
     write,
-    resize,
-    resizeSeed,
+    unsafeWrite,
+    growBy,
+    shrinkTo,
     -- * Accessors
     read,
-    length,
+    unsafeRead,
+    size,
     toList,
   )
 where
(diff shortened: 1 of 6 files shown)


src/Data/Array/Mutable/Linear.hs
```diff
@@ -60,31 +64,42 @@ import GHC.Exts hiding (toList, fromList)
...
+  | size >= 0 = f (Array (Unsafe.newMutArr size val))
   | otherwise = error $ "Trying to allocate an array of size " ++ show size
 
--- | Allocate an array from a non-empty list (and error on empty lists)
+-- | Allocate a constant array given a size and an initial value,
+-- using another array as a uniqueness proof.
+allocBeside :: Int -> a -> Array b #-> (Array b, Array a)
+allocBeside size val orig
+  | size >= 0 = (orig, Array (Unsafe.newMutArr size val))
+  | otherwise = orig `lseq` error ("Trying to allocate an array of size " ++ show size)
```
utdemir, line 97: I find this function interesting. It is required since when implementing HashMap's, insertions can cause a new, bigger array to be created. However, we can not use `alloc` function there since it requires a continuation; and making HashMap insertions take a continuation would be unwieldy. 

I wanted to replace it with something more generic, but couldn't figure out how to do it nicely. I think this connects to @aspiwack 's suggestion on https://github.com/tweag/linear-base/issues/130.

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -58,93 +61,105 @@ import GHC.Exts hiding (fromList)
...
+    Array.size arr & \(arr', Unrestricted i) ->
+      f (Vec i arr')
 
 -- | Length of the vector
-length :: Vector a #-> (Vector a, Int)
-length = Unsafe.toLinear unsafeLength
-  where
-    unsafeLength :: Vector a -> (Vector a, Int)
-    unsafeLength v@(Vec (len, _) _) = (v, len)
+length :: Vector a #-> (Vector a, Unrestricted Int)
```
utdemir, line 98: I replaced this function to return an `Unrestricted Int` instead of `Int`, to be more consistent with other functions like `read`.

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -58,93 +61,105 @@ import GHC.Exts hiding (fromList)
...
-write = Unsafe.toLinear writeUnsafe
-  where
-    writeUnsafe :: Vector a -> Int -> a -> Vector a
-    writeUnsafe v@(Vec (len, _) mutArr) ix val
-      | indexInRange len ix = case Unsafe.writeMutArr mutArr ix val of () -> v
-      | otherwise = error "Write index not in range."
-
--- | Read from a vector, with an in-range index and error for an index that is
--- out of range (with the usual range @0..length-1@).
-read :: HasCallStack => Vector a #-> Int -> (Vector a, a)
```
utdemir, line 124: The new version of this function returns a `Unrestricted a` instead. Since the vector does not contain its values linearly, I think it was overly restrictive.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -60,31 +64,42 @@ import GHC.Exts hiding (toList, fromList)
...
+import qualified Control.Monad as Control
+import Prelude hiding (size, read)
 import qualified Prelude
 
 -- # Data types
 -------------------------------------------------------------------------------
 
 data Array a where
-  Array :: Int -> (MutableArray# RealWorld a) -> Array a
+  Array :: MutableArray# RealWorld a -> Array a
```
utdemir, line 76: Now the `Array` does not have an extra size parameter, since `MutableArray#` already comes with a `sizeof` method.

aspiwack:
>     * I changed the type signatures so that the linear collection is the last parameter as opposed to the first. This might be controversial , but IMO it makes it nicer to compose the functions

Whatever I think of this change. It didn't belong to this PR. Additionally, this is not the sort of change that you should make without talking about it first. We've all got better things to do than revert this in the course of this PR, though I do hope this sort of conflation won't happen again. But please open an issue to discuss the order of arguments in these functions. (under the bikeshedding label, I assume)

aspiwack:
> `Array` had a few resize functions with counter-intutive semantics. I replaced them with three functions: `allocBeside`, `growBy` and `shrinkTo`. Let me know if you prefer something else.

This should have been another PR.

utdemir:
I'm happy to spend some time sending separate PR's for each. Sorry about this.

utdemir:
I split this up to #147, #148 and #149. Alongside with the original #144, it should cover all the changes here.

PR #152, opened 2020-09-03
Reimplement resize and resizeSeed functions on linear arrays

### LINKED ISSUE #148: Array resizing functions are counter-intutive.
`Data.Array.Mutable.Linear` module comes with these functions:

```haskell
-- | Using first element as seed, resize to a constant array
resize :: HasCallStack => Int -> Array a #-> Array a
resize newSize (Array _ mutArr) =
  let !(# a #) = Unsafe.readMutArr mutArr 0
  in  Array newSize (Unsafe.newMutArr newSize a)

-- | Resize to a new constant array given a seed value
resizeSeed :: HasCallStack => Int -> a -> Array a #-> Array a
resizeSeed newSize seed (Array _ _) =
  Array newSize (Unsafe.newMutArr newSize seed)
```

I find them counter-intutive, since I don't expect a `resize` operation to lose most/all of the elements unless we are shrinking the array. I suggest we replace them with `growBy` and `shrinkTo` functions:

```haskell
-- | Grows the array by given number of elements, using a default value.
growBy :: HasCallStack => Int -> a -> Array a #-> Array a

-- | Shrinks the array to contain given number of elements.
shrinkTo :: HasCallStack => Int -> Array a #-> Array a
```

We can also provide a `resize` function can do both. Also; since these functions almost always copy the array, we can modify the signatures to return the original array too.

It seems like on the codebase `resizeSeed` function is used for allocating a new array using another array, without requiring a continuation. I think it is better done explicitly with a new alloc variant, something like:

```haskell
-- | Allocate a constant array given a size and an initial value,
-- using another array as a uniqueness proof.
allocBeside :: Int -> a -> Array b #-> (Array b, Array a)
allocBeside size val orig = (orig, Array (Unsafe.newMutArr size val))
```

I _think_ above is safe. Let me know if it isn't.

This issue is broken up from: https://github.com/tweag/linear-base/pull/146

utdemir:
Closes #148.

This PR modifies the resize function on linear arrays to 

* Copy over the elements to the new array instead of returning a constant array
* Support shrinking

In order to provide the same abilities, it also introduces an `allocBeside` function for allocating a new array without a continuation using another array as a uniqueness proof.

See #148 for more details.

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -135,10 +135,12 @@ read = Unsafe.toLinear readUnsafe
 -- value; resize the array to the given size using the seed value to fill
 -- in the new cells when necessary and copying over all the unchanged cells.
 --
+-- @
 -- let b = resize n x a,
 --   then length b = n,
 --   and b[i] = a[i] for i < length a,
 --   and b[i] = x for length a <= i < n.
+-- @
 resize :: HasCallStack => Int -> a -> Array a #-> Array a
 resize newSize seed (Array _ mut) =
   Array newSize (Unsafe.resizeMutArrSeed mut seed newSize)
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -43,11 +43,11 @@ module Data.Array.Mutable.Linear
     Array,
     -- * Performing Computations with Arrays
     alloc,
+    allocBeside,
     fromList,
     -- * Mutators
     write,
     resize,
-    resizeSeed,
     -- * Accessors
     read,
     length,
@@ -80,6 +80,13 @@ alloc :: HasCallStack =>
 alloc size x f
   | size > 0 = f (Array size (Unsafe.newMutArr size x))
   | otherwise = error $ "Trying to allocate an array of size " ++ show size
+
+-- | Allocate a constant array given a size and an initial value,
+-- using another array as a uniqueness proof.
+allocBeside :: Int -> a -> Array b #-> (Array b, Array a)
+allocBeside size val orig
+  | size > 0 = (orig, Array size (Unsafe.newMutArr size val))
+  | otherwise = orig `lseq` error ("Trying to allocate an array of size " ++ show size)
 
 -- | Allocate an array from a non-empty list (and error on empty lists)
 fromList :: HasCallStack =>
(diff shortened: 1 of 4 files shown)


src/Data/Array/Mutable/Linear.hs
```diff
@@ -124,16 +131,17 @@ read = Unsafe.toLinear readUnsafe
...
-resizeSeed newSize seed (Array _ _) =
-  Array newSize (Unsafe.newMutArr newSize seed)
+-- | Resize an array. That is, given an array, a target size, and a seed
+-- value; resize the array to the given size using the seed value to fill
+-- in the new cells when necessary and copying over all the unchanged cells.
+--
+-- let b = resize n x a,
+--   then length b = n,
+--   and b[i] = a[i] for i < length a,
+--   and b[i] = x for length a <= i < n.
```
aspiwack, line 141: I don't think that the formatting will be preserved by Haddock.

src/Unsafe/MutableArray.hs
```diff
@@ -68,29 +68,37 @@ readMutArr mutArr (I# ix) =
...
 resizeMutArrSeed mutArr x newSize = case newMutArr newSize x of
   newArr -> case copyIntoMutArr mutArr newArr of
     () -> newArr
 
--- | Copy the first mutable array into the second, larger mutable array.
--- This fails if the second array is smaller than the first.
+-- | Same as 'resizeMutArraySeed', but using the first index of the initial
+-- array as the seed value.
+resizeMutArr :: MutArr# a -> Int -> MutArr# a
+resizeMutArr mutArr newSize =
```
aspiwack, line 89: This means that there is a precondition: the resized array must be non-empty. You should write it explicitly.

Also, all functions with such precondition should have a `HasCallStack` constraint, as I argue in: https://www.tweag.io/blog/2020-04-16-exceptions-in-haskell/ .

test/Test/Data/Mutable/Array.hs
```diff
@@ -227,7 +230,21 @@ lenResizeSeedTest :: Int -> Int -> ArrayTester
...
     (move newSize)
-    (move Linear.$ getSnd Linear.$ Array.length (Array.resizeSeed newSize val arr))
+    (move Linear.$ getSnd Linear.$ Array.length (Array.resize newSize val arr))
+
+resizeRef :: Property
+resizeRef = property $ do
+  l <- forAll list
+  n <- forAll $ Gen.int (Range.linear 0 (length l * 2))
+  let x = 42
+      expected = take n $ l ++ repeat x
```
aspiwack, line 240: Maybe `cycle l` would make a slightly better quality infinite list.

aspiwack (approved):

src/Data/Array/Mutable/Linear.hs
utdemir: Thanks, I will wrap it with a code block.

test/Test/Data/Mutable/Array.hs
utdemir: That `repeat` is there to give that expression the same semantics as `Array.resize`; using `cycle l` there would fail the test case when the array is growing.

aspiwack: Got it. Then, maybe you want to generate `x` randomly. Just in case.

src/Unsafe/MutableArray.hs
utdemir: None of the `Unsafe.MutableArray` functions had `HasCallStack` constraint, although most of them were partial. So I enabled call stacks for the other function too.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -81,6 +81,13 @@ alloc size x f
   | size > 0 = f (Array size (Unsafe.newMutArr size x))
   | otherwise = error $ "Trying to allocate an array of size " ++ show size
 
+-- | Allocate a constant array given a size and an initial value,
+-- using another array as a uniqueness proof.
+allocBeside :: Int -> a -> Array b #-> (Array b, Array a)
```
facundominguez, line 86:
```suggestion
-- using another value as a uniqueness proof.
allocBeside :: Int -> a -> b #-> (b, Array a)
```
Does it matter if the source of uniqueness is an Array or something else?

facundominguez: I'm finding this aspect of dealing with linear types rather slippery. It may be just me, though.

utdemir: > I'm finding this aspect of dealing with linear types rather slippery. It may be just me, though.

I do agree. This `allocBeside` function is pretty arbitrary here, as @aspiwack mentioned there is a lot of possible other types that would work just as well (`Vector.Mutable.Linear` and our other collections is the first that came to my mind), so there's another "m*n" problem here. Another arbitrary decision is that only `Array` provides this function, while its likes with the similar properties doesn't.

However, I do think we should use #130 for this discussion and this function could be refactored later.

Issue #100, opened 2020-04-27
Allow Empty Mutable Linear Data Structures


Divesh-Otwani:
It's possible to have some empty mutable linear data structures but `unsafeCoerce`ing unit or something, and internally making sure nothing crashes.

utdemir:
I think #153 closes this one too.

PR #153, opened 2020-09-03
Allow zero length mutable arrays and vectors

### LINKED ISSUE #149: Mutable vectors (and possibly arrays) should support having zero elements inside
Currently, our linear mutable `Array` and `Vector` implementations require the collection to always contain at least one element inside. 

Especially when using a `Vector`, it is common to start with an empty vector and append new elements to it. Having it require an element to start with is unusual, so we should support having zero elements inside (this includes resizing to zero elements) and provide an `empty` method.

I think this choice stems from `Unsafe.MutableArray.newMutArr`'s precondition that the size should be positive. However, I did a few quick tests and it seems to be happy when the size is 0 too. So we can probably relax that requirement and the `Vector` and `Array` changes would follow.

This issue is broken up from: #146

utdemir:
Closes #149.

This PR relaxes the precondition that `Unsafe.MutableArray`, `Data.Array.Mutable.Linear` and `Data.Vector.Mutable.Linear` collections should contain at least one element inside to allowing them to contain zero or more elements.

Also, it removes the `resizeMutArr` function on `Unsafe.MutableArray` where it was using the first element when growing an array.  Now it is replaced by `resizeMutArrSeed`, which takes an explicit default value. This is an API change, however I do not think we try to preserve backwards compatibility yet.

Tests seem to be happy, and I added a new case testing `snoc` on an empty vector. 

This PR depends on #152 , so we should merge that one first. It also means that GitHub includes those changes in the diff view, so you might want to only look at the second commit.

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -75,11 +75,11 @@ data Array a where
 -------------------------------------------------------------------------------
 
 -- | Allocate a constant array given a size and an initial value
--- The size must be greater than zero, otherwise this errors.
+-- The size must be non-negative, otherwise this errors.
 alloc :: HasCallStack =>
   Int -> a -> (Array a #-> Unrestricted b) #-> Unrestricted b
 alloc size x f
-  | size > 0 = f (Array size (Unsafe.newMutArr size x))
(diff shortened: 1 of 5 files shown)


utdemir:
Just noticed, this also closes #100.

Divesh-Otwani (approved):

PR #154, opened 2020-09-03
Consume the continuation linearly when allocating linear collections

### LINKED ISSUE #150: Consume the continuation linearly when allocating linear collections
Currently, our mutable `Vector`, `Array` and `HashMap` constructors are of this form:

```haskell
alloc :: ... -> (Thing a #-> Unrestricted b) -> Unrestricted b
```

This is too ~~general~~ specific; since it doesn't mention that the continuation will only be used once, we can not use a linear value from the outer scope inside the continuation. Consider this example:

```haskell
data Thing = Thing
instance Consumable Thing where consume Thing = ()

withThing :: (Thing #-> Unrestricted b) -> Unrestricted b
withThing f = f Thing

doSomething :: Unrestricted Int
doSomething =
  withThing (\t1 ->
    withThing (\t2 ->
      t1 `lseq` t2 `lseq` Unrestricted 42
    )
  )
```

This does not compile, since `t1` is used inside the inner continuation, however the result of the inner `withThing` does not guarantee that it'll use the computation once.

To fix this, we should replace them with the form:

```haskell
alloc :: ... -> (Thing a #-> Unrestricted b) #-> Unrestricted b
```

utdemir:
Closes #150 .

This PR generalizes the collection allocation functions where we delimit the scope of the linear collection with a continuation by consuming the continuation linearly. This allows nesting those continuations. See #150 for an example where the previous signature fails.

While doing this change, I had to inline a few functions to make the typechecker checker happy, although I am not 100% sure why they did not work as-is. However, I hope those changes are not a big deal.

Another noteworthy issue is about the partial functions. Most allocation functions are partial, they panic on things like negative sizes. In those cases, we now can not simply ignore the continuation. So, I am doing this ugly trick in a few places:

```haskell
alloc :: HasCallStack => Int -> a -> (Array a #-> Unrestricted b) #-> Unrestricted b
alloc size x f
  | size > 0 = ...
  | otherwise =
    (error ("Trying to allocate an array of size " ++ show size) :: x #-> x)
    (f undefined)
```

That looks pretty hacky to me. I guess an alternative is to implement a new function like:

```haskell
consumeViaError :: String -> a #-> b
consumeViaError = error
```

But that doesn't look so nice either. Especially because the only reason to have it is the lack of inference of linear functions. I'd be happy to hear about alternative solutions here.

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -76,20 +76,24 @@ data Array a where
 -- | Allocate a constant array given a size and an initial value
 -- The size must be greater than zero, otherwise this errors.
 alloc :: HasCallStack =>
-  Int -> a -> (Array a #-> Unrestricted b) -> Unrestricted b
+  Int -> a -> (Array a #-> Unrestricted b) #-> Unrestricted b
 alloc size x f
   | size > 0 = f (Array size (Unsafe.newMutArr size x))
-  | otherwise = error $ "Trying to allocate an array of size " ++ show size
+  | otherwise =
+    (error ("Trying to allocate an array of size " ++ show size) :: x #-> x)
+    (f undefined)
 
 -- | Allocate an array from a non-empty list (and error on empty lists)
 fromList :: HasCallStack =>
-  [a] -> (Array a #-> Unrestricted b) -> Unrestricted b
-fromList [] _ = error $ "Trying to allocate from an empty list."
+  [a] -> (Array a #-> Unrestricted b) #-> Unrestricted b
+fromList [] f =
+  (error "Trying to allocate from an empty list." :: x #-> x)
+  (f undefined)
 fromList list@(x:_) (f :: Array a #-> Unrestricted b) =
-  alloc (Prelude.length list) x insertThenf
+  alloc (Prelude.length list) x (\arr -> f (insert arr))
   where
-    insertThenf :: Array a #-> Unrestricted b
-    insertThenf arr = f (doWrites (zip list [0..]) arr)
+    insert :: Array a #-> Array a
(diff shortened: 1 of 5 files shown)


aspiwack:
> So, I am doing this ugly trick in a few places

No, I'd say that it's fine. It's not even ugly. It would be quite difficult to teach the compiler about `error`'s bottoming behaviour directly.

In principle, a more principled solution would be `case error "blah" of {}`. But I'm not sure empty cases currently works as intended. The implementation is a bit fragile.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -76,20 +76,24 @@ data Array a where
...
 fromList :: HasCallStack =>
-  [a] -> (Array a #-> Unrestricted b) -> Unrestricted b
-fromList [] _ = error $ "Trying to allocate from an empty list."
+  [a] -> (Array a #-> Unrestricted b) #-> Unrestricted b
+fromList [] f =
+  (error "Trying to allocate from an empty list." :: x #-> x)
+  (f undefined)
 fromList list@(x:_) (f :: Array a #-> Unrestricted b) =
-  alloc (Prelude.length list) x insertThenf
+  alloc (Prelude.length list) x (\arr -> f (insert arr))
```
aspiwack, line 93: Right, so what's happening here is that `let` and `where` are always unrestricted binders. This is not by design, just a byproduct of the current linear types being an mvp. It'll get better.

aspiwack (approved):

src/Data/Array/Mutable/Linear.hs
utdemir: Got it, thanks!

PR #155, opened 2020-09-03
Use Unrestricted on Array.toList and Vector.read

utdemir:
Mutable arrays and vectors are linear on the collection itself, not on the values. So, after we read a value it should be unrestricted. Returning the elements linearly restricts their usage unnecessarily, especially when the elements are not `Movable`.

Before:

```haskell
toList :: Array a #-> (Array a, [a])
read :: HasCallStack => Vector a #-> Int -> (Vector a, a)
```

After:

```haskell
toList :: Array a #-> (Array a, Unrestricted [a])
read :: HasCallStack => Vector a #-> Int -> (Vector a, Unrestricted a)
```

The rest of the PR is pretty much following the types. I decided to create a PR because the change was small. Happy to close it if there is a good reason for keeping the old behaviour.

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -58,9 +58,10 @@ where
 import Data.Unrestricted.Linear
 import GHC.Exts hiding (toList, fromList)
 import GHC.Stack
+import qualified Data.Functor.Linear as Control
 import qualified Unsafe.Linear as Unsafe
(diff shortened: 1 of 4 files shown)


aspiwack (approved):
This was the right change to do.

Generally speaking, you are right to propose a change together with its code if its small and isolated. My complaint earlier was about undiscussed changes buried in a large PR.

Issue #164, opened 2020-09-05
Expand the linear HashMap interface


Divesh-Otwani:
Linear HashMap currently only provide simple lookup and insert functions. We should have some more primitives available so that at least other use cases can be implemented externally when necessary.

We can use [Data.HashMap.Strict](https://hackage.haskell.org/package/unordered-containers-0.2.12.0/docs/Data-HashMap-Strict.html) module from `unordered-containers` as a reference.

At minimum:

* An `alter`-like function to handle key collisions.
* `fromList` and `toList`.
* Applicable typeclass instances like `Functor`, `Consumable` and `Dupable`.
* `filter`-like functions.
* Functions merging two `HashMap`'s (`union`, `intersection`, `mergeWithKey`).

PR #176, opened 2020-09-08
Internal HashMap improvements

utdemir:
This PR (hopefully objectively) improves HashMap's internals so that our planned improvements on #164 can be implemented with less changes to the code.

It has three standalone changes each have a separate commit. Unfortunately they all change the same area of the code, so I decided to create a single commit instead of having to solve merge conflicts next. But, ideally, you should be able to review each PR (from older to newer) relatively independently.

The commits have detailed messages about the changes. The only API change is removing the `Size` wrapper which only affects the `empty` function, without causing any behaviour change.

The same tests still pass, but it is a tricky piece of code and I am not 100% sure about the correctness yet. I will spend some time on adding some more tests in a separate PR.

### DIFF
--- a/src/Data/HashMap/Linear.hs
+++ b/src/Data/HashMap/Linear.hs
@@ -83,11 +83,18 @@ newtype Count = Count Int
   deriving (Prelude.Num)
 
 -- | An array of Robin values
-type RobinArr k v = Array (RobinVal k v)
+type RobinArr k v = Array (Maybe (RobinVal k v))
 
 -- | Robin values are triples of the key, value and PSL
 -- (the probe sequence length).
-type RobinVal k v = (PSL, (k, v))
+data RobinVal k v = RobinVal {-# UNPACK #-} !PSL k v
+  deriving (Show)
+
+incRobinValPSL :: RobinVal k v -> RobinVal k v
+incRobinValPSL (RobinVal (PSL p) k v) = RobinVal (PSL (p+1)) k v
+
+decRobinValPSL :: RobinVal k v -> RobinVal k v
+decRobinValPSL (RobinVal (PSL p) k v) = RobinVal (PSL (p-1)) k v
 
 -- | A probe sequence length
 newtype PSL = PSL Int
@@ -117,11 +124,7 @@ empty :: forall k v b.
 empty sz@(Size s) scope =
   alloc
     s
-    ( -1
-    , ( error "reading error hashmap key"
-      , error "reading error hashmap val"
-      )
-    )
+    Nothing
     (\arr -> scope (HashMap sz (Count 0) arr))
 
 -- | If the key is present, this update the value.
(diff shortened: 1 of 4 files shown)


src/Data/HashMap/Linear.hs
```diff
@@ -63,32 +62,34 @@ import GHC.Stack
...
-newtype Size = Size Int
-  deriving (Prelude.Num)
-
 -- | The number of pairs stored in the hashmap
 newtype Count = Count Int
   deriving (Prelude.Num)
 
 -- | An array of Robin values
-type RobinArr k v = Array (RobinVal k v)
+type RobinArr k v = Array (Maybe (RobinVal k v))
```
Divesh-Otwani, line 81: If you're changing the representation you need to state the invariants, even if obvious:

> Each cell is `Nothing` if empty and is a `RobinVal` with the correct PSL otherwise.

src/Data/HashMap/Linear.hs
```diff
@@ -216,62 +216,65 @@ insertAll ((k, v) : xs) hmap = insertAll xs (insert hmap k v)
...
-      False -> probeFrom (k, p+1) ((ix+1)`mod` s) (HashMap sz ct arr')
-
+  (k, PSL) -> Int -> HashMap k v #-> (HashMap k v, ProbeResult k v)
+probeFrom (k, p) ix (HashMap ct arr) = Array.read arr ix & \case
+  (arr', Ur Nothing) ->
+    (HashMap ct arr', IndexToInsert p ix)
+  (arr', Ur (Just robinVal'@(RobinVal psl k' v'))) ->
+    case k == k' of
+      -- Note: in the True case, we must have p == psl
+      True -> (HashMap ct arr', IndexToUpdate v' psl ix)
```
Divesh-Otwani, line 226: Does the `IndexToUpdate` constructor have documentation that the value and PSL it holds is of the cell already there and not for the query?

src/Data/HashMap/Linear.hs
```diff
@@ -216,62 +216,65 @@ insertAll ((k, v) : xs) hmap = insertAll xs (insert hmap k v)
...
+      False -> case psl < p of
+        True -> (HashMap ct arr', IndexToSwap robinVal' p ix)
```
Divesh-Otwani, line 228: See this is confusing. The PSL `p` is of the query but in the `IndexToUpdate` constructor, the PSL is of the cell already there. Also, the `RobinVal` is of the current existing cell. All of this makes sense, it should just be documented in comments on the constructors.

src/Data/HashMap/Linear.hs
```diff
@@ -216,62 +216,65 @@ insertAll ((k, v) : xs) hmap = insertAll xs (insert hmap k v)
...
-    -- Note: in the True case, we must have p == psl
-    True -> (HashMap sz ct arr', IndexToUpdate (k, psl) ix)
-    False -> case psl < p of
-      True -> (HashMap sz ct arr', IndexToSwap (k, p) ix)
-      False -> probeFrom (k, p+1) ((ix+1)`mod` s) (HashMap sz ct arr')
-
+  (k, PSL) -> Int -> HashMap k v #-> (HashMap k v, ProbeResult k v)
+probeFrom (k, p) ix (HashMap ct arr) = Array.read arr ix & \case
+  (arr', Ur Nothing) ->
+    (HashMap ct arr', IndexToInsert p ix)
```
Divesh-Otwani, line 222: Same as below. Even if obvious, we should document that the `p` is the PSL of the query performed of the empty cell.

src/Data/HashMap/Linear.hs
```diff
@@ -99,112 +100,111 @@ newtype PSL = PSL Int
...
-    HashMap sz (Count (c + 1)) (write arr ix' (psl', (k',v)))
-  (HashMap (Size s) ct arr, IndexToSwap (k',psl') ix') -> read arr ix' & \case
-    (arr'', Ur (p_old, (k_old, v_old))) ->
-      tryInsertAtIndex
-        (HashMap (Size s) ct (write arr'' ix' (psl', (k',v))))
-        ((ix' + 1) `mod` s)
-        (k_old, p_old + 1)
-        v_old
+  HashMap k v #-> Int -> RobinVal k v -> HashMap k v
+tryInsertAtIndex hmap ix (RobinVal psl key val) =
```
Divesh-Otwani, line 144: Very nice.

src/Data/HashMap/Linear.hs
```diff
@@ -99,112 +100,111 @@ newtype PSL = PSL Int
...
-    case psl <= 0 of
-      True -> arr'
-      False -> write arr' ix (-1, (error "key", error "val")) & \case
-        arr'' ->
-          shiftSegmentBackward
-            (Size s)
-            (write arr'' ((ix+s-1) `mod` s) (psl + (-1),kv))
-            ((ix+1) `mod` s)
+  Int -> RobinArr k v #-> Int -> RobinArr k v
+shiftSegmentBackward s arr ix = Array.read arr ix & \case
```
Divesh-Otwani, line 199: Good.

src/Data/HashMap/Linear.hs
```diff
@@ -99,112 +100,111 @@ newtype PSL = PSL Int
...
 delete :: Keyed k => HashMap k v #-> k -> HashMap k v
-delete (HashMap sz@(Size s) ct@(Count c) arr) k =
-  probeFrom (k,0) ((hash k) `mod` s) (HashMap sz ct arr) & \case
-    (h, IndexToInsert _ _) -> h
-    (h, IndexToSwap _ _) -> h
-    (HashMap sz _ arr', IndexToUpdate _ ix) ->
-      write arr' ix (-1, (error "key", error "val")) & \case
-        arr'' -> shiftSegmentBackward sz arr'' ((ix + 1) `mod` s) & \case
-          arr''' -> HashMap sz (Count (c - 1)) arr'''
+delete hm k =
```
Divesh-Otwani, line 184: Good.

src/Data/HashMap/Linear.hs
```diff
@@ -99,112 +100,111 @@ newtype PSL = PSL Int
...
-resizeMap (HashMap (Size s) _ arr) = extractPairs arr & \case
-  (arr', Ur kvs) ->
-    allocBeside (s*2 + s`div`2) (-1, errKV) arr' & \case
-      (oldArr, arr'') ->
-        oldArr `lseq`
-          insertAll kvs (HashMap (Size (s*2 + s`div`2)) (Count 0) arr'')
+resizeMap hm =
+  capacity hm & \(HashMap _ arr, Ur cap) ->
+    extractPairs cap arr & \(arr', Ur kvs) ->
+      let newCap = cap*2 + cap`div`2
```
Divesh-Otwani, line 162: Add parentheses for those of us who are paranoid: `(cap*2) + (div cap 2)`.

src/Data/HashMap/Linear.hs
```diff
@@ -99,112 +100,111 @@ newtype PSL = PSL Int
...
 
 -- | If the key is present, this update the value.
 -- If not, if there's enough space, insert a new pair.
 -- If there isn't enough space, resize and insert
 insert :: (Keyed k, HasCallStack) => HashMap k v #-> k -> v -> HashMap k v
-insert (HashMap sz@(Size s) ct@(Count c) arr) k v = case c < s of
-  False -> resizeMap (HashMap sz ct arr) & \case
-    !hmap' -> insert hmap' k v
-  True -> tryInsertAtIndex (HashMap sz ct arr) ((hash k) `mod` s) (k,0) v
+insert hm k v =
```
Divesh-Otwani, line 129: Good.

Divesh-Otwani (approved):
Objectively better. Please add that documentation and then merge :smile:

src/Data/HashMap/Linear.hs
utdemir: I think it does, because it says:

> -- | A matching cell at index with a PSL and a value to update.

And also, I think; because they key is the same there (meaning the hash is the same), both PSL's needs to be the same at that point.

src/Data/HashMap/Linear.hs
utdemir: So, the `psl` on the constructurs are always the `psl` of the query. In `IndexToUpdate` case, they also happen to be the same as `psl` of the cell too.

But yeah, having different `psl`'s confusing. I'll document them.

PR #178, opened 2020-09-09
Renamed Unrestricted to Ur

### LINKED ISSUE #131: The name of Unrestricted

Divesh-Otwani:
This PR renames `Unrestricted` to `Ur` and closes #131. 

The discussion of how we arrived at this name is documented in the issue. Basically, `Unrestricted` is a really good name but too long. Since this data type is so fundamental to linear programming that any user will quickly become familiar with its definition, it's sensible to keep the name short for line length.

Remarks:
- In the spirit of keeping things short, I've changed qualified imports that were `Unrestricted` to `Ur`. Is this problematic or can we allow it since it will be so common? (I'm leaning towards the latter.)
- `unUnrestricted` has been renamed to `unUr`

### DIFF
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -71,7 +71,7 @@ inlined1 x = x
 useSubfunction :: Array a #-> Array a
 useSubfunction arr = fromRead (read arr 0)
   where
-    fromRead :: (Array a, Unrestricted a) #-> Array a
+    fromRead :: (Array a, Ur a) #-> Array a
     fromRead = undefined
 ```
 
@@ -89,16 +89,16 @@ Moveable`.
 
 ## Design patterns
 
-### `f :: X -> (SomeType #-> Unrestricted b) -> b` functions
+### `f :: X -> (SomeType #-> Ur b) -> b` functions
 
 This function limits the **scope** of using `SomeType` by taking
-a scope function of type `(SomeType #-> Unrestricted b)`
+a scope function of type `(SomeType #-> Ur b)`
 as its second argument and using it with a value of type `SomeType` to
-produce an `Unrestricted b`.
+produce an `Ur b`.
 
 The `SomeType` cannot escape by having it be inside the type `b` 
 in some way. This is because the `SomeType` is bound linearly in the scope
-function and `Unrestricted` holds its value with a non-linear constructor
+function and `Ur` holds its value with a non-linear constructor
 arrow.
 
 Now, if `f` is the only function that can make a `SomeType`,
--- a/src/Data/Unrestricted/Linear.hs
+++ b/src/Data/Unrestricted/Linear.hs
@@ -19,11 +19,12 @@
 -- be used in a linear way. Its use is __restricted__ while
 -- an argument in a non-linear function is __unrestricted__.
 --
--- Hence, a linear function with an argument of @Unrestricted a@ can use the
--- @a@ in an unrestricted way. That is, we have the following equivalence:
+-- Hence, a linear function with an argument of @Ur a@ (@Ur@ is short for
+-- /unrestricted/) can use the @a@ in an unrestricted way. That is, we have
+-- the following equivalence:
 --
 -- @
--- (Unrestricted a #-> b) ≌ (a -> b)
+-- (Ur a #-> b) ≌ (a -> b)
 -- @
 --
 -- = Consumable, Dupable, Moveable classes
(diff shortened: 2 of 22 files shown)


src/Data/Unrestricted/Linear.hs
```diff
@@ -95,35 +96,35 @@ import qualified Unsafe.Linear as Unsafe
...
--- > restricted x = f (unUnrestricted x)
+-- > restricted :: Ur a #-> b
+-- > restricted x = f (unUr x)
 -- >   where
 -- >     -- f __must__ be linear
 -- >     f :: a #-> b
 -- >     f x = ...
-unUnrestricted :: Unrestricted a #-> a
-unUnrestricted (Unrestricted a) = a
+unUr :: Ur a #-> a
```
mboes, line 118: `Ur` is itself a contraction. So I'd even go for `unur` here. Simple easy-to-spell and type combinator.

utdemir (approved):
LGTM (after addressing Mathieu's comment).

aspiwack:
> * `unUnrestricted` has been renamed to `unUr`

Sounds like the right choice. `unur` as @mboes is just fine too.

> * In the spirit of keeping things short, I've changed qualified imports that were `Unrestricted` to `Ur`. Is this problematic or can we allow it since it will be so common? (I'm leaning towards the latter.)

I'm less inclined to this.But let's try it and see how it pans out in practice.

docs/USER_GUIDE.md
```diff
@@ -89,16 +89,16 @@ Moveable`.
 
 ## Design patterns
 
-### `f :: X -> (SomeType #-> Unrestricted b) -> b` functions
+### `f :: X -> (SomeType #-> Ur b) -> b` functions
```
aspiwack, line 92: Ah, there is a bit of a mistake here. The proper pattern is `f :: X -> (SomeType #-> Ur b) #-> Ur b` (fix it after you merge, it's unrelated to this PR).

aspiwack (approved):

examples/Simple/Pure.hs
```diff
@@ -300,5 +300,5 @@ promote f (ForcedUnlinear x) = f x
    (ForcedUnlinear a  #-> b) ≅ (a -> b)
 
    In the Linear Haskell POPL '18 paper, this datatype is called
-   Unrestricted.
+   Ur.
```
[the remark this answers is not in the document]
Divesh-Otwani: Thx :smile:

src/Data/Unrestricted/Linear.hs
```diff
@@ -19,11 +19,12 @@
 -- be used in a linear way. Its use is __restricted__ while
 -- an argument in a non-linear function is __unrestricted__.
 --
--- Hence, a linear function with an argument of @Unrestricted a@ can use the
--- @a@ in an unrestricted way. That is, we have the following equivalence:
+-- Hence, a linear function with an argument of @Ur a@ (@Ur a@ abbreivates
+-- @Unrestricted a@) can use the @a@ in an unrestricted way. That is, we have
```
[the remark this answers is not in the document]
Divesh-Otwani: Yes, agreed.

src/Data/Unrestricted/Linear.hs
Divesh-Otwani: Cool :sunglasses:

Issue #165, opened 2020-09-05
Expand the linear mutable vector interface


Divesh-Otwani:
Linear mutable vectors currently only provide simple reads and writes and a `cons` function. We should have some more primitives available so that at least other use cases can be implemented externally when necessary.

We can use [`Data.Vector` module from `vector` package](https://hackage.haskell.org/package/vector-0.12.1.2/docs/Data-Vector.html) as a reference. At minimum:

* Versions of `read` and `write` without bounds checking for performance reasons.
* Ability to remove elements (and the vector should shrink accordingly).
* `fromList` and `toList` functions.
* Ability to get a slice of a vector.
* `map` and `filter` functions.
* Instances like `Functor`, `Foldable`, `Consumable`, `Dupable` where appropriate.

utdemir:
For the `Dupable` instance, I'll await the result of https://github.com/tweag/linear-base/issues/151.

Issue #151, opened 2020-09-02
Simpler 'Dupable' typeclass


utdemir:
Currently `Dupable` typeclass requires implementing the `dupV` function:

```haskell
class Consumable a => Dupable a where
  dupV :: KnownNat n => a #-> V n a
```

However, to me, it's not obvious how that can be implemented; and after spending hours on it today I couldn't figure it out. However

```haskell
dup2 :: Dupable a => a #-> (a, a)
```

function is simpler to implement, and I think we should define `Dupable` in terms of this. `Dupable`'s laws are already defined in terms of `dup2` instead of `dupV`, which also points to `dup2` being the simpler option.

I had this issue when trying to implement a `Dupable` instance for `Data.Array.Mutable.Linear`. The existing `Dupable` implementations all have simple implementations implemented using `V`'s applicative instance (or using other `Dupable` instances), but we can not use that if we can not share the value. In `Array`'s case, it's easy to implement a function like `clone :: Array a #-> (Array a, Array a)`, but going from that to `dupV` was not easy (for me).

We should still keep `dupV` in the class and define `{-# MINIMAL dup2 | dupV #-}`, in case when `dupV` is more efficient.

aspiwack:
There are a lot of design questions here, to be honest.

First, the vector-of-known-length library, while very useful in principle with linear types is… not great. It won't be super efficient, and the API is not the best (some stuff is missing, that's for sure). It's fine to include `dup2` into the typeclass for a quick-and-dirty definition.

By the way, one of the reason for moving away from `dup2` is that `V n` is a (data) applicative functor, and that really helps defining instances (of course, we could use `V 2` everywhere, instead, that would work too, but, in a sense, you may want more than 2, and contrary to monoids, you can't really rely on lists to represent your output, and generalise from `dup2` nicely, hence the `V n`).

[...]

PR #180, opened 2020-09-10
 Implement toList and toListLazy for mutable arrays and vectors 

utdemir:
Depends on #179. As part of #165,

It turns out, there is two different flavours of `toList` functions.

* 'toList' returns a new list alongside with the array. In order to do that, it has to read the entire array before returning; this can not done lazily since the underlying mutable array might change.
* 'toListLazy' converts an array to a lazy list, consuming the array in the process. It can do it lazily, since it does not give the array back, therefore guaranteeing in won't be changed.

Vector's 'toList' and 'toListLazy' functions also have the same constraint above.

In order to implement the vector changes, it also adds `unsafeRead` and `unsafeWrite` functions to the vector interface.

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -59,7 +59,6 @@ where
 
 import Data.Unrestricted.Linear
 import GHC.Stack
-import qualified Data.Functor.Linear as Control
 import Data.Array.Mutable.Unlifted.Linear (Array#)
 import qualified Data.Array.Mutable.Unlifted.Linear as Unlifted
 import Prelude.Linear ((&))
@@ -174,16 +173,9 @@ resize newSize seed (Array arr :: Array a)
       wrap (# old, new #) = old `Unlifted.lseq` Array new
 
 
--- XXX: Replace with toVec
-toList :: Array a #-> (Array a, Ur [a])
-toList arr = size arr & \case
-  (arr', Ur len) -> toListWalk (len - 1) arr' (Ur [])
-  where
-  toListWalk :: Int -> Array a #-> Ur [a] -> (Array a, Ur [a])
-  toListWalk ix arr accum
-    | ix < 0 = (arr, accum)
-    | otherwise = read arr ix & \case
-        (arr', Ur x) -> toListWalk (ix - 1) arr' ((x:) Control.<$> accum)
+-- | Return the array elements as a lazy list.
+toList :: Array a #-> Ur [a]
+toList (Array arr) = Unlifted.toList arr
 
 -- # Instances
 -------------------------------------------------------------------------------
--- a/src/Data/Array/Mutable/Unlifted/Linear.hs
+++ b/src/Data/Array/Mutable/Unlifted/Linear.hs
@@ -25,6 +25,7 @@ module Data.Array.Mutable.Unlifted.Linear
   , read
   , write
   , copyInto
+  , toList
   ) where
 
 import Data.Unrestricted.Linear hiding (lseq)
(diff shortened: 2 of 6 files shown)


src/Data/Array/Mutable/Linear.hs
```diff
@@ -111,8 +112,8 @@ fromList list (f :: Array a #-> Unrestricted b) =
 -- # Mutations and Reads
 -------------------------------------------------------------------------------
 
-size :: Array a #-> (Array a, Int)
-size (Array size' arr) =  (Array size' arr, size')
+size :: Array a #-> (Array a, Unrestricted Int)
+size (Array size' arr) =  (Array size' arr, Unrestricted size')
```
aspiwack, line 116: Careful to rebase if the `Unrestricted -> Ur` change is merged before (and vice versa).

src/Data/Array/Mutable/Linear.hs
```diff
@@ -167,18 +168,40 @@ resize newSize seed (Array _ mut)
...
+-- We implement 'toListLazy' below relying on two things:
+--
+-- * We hold the only reference to the underlying 'MutArr#'.
+-- Otherwise someone else could write to it before the returned
+-- list is evaluated.
+-- * 'Consume' instance of 'Array' is no-op.
+
+-- | Return the array elements as a lazy list. The array
+-- lookups will happen lazily, so this function also consumes
+-- the array.
```
aspiwack, line 194: You probably want to explain this a little better.

Also, maybe this ought to be the default, and the other one the one with a long name. I strongly suspect that this is a more common case for `toList`: you do `toList` to have a lazy traversal of your array, and probably you are abandoning the array there. (the other one is “just” the composition of `dup2` and `toList` (which, by the way, is easier to express with the `_2` lens, maybe that's how it ought to be defined (when the `Dupable` instance is implemented)))

utdemir: > You probably want to explain this a little better.

Agreed, will do.

> ... the other one is “just” the composition of dup2 and toList ....

This is a great point, thank you; I completely agree. So, now my plan is to drop the strict function, make the lazy function the default. Once we have the dupable instance (awaiting #151), it'd be trivial to implement that.

I will do the same to the `Vector` interface too.

utdemir:
I rebased this PR after `Ur` rename and made the default (shorter) `toList` method lazy as @aspiwack suggested. I did not add the old method as `toListStrict`, because I think it will be very trivial to implement in terms of `dup` and `toList`.

Divesh-Otwani (approved):
Looks good.

utdemir:
After the merging of #187, I had to move the `toList` function to unlifted mutable arrays. Since obeying the linear API forced me to strictly consume the array:

```haskell
toList :: Array a #-> Ur [a]
toList (arr :: Array a) =
  size arr & \(arr', Ur s) ->
    go 0 s arr'
 where
  go :: Int -> Int -> Array a #-> Ur [a]
  go i len arr'
    | i == len = arr' `lseq` Ur []
    | otherwise =
        unsafeRead arr' i & \(arr'', Ur a) ->
          go (i+1) len arr'' & \(Ur xs) ->
            Ur (a : xs)
```

Here, _I think_, pattern matching on the recursive call before the `:` constructor would make us traverse the entire array before returning the first element.

However, I will merge in this change, because it is holding up a few others. Please, still do comment here if you think that approach is wrong, and I will address it.

aspiwack:
> Here, _I think_, pattern matching on the recursive call before the `:` constructor would make us traverse the entire array before returning the first element.

It does sound accurate, indeed. I think that what we are seeing here is that in order to force an `Ur …`, we do have to consume every linear thingy it its closure. So it stands to reason that by the time we access the list, the array is consumed. Now, the only way we have of consuming an mutable array, currently, is to make it unavailable. So we kind of have to construct the entire list strictly.

Very handwavy, but I think that it's the idea.

But we can think this a bit further. There are really two `toList` that we may want to consider:
- A _linear_ `toList :: Array a #-> [Ur a]`. This one is lazy.
- The `toList` on immutable arrays: we can augment the arrays with an `O(1)` freezing function (like in the article), `Array a #-> Vector a` (or whatever immutable arrays are cf #185 ). Then `toList :: Array a -> Ur [a]; toList = fmap toList . freeze`.

Now, since the second one doesn't actually require a copy, it does subsume the first one. So probably we only need to do that one.

utdemir:
Just to clarify, we currently have a lazy `toList :: Array a #-> Ur [a]` function; it's just implemented with `MutableArray#` primitives on unlifted arrays, and lifted arrays refer to that one. Howeve I think it should be safe since it consumes the array as you mentioned.

However, I do agree that a `O(1)` freeze function is what we need here. I _think_ we should be able to go from our arrays to a `Vector`, that'd definitely make a lot of things easier. I'll create an issue for it.

PR #181, opened 2020-09-10
Add ability to shrink/slice vectors and arrays

utdemir:
Depends on #180,  As part of #165.

Previously, vectors could only grow. This PR adds functions around resizing and slicing vectors and arrays.

* Adds a `pop` function, and renames `snoc` to `push`, just because it sounds more natural to me than `unsnoc`. Happy to revert this change if you prefer the former.
* Adds `slice` functions to arrays and vectors that can return a sub-range of elements.
* Adds a `shrinkToFit` function to vectors to reduce the capacity of a vector.

Alongside with tests for each, of course.

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -179,7 +179,7 @@ toList :: Array a #-> Ur [a]
 toList (Array arr) = Unlifted.toList arr
 
 -- | Copy a slice of the array, starting from given offset and copying given
--- number of elements.
+-- number of elements. Returns the pair (oldArray, slice).
 --
 -- Start offset + target size should be within the input array, and both should
 -- be non-negative.
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -53,6 +53,7 @@ module Data.Array.Mutable.Linear
     read,
     unsafeRead,
     size,
+    slice,
     toList,
   )
 where
@@ -167,7 +168,7 @@ resize newSize seed (Array arr :: Array a)
       doCopy (Unlifted.allocBeside newSize seed arr)
      where
       doCopy :: (# Array# a, Array# a #) #-> Array a
-      doCopy (# src, dest #) = wrap (Unlifted.copyInto src dest)
+      doCopy (# src, dest #) = wrap (Unlifted.copyInto 0 src dest)
 
       wrap :: (# Array# a, Array# a #) #-> Array a
       wrap (# old, new #) = old `Unlifted.lseq` Array new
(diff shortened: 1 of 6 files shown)


src/Data/Array/Mutable/Linear.hs
```diff
@@ -167,18 +169,67 @@ resize newSize seed (Array _ mut)
   | otherwise =
       Array newSize (Unsafe.resizeMutArr mut seed newSize)
 
--- XXX: Replace with toVec
+-- | Copy a slice of the array, starting from given offset and copying given
```
Divesh-Otwani, line 172: Document which of the returned arrays is the slice.

src/Data/HashMap/Linear.hs
```diff
@@ -231,15 +231,15 @@ probeFrom (k, p) ix (HashMap sz@(Size s) ct arr) = read arr ix & \case
 -- # Querying
 --------------------------------------------------
 
-size :: HashMap k v #-> (HashMap k v, Int)
-size (HashMap sz ct@(Count c) arr) = (HashMap sz ct arr, c)
+size :: HashMap k v #-> (HashMap k v, Unrestricted Int)
```
Divesh-Otwani, line 234: Technically should be in a different PR but don't bother right now. I think it's cumbersome so I'll approve this one but going forward that's a good idea and cleaner.

src/Data/Set/Mutable/Linear.hs
```diff
@@ -61,17 +61,13 @@ delete (Set hmap) a = Set (Linear.delete hmap a)
 -- # Accessors
 -------------------------------------------------------------------------------
 
-size :: Keyed a => Set a #-> (Set a, Int)
-size (Set hmap) = fromHashMapSize (Linear.size hmap)
-  where
-    fromHashMapSize :: (Linear.HashMap a (), Int) #-> (Set a, Int)
-    fromHashMapSize (hmap, size) = (Set hmap, size)
+size :: Keyed a => Set a #-> (Set a, Unrestricted Int)
```
Divesh-Otwani, line 64: Same as comment on hashmaps.

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -103,26 +122,49 @@ constant size' x f
...
-snoc (Vec size' arr) x =
-  Array.size arr & \(arr', cap') ->
-    move cap' & \(Unrestricted cap) ->
-      if size' < cap
-      then write (Vec (size' + 1) arr') size' x
-      else write (unsafeResize ((max size' 1) * 2) (Vec (size' + 1) arr')) size' x
-
--- | Write to an element already written to before.  Note: this will not write
--- to elements beyond the current size of the array and will error instead.
+-- | Number of elements inside the vector.
```
Divesh-Otwani, line 125: Mention that this is different from the capacity.

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -141,6 +241,29 @@ instance Consumable (Vector a) where
...
+growToFit n vec =
+  capacity vec & \(vec', Unrestricted cap) ->
+    size vec' & \(vec'', Unrestricted s') ->
+      if s' + n <= cap
+      then vec''
+      else
+        let -- Calculate the closest power of growth factor
+            -- larger than required size.
+            newSize =
+              constGrowthFactor
```
Divesh-Otwani, line 256: Add a comment mentioning this is a constant defined in this file above.

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -132,6 +174,64 @@ read (Vec size' arr) ix
         & \(arr', val) -> (Vec size' arr', val)
   | otherwise = arr `lseq` error "Read index not in range."
 
+-- | Same as 'read', but does not do bounds-checking. The behaviour is undefined
+-- when passed an invalid index.
+unsafeRead :: HasCallStack => Vector a #-> Int -> (Vector a, Unrestricted a)
```
Divesh-Otwani, line 179: For consistency, could you change `read` and `write` to use the same elegant design as the arrays that call `unsafeRead` and `unsafeWrite` with `assertIndexInRange ix arr` that panics if the index is not in range?

Divesh-Otwani (approved):

src/Data/Vector/Mutable/Linear.hs
utdemir: It feels a bit odd to add a comment like `-- this is defined at the top of this file`. I would say that it is the IDE's responsibility (or `grep`'s). Anyway, a comment wont harm anyone, so I'm happy to put the it there (unless you suggest a better one :)).

PR #182, opened 2020-09-10
Implement Functor and Semigroup instances for mutable vectors and arrays

utdemir:
Depends on #181, As part of #165.

This PR adds a (data) `Functor` instance for mutable arrays and vectors. The `fmap` implementation for both instances do in-place updates (yay linear types). Internal implementation does `unsafeCoerce`, but hopefully it is simple enough, and there's test coverage.

It also adds a `Semigroup` instance for `Vector`'s, where `mappend` tries not to allocate a new vector when there is enough space available.

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -63,7 +63,6 @@ import GHC.Stack
 import Data.Array.Mutable.Unlifted.Linear (Array#)
 import qualified Data.Array.Mutable.Unlifted.Linear as Unlifted
 import qualified Data.Functor.Linear as Data
-import qualified Unsafe.Linear as Unsafe
 import Prelude.Linear ((&))
 import Prelude hiding (read)
 
@@ -224,21 +223,7 @@ instance Consumable (Array a) where
   consume (Array arr) = arr `Unlifted.lseq` ()
 
 instance Data.Functor Array where
-  fmap (f :: a #-> b) arr =
-    size arr & \(arr', Ur s) -> go 0 s arr'
-   where
-    -- When we're mapping  an array, we first insert `b`'s
-    -- inside an `Array b` by unsafeCoerce'ing, and then we
-    -- unsafeCoerce the result to an `Array b`.
-    go :: Int -> Int -> Array a #-> Array b
-    go i s arr
-      | i == s =
-          Unsafe.coerce arr
-      | otherwise =
-          unsafeRead arr i
-            & \(arr', Ur a) ->
-                unsafeWrite arr' i (Unsafe.coerce (f a))
-            & \arr'' -> go (i+1) s arr''
+  fmap f (Array arr) = Array (Unlifted.map f arr)
 
 -- # Internal library
 -------------------------------------------------------------------------------
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -62,6 +62,8 @@ import Data.Unrestricted.Linear
 import GHC.Stack
 import Data.Array.Mutable.Unlifted.Linear (Array#)
 import qualified Data.Array.Mutable.Unlifted.Linear as Unlifted
+import qualified Data.Functor.Linear as Data
+import qualified Unsafe.Linear as Unsafe
 import Prelude.Linear ((&))
 import Prelude hiding (read)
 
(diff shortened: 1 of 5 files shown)


aspiwack:
> where `mappend` tries not to allocate a new vector when there is enough space available.

This is a neat trick.

utdemir:
Sorry, I also added `filter` and `mapMaybe` functions to this PR (alongside with their tests), and implemented `fmap` in terms of `mapMaybe`.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -221,6 +223,23 @@ instance Consumable (Array a) where
...
+    -- unsafeCoerce the result to an `Array b`.
+    go :: Int -> Int -> Array a #-> Array b
+    go i s arr
+      | i == s =
+          Unsafe.coerce arr
+      | otherwise =
+          unsafeRead arr i
+            & \(arr', Ur a) ->
+                unsafeWrite arr' i (Unsafe.coerce (f a))
+            & \arr'' -> go (i+1) s arr''
```
Divesh-Otwani, line 241: Nitpicks all of which you can ignore if you want to:
- We could `go` from `s` down to `0` which is a bit cleaner
- Eventually we might want to do this in the unlifted array and then wrap the `map` on unlifted arrays to lifted ones

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -179,14 +187,67 @@ unsafeRead (Vec size' arr) ix =
...
+-- use 'shrinkToFit' for that if necessary.
+filter :: Vector a #-> (a -> Bool) -> Vector a
+filter v f = mapMaybe v (\a -> if f a then Just a else Nothing)
+-- TODO A slightly more efficient version exists, where we skip the writes
+-- until the first time the predicate fails. However that requires duplicating
+-- most of the logic at `mapMaybe`, so lets not until we have benchmarks to
+-- see the advantage.
+
+-- | A version of 'fmap' which can throw out elements.
+mapMaybe :: Vector a #-> (a -> Maybe b) -> Vector b
```
Divesh-Otwani, line 227: Very nice implementation with `go`.

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -219,6 +280,28 @@ slice from newSize (Vec oldSize arr) =
...
+  -- This operation tries to use the existing capacity of v1 when possible.
+  v1 <> v2 =
+    size v2 & \(v2', Ur s2) ->
+      growToFit s2 v1 & \v1' ->
+        toList v2' & \(Ur xs) ->
+          go xs v1'
+   where
+     go :: [a] -> Vector a #-> Vector a
+     go [] vec = vec
+     go (x:xs) (Vec cap arr) =
```
Divesh-Otwani, line 298: `cap` is the size and the variable name should probably be something like `sz`.

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -219,6 +280,28 @@ slice from newSize (Vec oldSize arr) =
...
+       unsafeWrite (Vec (cap+1) arr) cap x
+         & go xs
+
+instance Data.Functor Vector where
+  fmap f vec = mapMaybe vec (\a -> Just (f a))
```
Divesh-Otwani, line 303: Nice.

test/Test/Data/Mutable/Vector.hs
```diff
@@ -51,6 +54,10 @@ group =
   , testProperty "toList can be implemented with repeated pops" refToListViaPop
   , testProperty "fromList can be implemented with repeated pushes" refFromListViaPush
   , testProperty "toList works with extra capacity" refToListWithExtraCapacity
+  , testProperty "fromList xs <> fromList ys = fromList (xs <> ys)" refMappend
+  , testProperty "mapMaybe f (fromList xs) = fromList (mapMaybe f xs)" refMapMaybe
+  , testProperty "filter f (fromList xs) = fromList (filter f xs)" refFilter
+  , testProperty "f <$> fromList xs == fromList (f <$> xs)" refFmap
```
Divesh-Otwani, line 60: Nice.

Divesh-Otwani (approved):
Very cool stuff.

src/Data/Array/Mutable/Linear.hs
utdemir: > We could go from s down to 0 which is a bit cleaner

That is confusing to me. I can easily understand it conciously, but it's just too ingrained to me that I should iterate the array from beginning to the end. If I move the other way around, I need to start from `len-1` and end at `-1` in this case, which is ugly to me. So if it's okay I'd prefer to keep it like it is :).

> Eventually we might want to do this in the unlifted array and then wrap the map on unlifted arrays to lifted ones

I did that, thanks! I couldn't even use `unsafeCoerce` because of the kind mismatch, so I had to use `unsafeCoerce#`, which was scary. I think since we're casting `MutableArray# a` to `MutableArray# b`, it should be fine.

PR #187, opened 2020-09-15
Replace Unsafe.MutableArray with Data.Array.Mutable.Unlifted.Linear

### LINKED ISSUE #177: Ensure effects happen in order when using `Unsafe.MutableArray`
Currently, linear mutable arrays and vectors use Unsafe.MutableArray, which requires forcing evaluation of seemingly pure expressions that have side effects. However, there is little control on the order of side-effects.

```Haskell
case writeMutArray ix val arr of
  () -> k
```
is equivalent to `seq (writeMutArray ix val arr) k` which the documentation of [Control.Parallel.pseq](https://www.stackage.org/haddock/lts-16.13/parallel-3.2.2.0/Control-Parallel.html#v:pseq) warns that might evaluate effects in a different order.

Similarly for reading arrays we do
```Haskell
case readMutArray ix arr of
  (# a #) -> f a
```
which for the same reasons as `seq`, it could be reordered to `f (case readMutArray ix arr of (# a #) -> a)` if GHC thinks that `f` is strict.

It turns out the only mechanism GHC offers to order effects is `GHC.Conc.pseq`, which can be used to fix the write case, and less clearly the read case because the levity of the boxed tuple prevents it from being given to `pseq`.

I think the best fix for linear base would be to make the interface of `Unsafe.MutableArray` harder to misuse. For instance,
```Haskell
writeMutArr :: HasCallStack => MutArr# a -> Int -> a -> b -> b
writeMutArr mutArr (I# ix) val k =
  case doWrite of _ -> GHC.Exts.lazy k
  where
    doWrite = runRW# $ \stateRW -> writeArray# mutArr ix val stateRW

-- | Read from a given mutable array arr, given an index
-- in [0, size(arr)-1]
--
-- This returns an unboxed tuple to give the callee the
-- ability to evaluate the function call without evaluating
-- the final result.
readMutArr :: HasCallStack => MutArr# a -> Int -> (a -> b) -> b
readMutArr mutArr (I# ix) k =
  case doRead of (# _, a #) -> GHC.Exts.lazy (k a)
  where
    doRead = runRW# $ \stateRW -> readArray# mutArr ix stateRW
```

The fix above relies on continuations and on the definition of [pseq](https://www.stackage.org/haddock/lts-16.13/base-4.13.0.0/src/GHC-Conc-Sync.html#pseq) being
```
pseq a b = a `seq` lazy b
```

### LINKED ISSUE #184: HashMap test failure caused by innocent code change
I found an issue at the mutable HashMap implementation on current master (`191bade`) alongside with a way to reliably reproduce it, however I couldn't pinpoint the exact problem. This issue is just to track my findings.

Here's the reproducer:

```haskell
--- a/src/Data/HashMap/Linear.hs
+++ b/src/Data/HashMap/Linear.hs
@@ -118,13 +118,13 @@ empty :: forall k v b.
   Keyed k => Size -> (HashMap k v #-> Ur b) #-> Ur b
 empty sz@(Size s) scope =
   alloc
-    s
+    16
     ( -1
     , ( error "reading error hashmap key"
       , error "reading error hashmap val"
       )
     )
-    (\arr -> scope (HashMap sz (Count 0) arr))
+    (\arr -> scope (HashMap (Size 16) (Count 0) arr))
 
 -- | If the key is present, this update the value.
 -- If not, if there's enough space, insert a new pair.
```

With the above patch, various HashMap tests fail with mismatching expected and actual values. Here is where it gets interesting:

* That value should not matter, since it's just the initial capacity of the `HashMap`. It should grow automatically.
* As far as I can see, we're already passing `16` as the size parameter from the tests. So, something is breaking referential transparency.

Since I do not want to believe the second point above, it's likely that I'm missing something crucial. @Divesh-Otwani do you have an idea?

utdemir:
Depends on #179. Closes #184, and closes #177 .

This commit adds an internal linear unlifted mutable array type, and implements our lifted arrays in terms of that. Hopefully, this should result in a safer code, since it _should_ prevent laziness issues surrounding `Unsafe.MutableArray`. As @aspiwack predicted, this already resolves #184; and our tests regarding the read/write ordering on `Array`'s (#135) still pass.

It is pretty similar to our existing `Array` module, other than the fact that it is unlifted. However, since it is an unlifted type and levity polymorphism has some restrictions; we can not use our usual combinators like `&`. Hence, I had to use a lot of intermediate functions.

Also, because of the fact that `IO` can not contain unlifted types (and `unsafeDupablePerformIO` can not return them); I couldn't implement this in `IO` as @aspiwack suggested (I guess we could pass it around within a lifted type, but I choose not to). So it still uses `runRW#`. The way we use `runRW#` is exactly the same as `unsafeDupablePerformIO`, so I think there shouldn't be any differences. However, that requires us spending extra time reviewing especially the `read` and `write` functions on `Array#` type.


aspiwack:
> So it still uses `runRW#`. The way we use `runRW#` is exactly the same as `unsafeDupablePerformIO`, so I think there shouldn't be any differences.

I yes, I hadn't thought of this. Ah well, it's fine to use `runRW#`, the only limit is that we don't have a `(>>=)` operator, and must do sequencing manually with `case`. It's fine.

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,107 @@
...
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UnboxedTuples #-}
+{-# LANGUAGE UnliftedNewtypes #-}
+
+-- |
+-- This module defines a linear wrapper around 'GHC.MutableArray#'
+-- which can be manipulated in pure contexts.
```
aspiwack, line 11: We should work on this description. The module defines an unlifted array type, with a pure linear interface, which makes sure that many operations can be implemented by mutation, rather that copy. Something to this effect.

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,107 @@
...
+--
+-- It ensures that effects happen in order. However, access to
```
aspiwack, line 13: > It ensures that effects happen in order

Of course it does: officially, there is no effect. This is more a note to your past self than documentation.

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,107 @@
...
+  ) where
+
+import Data.Unrestricted.Linear hiding (lseq)
+import Prelude (Int)
+import qualified Prelude as Prelude
+import qualified Unsafe.Linear as Unsafe
+import qualified GHC.Exts as GHC
+
+-- | A mutable array holding @a@s
+newtype Array# a = Array# (GHC.MutableArray# GHC.RealWorld a)
```
aspiwack, line 36: I hadn't realise that the existence of this file at all was predicated on this new unlifted newtype feature. Quite convenient, isn't it?

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,107 @@
...
+
+-- | Run an unsafe computation using the underlying MutableArray,
+-- consuming the 'Array#' in process.
+unArray# :: (GHC.MutableArray# GHC.RealWorld a -> b) -> Array# a #-> Ur b
```
aspiwack, line 40: Is it unsafe? If so, `unsafe` should occur in the name of the function (and, maybe, we ought to consider not exporting the functino at all, if it's unsafe). If it's not unsafe, then don't say that it is.

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,107 @@
...
+  let !new = GHC.runRW# Prelude.$ \st ->
+        case GHC.newArray# s a st of
+          (# _, arr #) -> Array# arr
+   in f new
+
+-- | Allocate a mutable array of given size using a default value,
+-- using another 'Array#' as a uniqueness proof.
+--
+-- The size should be non-negative.
+allocBeside :: Int -> a -> Array# b #-> (# Array# b, Array# a #)
```
aspiwack, line 64: Ha. You went the extra mile, here. And used unboxed pairs. It's probably wise, fewer lazy things, means fewer opportunities to mix up the effect ordering.

Ah! But you don't really have a choice, really, since `(,)` can only contain lifted values. How convenient.

src/Data/Array/Mutable/Linear.hs
```diff
@@ -79,18 +77,25 @@ data Array a where
...
 -- using another array as a uniqueness proof.
 allocBeside :: Int -> a -> Array b #-> (Array b, Array a)
-allocBeside size val orig
-  | size >= 0 = (orig, Array size (Unsafe.newMutArr size val))
-  | otherwise = orig `lseq` error ("Trying to allocate an array of size " ++ show size)
+allocBeside s x (Array orig)
+  | s < 0 =
+     Unlifted.lseq
+       orig
+       (error ("Array.allocBeside: negative size: " ++ show s))
```
aspiwack, line 93: I'm curious: doesn't `error (…) orig` work here? Is `error`'s type not polymorphic enough in the type representation?

src/Data/Array/Mutable/Linear.hs
```diff
@@ -100,53 +105,46 @@ fromList list (f :: Array a #-> Ur b) =
...
 -- if an out-of-bounds index is provided.
 unsafeWrite :: Array a #-> Int -> a -> Array a
-unsafeWrite (Array size' arr) ix val =
-  case Unsafe.writeMutArr arr ix val of
-    () -> Array size' arr
+unsafeWrite arr ix val =
+  withUnlifted
+    (\arr -> (# Unlifted.write ix val arr, () #))
+    arr
+    & \(a, ()) -> a
```
aspiwack, line 137: This one is a bit weird, maybe optimise it later. (there is a change that GHC can optimise it on its own, but it's a tad hard to make sure)

src/Data/Array/Mutable/Linear.hs
```diff
@@ -182,16 +188,24 @@ toList arr = size arr & \case
...
-indexInRange size ix = 0 <= ix && ix < size
+-- | Run a computation with the underlying unlifted array.
+withUnlifted :: (Array# a #-> (# Array# a, b #)) #-> Array a #-> (Array a, b)
+withUnlifted f (Array arr) = unwrap (f arr)
+ where
+  unwrap :: (# Array# a, b #) #-> (Array a, b)
+  unwrap (# a, b #) = (Array a, b)
+
+-- | Check if given index is within the Array, otherwise panic.
+assertIndexInRange :: HasCallStack => Int -> Array a #-> Array a
```
aspiwack, line 206: Clever.

aspiwack (approved):

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: Reading it again, you're right. The current documentation just talks about the implementation details. So, what do you think about:

> This modules defines an unlifted array type with a linear interface, where the operations are implemented with in-place updates rather than copying whenever possible.

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: You are right, I will remove that sentence.

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: I don't think it is unsafe. It returns a `MutableArray#`, which is inherently unsafe, but specifically the interface of this function is not unsafe, eg. there is no way to use it incorrectly. I'll fix the documentation.

As a side note, I actually wanted to write `:: Array# a #-> Ur (MutableArray# RealWorld a)`, but that doesn't work since `Ur` can not carry unlifted types.

src/Data/Array/Mutable/Linear.hs
utdemir: Sure, I can inline the `withUnlifted` call here.

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: It was very convenient. I honestly don't yet know how I'd implement this module without it.

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,107 @@
...
+
+-- | Allocate a mutable array of given size using a default value.
+--
+-- The size should be non-negative.
+alloc :: Int -> a -> (Array# a #-> Ur b) #-> Ur b
+alloc (GHC.I# s) a f =
+  let !new = GHC.runRW# Prelude.$ \st ->
+        case GHC.newArray# s a st of
+          (# _, arr #) -> Array# arr
+   in f new
```
facundominguez, line 58: Why is necessary to force `new`? Doesn't the type of `f` force it to be strict?

Also, what prevents `new` from being floated and reused for two different invocations of `alloc`?

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,107 @@
...
+-- | Allocate a mutable array of given size using a default value,
+-- using another 'Array#' as a uniqueness proof.
+--
+-- The size should be non-negative.
+allocBeside :: Int -> a -> Array# b #-> (# Array# b, Array# a #)
+allocBeside (GHC.I# s) a orig =
+  let !new = GHC.runRW# Prelude.$ \st ->
+        case GHC.newArray# s a st of
+          (# _, arr #) -> Array# arr
+   in (# orig, new #)
```
facundominguez, line 69: Same question about forcing `new` here and about preventing it from floating.

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,107 @@
...
+      let !s = GHC.sizeofMutableArray# arr
+      in  (# Array# arr, Ur (GHC.I# s) #)
+
+read ::  Int -> Array# a #-> (# Array# a, Ur a #)
+read (GHC.I# i) = Unsafe.toLinear go
+  where
+    go :: Array# a -> (# Array# a, Ur a #)
+    go (Array# arr) =
+      case GHC.runRW# (GHC.readArray# arr i) of
+        (# _, ret #) -> (# Array# arr, Ur ret #)
```
facundominguez, line 85: I don't see how this prevents the `case` expression from being rewritten to
```Haskell
 (# Array# arr, case GHC.runRW# (GHC.readArray# arr i) of (# _, ret #) -> Ur ret #)
```

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,107 @@
...
+copyInto :: Array# a #-> Array# a #-> (# Array# a, Array# a #)
+copyInto = Unsafe.toLinear2 go
+  where
+    go :: Array# a -> Array# a -> (# Array# a, Array# a #)
+    go (Array# src) (Array# dst) =
+      let !(GHC.I# len#) = Prelude.min
+            (GHC.I# (GHC.sizeofMutableArray# src))
+            (GHC.I# (GHC.sizeofMutableArray# dst))
+      in  case GHC.runRW# (GHC.copyMutableArray# src 0# dst 0# len#) of
+            _ -> (# Array# src, Array# dst #)
```
facundominguez, line 107: Why is the bang pattern needed here? I'd expect `len#` to be forced by the `case` expression.

facundominguez (commented):
I still find the implementation far from obviously correct. But I'm curious at the reasons that make it work, nonetheless.

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,107 @@
...
+      case GHC.runRW# (GHC.readArray# arr i) of
+        (# _, ret #) -> (# Array# arr, Ur ret #)
+
+write :: Int -> a -> Array# a #-> Array# a
+write (GHC.I# i) (a :: a) = Unsafe.toLinear go
+  where
+    go :: Array# a -> Array# a
+    go (Array# arr) =
+      case GHC.runRW# (GHC.writeArray# arr i a) of
+        _ -> Array# arr
```
facundominguez, line 93: Does `Unsafe.toLinear` prevent this code from being inlined?
Otherwise, what stops
```Haskell
case GHC.runRW# (GHC.writeArray# arr i a) of
  _ -> case GHC.runRW# (GHC.writeArray# arr i b) of
    _ -> Array# arr
```
from being rewritten to
```Haskell
case GHC.runRW# (GHC.writeArray# arr i b) of
  _ -> case GHC.runRW# (GHC.writeArray# arr i a) of
    _ -> Array# arr
```

src/Data/Array/Mutable/Unlifted/Linear.hs
facundominguez: Ah, never mind. The unboxed tuple is not strict on the second component.

src/Data/Array/Mutable/Unlifted/Linear.hs
Divesh-Otwani: That's still too developer focused, I think. Maybe just:

> This module provides an unlifted mutable array with a pure interface. Though the array itself is unlifted, it's elements are lifted types. This is made possible by using linear types to make sure array references are single threaded through reads and writes.

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: I do not know; I don't think `Unsafe.toLinear` would prevent this code from being inlined, since it is nothing more than a `unsafeCoerce`. Do you think adding a bang to the wildcard pattern would fix it?

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: There's a GHC warning if I don't use the bang pattern there:

```
/home/utdemir/workspace/github.com/tweag/linear-base6/src/Data/Array/Mutable/Unlifted/Linear.hs:103:11: error: [-Wunbanged-strict-patterns, -Werror=unbanged-strict-patterns]
    Pattern bindings containing unlifted types should use
    an outermost bang pattern:
      (GHC.I# len#)
        = Prelude.min
            (GHC.I# (GHC.sizeofMutableArray# src))
            (GHC.I# (GHC.sizeofMutableArray# dst))
    |       
103 |       let (GHC.I# len#) = Prelude.min
    |           ^
```

Just had a quick search, and the clarification behing that warning is explained here: https://ghc-proposals.readthedocs.io/en/latest/proposals/0035-unbanged-strict-patterns.html

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: > Why is necessary to force new? Doesn't the type of f force it to be strict?

You are right, I can remove the bang, I think.

> Also, what prevents new from being floated and reused for two different invocations of alloc?

Wouldn't the fact that it uses bindings from outer scope (`a`, `s`)  prevents it being floated out?

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: Same answer :) , I can remove the forcing of `new`, and I was hoping the variables from the outer scope would prevent the floating.

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: @Divesh-Otwani I like it. Should we swap your second and third sentence, because I think the "this" on your third sentence refers to the first sentence rather than the second?

src/Data/Array/Mutable/Linear.hs
utdemir: I think `error`'s type is polymorphic enough; but our problem is with linearity. Since type inference doesn't infer linear arrows (bold statement, let me know if this is incorrect), we need to put an explicit type signature. However, in this specific case it's especially annoying since `orig` is unlifted, so we should also specify the kind on that signature.

In the end, if we enable `-XKindSignatures` and `-XDataKinds`, we can do something like:

 ```haskell
    (error ("Array.allocBeside: negative size: " ++ show s)
      :: (a :: TYPE UnliftedRep) #-> b
    ) orig
```

I can not think a shorter way to express this, so in this state it's not worth the hassle.

File: src/Data/Array/Mutable/Unlifted/Linear.hs (hunk 9 above)
Comment (facundominguez) [reply src/Data/Array/Mutable/Unlifted/Linear.hs:59]: > Wouldn't the fact that it uses bindings from outer scope (a, s) prevents it being floated out?
I don't think so.

Suppose cse transforms this `alloc s a (\arr -> alloc s a (f arr))` to
```Haskell
let myAlloc x = alloc s a x
 in myAlloc (\arr -> myAlloc (f arr))
```
Then inline alloc, and float the `new` binding.

`{-# NOINLINE alloc #-}` looks necessary here, plus threading the f parameter as well:
```Haskell
alloc (GHC.I# s) a f0 =
  let (!new, f) = GHC.runRW# Prelude.$ \st ->
        case GHC.newArray# s a st of
          (# _, arr #) -> (Array# arr, f0)
   in f new
```

facundominguez: Or NOINLINE + maybe:
```Haskell
alloc (GHC.I# s) a f =
  GHC.runRW# Prelude.$ \st ->
        case GHC.newArray# s a st of
          (# _, arr #) -> f (Array# arr)
```

src/Data/Array/Mutable/Unlifted/Linear.hs
facundominguez: I'm thinking `unsafeCoerce (\x -> f x) a` prevents GHC to simplify the expression to `f x` because it would be ill typed.
Even if `write` is inlined, I think this could somewhat prevent `go` from being affected by the surrounding code. But it is not a solid argument.

This discussion affects correctness of `read`, `write` and `copyInto`. One workaround would be to declare the three functions as non-inlineable.

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: Thank you @facundominguez , I did not know CSE could do that. I will add the `NOINLINE` pragma.

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: I will make those three functions as non-inlineable. I don't know how big of a performance hit it'll be, but correctness should come first.

utdemir:
@facundominguez I added the `NOINLINE` pragmas you suggested. Thank you for looking into these, there is no way I'd have figured them myself. 

> I still find the implementation far from obviously correct.

I agree to that, after reading your comments. I guess the point now would be to reduce the unsafe surface area to this module.

src/Data/Array/Mutable/Linear.hs
aspiwack: Fair enough.

src/Data/Array/Mutable/Unlifted/Linear.hs
aspiwack: I'm not entirely sure that you could have. At best it would have been quite unpleasant. Let us celebrate the coïncidence.

src/Data/Array/Mutable/Unlifted/Linear.hs
aspiwack: > I'm thinking `unsafeCoerce (\x -> f x) a` prevents GHC to simplify the expression to `f x` because it would be ill typed.

It doesn't, actually. There are special provisions in the compiler to let the computation go further (basically, this is changed to `(\x -> unsafeCoerce (f x)) (unsafeCoerce a)`. At present, `unsafe.toLinear` will, however, block this computation. But this is considered a limitation of the implementation, and will be fixed.

Generally, speaking, it is a good point that we ought to add `NOINLINE` on every function calling `runRW#`. Otherwise, issues like what @facundominguez pointed out may arise.

src/Data/Array/Mutable/Linear.hs
Divesh-Otwani: An easy trick is `error "something" $ linearVar1 linearVar2 ...` with the linear variant of `($)`. That works.

utdemir: Oh, that is smart. Thanks @Divesh-Otwani !

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,116 @@
...
+size :: Array# a #-> (# Array# a, Ur Int #)
+size = Unsafe.toLinear go
+  where
+    go :: Array# a -> (# Array# a, Ur Int #)
+    go (Array# arr) =
+      let !s = GHC.sizeofMutableArray# arr
+      in  (# Array# arr, Ur (GHC.I# s) #)
+
+read ::  Int -> Array# a #-> (# Array# a, Ur a #)
+read (GHC.I# i) = Unsafe.toLinear go
```
Divesh-Otwani, line 86: Okay, the rough argument for why this works needs to be documented in /docs/DESIGN.md. 

Here's my understanding of it:

For mutable data to work properly in Haskell, the constructing, mutating and accessing operations need to atomic. That does not mean that they are used strictly. We can have a lazy `(read arr ix)`. We just require that any evaluation of `(read arr ix)` fully does the read (even if the value read is itself lazy). For constructors and mutators that return a single argument, this is already done by necessity. Evaluating `((f :: Array Int #-> Ur Bool) newArray)` forces the array to be created before any writes or reads are done to it. Similarly, any computation that uses a `(write arr ix a) & \arr' -> (...)` must do the write before reading from the written array.

The issue arises when we have API functions that return multiple things like `read`. Here, when we evaluate `(read arr ix)` and get out the read value and the array, we must never have a thunk of the read value that has yet to perform an mutable read of the underlying array. The same goes for a copying function like `slice`. We should never be able to use the first part of `(slice arr ix n)` (that is, the original array) without having the slice be done to the mutable array underneath.

Here, we can reason as follows. We want to show that pattern matching on `read arr ix` forces the underlying read to occur even if we just discard it. First, note that when `(read arr ix)` is evaluated, it is done without inlining. Thus, we only need to argue that local optimizations within this functions won't mess things up. 

Now, when matching on unlifted tuples, we can be sure what's held inside is also unlifted. That is, the values cannot be bottom. If `Ur ret` is not bottom, then (I think) `ret` is not bottom. For this to be the case, GHC must evaluate the `GHC.runRW#  (GHC.readArray# arr i)`. Even if GHC re-writes this to `(# Array# arr, Ur (GHC.runRW# ...))` it must not leave the thunk lest it be bottom. Hence, the underlying read occurs on the pattern match and `read` is atomic.

@facundominguez @aspiwack Is this argument clear and correct?



A mutator or constructor like `write` or `newArray` will never cause issues because to do anything with their result, they must be evaluated first.

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -0,0 +1,116 @@
...
+{-# NOINLINE allocBeside #-}
+
+size :: Array# a #-> (# Array# a, Ur Int #)
+size = Unsafe.toLinear go
+  where
+    go :: Array# a -> (# Array# a, Ur Int #)
+    go (Array# arr) =
+      let !s = GHC.sizeofMutableArray# arr
+      in  (# Array# arr, Ur (GHC.I# s) #)
+
```
Divesh-Otwani, line 84: Link to developer documentation (see review comment below) and link at the top of the file.

Divesh-Otwani (approved):
I think this is correct which makes me very happy. The idea of an unlifted array was actually brilliant since it forces us to use unlifted tuples and isolates all issues to really small places. (Epic move @aspiwack.) If you don't add developer docs before merging, please make an issue for it and link to this PR.

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -57,6 +57,10 @@ alloc (GHC.I# s) a f =
         case GHC.newArray# s a st of
           (# _, arr #) -> Array# arr
    in f new
+{-# NOINLINE alloc #-}
```
facundominguez, line 60:
```suggestion
{-# NOINLINE alloc #-}  -- prevents runRW# from floating outwards
```

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -68,6 +72,7 @@ allocBeside (GHC.I# s) a orig =
         case GHC.newArray# s a st of
           (# _, arr #) -> Array# arr
    in (# orig, new #)
+{-# NOINLINE allocBeside #-}
```
facundominguez, line 75:
```suggestion
{-# NOINLINE allocBeside #-}  -- prevents runRW# from floating outwards
```

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -84,6 +89,7 @@ read (GHC.I# i) = Unsafe.toLinear go
     go (Array# arr) =
       case GHC.runRW# (GHC.readArray# arr i) of
         (# _, ret #) -> (# Array# arr, Ur ret #)
+{-# NOINLINE read #-}
```
facundominguez, line 92:
```suggestion
{-# NOINLINE read #-}  -- prevents the runRW# effect from being reordered
```

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -92,6 +98,7 @@ write (GHC.I# i) (a :: a) = Unsafe.toLinear go
     go (Array# arr) =
       case GHC.runRW# (GHC.writeArray# arr i a) of
         _ -> Array# arr
+{-# NOINLINE write #-}
```
facundominguez, line 101:
```suggestion
{-# NOINLINE write #-}  -- prevents the runRW# effect from being reordered
```

src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -106,3 +113,4 @@ copyInto = Unsafe.toLinear2 go
             (GHC.I# (GHC.sizeofMutableArray# dst))
       in  case GHC.runRW# (GHC.copyMutableArray# src 0# dst 0# len#) of
             _ -> (# Array# src, Array# dst #)
+{-# NOINLINE copyInto #-}
```
facundominguez, line 116:
```suggestion
{-# NOINLINE copyInto #-}  -- prevents the runRW# effect from being reordered
```

facundominguez (commented):
I suggested some comments to tag the NOINLINE pragmas with their purpose.

src/Data/Array/Mutable/Unlifted/Linear.hs
facundominguez: > mutating and accessing operations need to be atomic.

I'm guessing atomic here means that the compiler can't inline them and mix their subexpressions? I need to ask because I've only seen atomic used to describe transactions and concurrent operations.

> Now, when matching on unlifted tuples, we can be sure what's held inside is also unlifted. 

This is true when the type of the component is unlifted.
```Haskell
case (# Array# arr, undefined #) of
  (# arr, d #) -> ()  -- this should force the first component of the pair only.
```

src/Data/Array/Mutable/Unlifted/Linear.hs
facundominguez: > But this is considered a limitation of the implementation, and will be fixed.

It will be interesting to compare pure and monadic linear mutable arrays when this is fixed. This would make observable the cost introduced by NOINLINE.

src/Data/Array/Mutable/Unlifted/Linear.hs
aspiwack: I think that this explanation isn't quite accurate.

The requirement that you have is rather that, outside the definition of primitives, there is no aliasing of the array. It means, in particular, that if you return the array and something else, then the something else must never point in any way to the array. Otherwise, that something could observe a subsequent mutation.

It's pretty much the same story as in Rust, here.

Divesh-Otwani: Oh ... right! And that's why we need the `(GHC.runRW# ...)` thing to be evaluated when the `(read arr ix)` is evaluated. More pointedly then, is my argument for why the read is evaluated sound? From what @facundominguez pointed out, since  `(GHC.runRW# ...)` has a lifted type, it might not hold. In that case, are we sure that `case (GHC.runRW# ...) of (_, a) -> (Array arr, a)` can never be rewritten to `(Array arr, case (GHC.runRW# ...) of (_, a) -> a)`? If so, then we're still in the green zone.

PR #189, opened 2020-09-17
Implement Data.List.Linear

### LINKED ISSUE #170: Add Data.List or Data.Foldable
We should have the equivalent functions working on linear lists (or list-like collections).

On `base`, this is usually done by lists `Foldable` instance. There is a current PR #126 which creates a linear `Foldable` class. However, it does require some design decisions to be made (see the discussion on the PR). This issue is here to make sure that we implement an alternative if we decide to not go ahead with that PR.

utdemir:
Closes #170, discussion at #126 .

This PR adds linear versions of the majority of Data.List and Data.Foldable functions, specializing them to work on lists. Most of them are implemented by unsafe coerces, and the ones requiring `Consumable` or `Dupable` typeclasses reimplemented. I hope I got the casts right (meaning, they aren't omitting or duplicating any elements without using the required typeclasses).

I also omitted a few methods from `Data.List`, if they

* work on infinite lists, eg: `iterate`, `repeat`.
* discard most of the elements, eg: `head`, `find`, `null` (or they require significant changes on the type signatures)
* are partial, eg: `scan1`, `foldl1`

The latter two points are arbitrary decisions on my part, so let me know if you disagree and I can add them back. In my opinion, for those cases the user might find manually implementing them tailored to they needs easier. And we can always add them in future if someone requests them, or when we decide to.

It introduces a minor interface change:

```haskell
-- before
foldl :: (b #-> a -> b) -> b #-> [a] -> b
-- after
foldl :: (b #-> a #-> b) -> b #-> [a] #-> b
```

Which, I think is not a major issue since the former can be derived from latter by wrapping list elements with `Ur`. I had to do this for one of our examples.

Similar to #174, it only copies over minimal amount of documentation from `Data.List`, to reduce the amount of duplication. Let me know if you prefer otherwise. 

It also:

* Moves `forget` to `Prelude.Linear.Internal` to break an import cycle.
* It also adds `Consumable` instances for `NonEmpty` and `Either`, since they were necessary to reduce some duplication on zipping functions.
* Rewrite `Control.Monad.Linear.foldM` with explicit recursion instead of `foldr` to break an import cycle (because now `foldr` is imported from `Data.List.Linear` instead of `Prelude.Linear.Internal`).

### DIFF
--- a/src/Data/List/Linear.hs
+++ b/src/Data/List/Linear.hs
@@ -1,0 +1,219 @@
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+
+{-|
+Linear versions of 'Data.List' functions.
+
+This module only contains minimal amount of documentation; consult the
+original "Data.List" module for more detailed information.
+-}
+module Data.List.Linear
+  ( -- * Basic functions
+    map
+  , filter
+  , uncons
+  , reverse
+  , length
+  , splitAt
+  , span
+  , intersperse
+  , intercalate
+  , transpose
+  -- * Folds
+  , foldr
+  , foldl
+  , foldl'
+  , foldMap
+  , foldMap'
+  -- * Special folds
+  , concat
+  , concatMap
+  , and
+  , or
+  , any
+  , all
+  -- * Building lists
+  , scanl
+  , replicate
+  , unfoldr
+  -- * Zipping lists
+  , zip
+  , zip'
+  , zip3
+  , zipWith
+  , zipWith'
+  , zipWith3
+  ) where
+
+import qualified Unsafe.Linear as Unsafe
+import qualified Prelude as Prelude
+import Prelude (Maybe(..), Either(..), Int, otherwise)
+import Data.Num.Linear
+import Prelude.Linear.Internal
+import Data.Bool.Linear
+import Data.Unrestricted.Linear
+import Data.Functor.Linear
+import Data.Monoid.Linear
+import Data.List.NonEmpty (NonEmpty ((:|)))
+import qualified Data.List as NonLinear
+
+-- # Basic functions
+--------------------------------------------------
+
+map :: (a #-> b) -> [a] #-> [b]
+map = fmap
+
+-- | @filter p xs@ returns a list with elements satisfying the predicate.
+--
+-- See 'Data.Maybe.Linear.mapMaybe' if you do not want the 'Dupable' constraint.
+filter :: Dupable a => (a #-> Bool) -> [a] #-> [a]
+filter _ [] = []
+filter p (x:xs) =
+  dup x & \case
+    (x', x'') ->
+      if p x'
+      then x'' : filter p xs
+      else x'' `lseq` filter p xs
+
+uncons :: [a] #-> Maybe (a, [a])
+uncons [] = Nothing
+uncons (x:xs) = Just (x, xs)
+
+reverse :: [a] #-> [a]
+reverse = Unsafe.toLinear NonLinear.reverse
+
+-- | Return the length of the given list alongside with the list itself.
+length :: [a] #-> ([a], Ur Int)
+length = Unsafe.toLinear $ \xs ->
+  (xs, Ur (NonLinear.length xs))
+-- We can only do this because of the fact that 'NonLinear.length'
+-- does not inspect the elements.
+
+--  'splitAt' @n xs@ returns a tuple where first element is @xs@ prefix of
+-- length @n@ and second element is the remainder of the list.
+splitAt :: Int -> [a] #-> ([a], [a])
+splitAt i = Unsafe.toLinear (Prelude.splitAt i)
+
+-- | 'span', applied to a predicate @p@ and a list @xs@, returns a tuple where
+-- first element is longest prefix (possibly empty) of @xs@ of elements that
+-- satisfy @p@ and second element is the remainder of the list.
+span :: Dupable a => (a #-> Bool) -> [a] #-> ([a], [a])
+span _ [] = ([], [])
+span f (x:xs) = dup x & \case
+  (x', x'') ->
+    if f x'
+    then span f xs & \case (ts, fs) -> (x'':ts, fs)
+    else ([x''], xs)
+
+-- | The intersperse function takes an element and a list and
+-- `intersperses' that element between the elements of the list.
+intersperse :: a -> [a] #-> [a]
+intersperse sep = Unsafe.toLinear (NonLinear.intersperse sep)
+
+-- | @intercalate xs xss@ is equivalent to @(concat (intersperse xs
+-- xss))@. It inserts the list xs in between the lists in xss and
+-- concatenates the result.
+intercalate :: [a] -> [[a]] #-> [a]
+intercalate sep = Unsafe.toLinear (NonLinear.intercalate sep)
+
+-- | The transpose function transposes the rows and columns of its argument.
+transpose :: [[a]] #-> [[a]]
+transpose = Unsafe.toLinear NonLinear.transpose
+
+-- # Folds
+--------------------------------------------------
+
+foldr :: (a #-> b #-> b) -> b #-> [a] #-> b
+foldr f = Unsafe.toLinear2 (NonLinear.foldr (\a b -> f a b))
+
+foldl :: (b #-> a #-> b) -> b #-> [a] #-> b
+foldl f = Unsafe.toLinear2 (NonLinear.foldl (\b a -> f b a))
+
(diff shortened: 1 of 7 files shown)


aspiwack:
> * work on infinite lists, eg: `iterate`, `repeat`.

There is nothing wrong with linear functions for infinite lists. Presumably they will eventually be used in an unrestricted context. But this is not an issue

> * discard most of the elements, eg: `head`, `find`, `null` (or they require significant changes on the type signatures)

It's hard to justify linear `head` and `null` as they become identical to the corresponding pattern matching. `find` might, but it's not clear what the actual definition ought to be. You should re-export the unrestricted versions of these functions, though.

> * are partial, eg: `scan1`, `foldl1`

There is nothing wrong with partial function (as long as they have `HasCallStack` :slightly_smiling_face: ), so let's have these.

utdemir:
Thanks @aspiwack , I applied your suggestions, with below exception:

> It's hard to justify linear head and null as they become identical to the corresponding pattern matching. find might, but it's not clear what the actual definition ought to be. You should re-export the unrestricted versions of these functions, though.

I still think from an user perspective it would be surprising to import a linear module and get an unrestricted function, especially when there is a (admittedly not ideal) linear implementation too. So I tried to find a middle-ground, and implemented the linear versions:

```haskell
head :: (HasCallStack, Consumable a) => [a] #-> a
tail :: (HasCallStack, Consumable a) => [a] #-> [a]
last :: (HasCallStack, Consumable a) => [a] #-> a
init :: (HasCallStack, Consumable a) => [a] #-> [a]

find :: Dupable a => (a #-> Bool) -> [a] #-> Maybe a
```

Especially `find`'s implementation is pretty ugly; since for each element it has to first duplicate it, check one copy against the predicate, and then consume the other copy when the predicate fails.

I think this way things will be less surprising. Let me know if you still prefer the unrestricted versions instead of the above ones.

aspiwack:
>  So I tried to find a middle-ground, and implemented the linear versions:

These are mostly useless though. Plus, they are not compatible with the corresponding non-linear version. Same as record projections: they require unrestricted arguments. Even in a linearly typed module. I think that these functions ought to as well. We shouldn't require programmers to import two modules just for the sake of having mixed linear and non-linear code (which is the norm in linearly typed modules anyway).

utdemir:
> These are mostly useless though. 

I agree.

> We shouldn't require programmers to import two modules just for the sake of having mixed linear and non-linear code (which is the norm in linearly typed modules anyway).

I am not a fan of having to import multiple modules either. My omittance was mostly about providing a syntactic clue about whether something is working on linear or unrestricted values. Qualified imports are one way to achieve it, putting a suffix to the names would also work (but not pretty), or just having it absent from `.Linear` module is a strong clue that there is no linear version.

As you mentioned, whether a linear version of something is useful or not is a blurry line. Especially, for examples like `takeWhile` or `dropWhile`, I can think of them either way. So, I can imagine someone importing `Data.List.Linear` and assuming the `head` function to be `:: Consumable a => [a] #-> a`, and getting an error. It wouldn't even be clear that `head` is the issue, since the error messages point to the binding rather than the unrestricted use. So, it would take some cross-referencing docs to figure out one of the functions there is not linear on the parameter. I imagine that to be frustrating.

But, it is not a huge issue, and definitely not a hill I want to die on before the release. So I removed the linear implementations and re-exported the unrestricted ones from `Data.List`. We can always change our decision in the future.

src/Data/List/Linear.hs
```diff
@@ -0,0 +1,346 @@
...
+  , zip'
+  , zip3
+  , zipWith
+  , zipWith'
+  , zipWith3
+  , unzip
+  , unzip3
+  ) where
+
+import qualified Unsafe.Linear as Unsafe
```
Divesh-Otwani, line 77: No.......... I had hoped this wasn't going to be here. But, it might all work out :crossed_fingers:

src/Data/List/Linear.hs
```diff
@@ -0,0 +1,346 @@
...
+import Data.Num.Linear
+import Data.List.NonEmpty (NonEmpty ((:|)))
+import GHC.Stack
+import qualified Data.List as NonLinear
+
+-- # Basic functions
+--------------------------------------------------
+
+(++) :: [a] #-> [a] #-> [a]
+(++) = Unsafe.toLinear2 (NonLinear.++)
```
Divesh-Otwani, line 94: Why? The implementation is not optimized. Just re-write it.

src/Data/List/Linear.hs
```diff
@@ -0,0 +1,346 @@
...
+      if p x'
+      then x'' : filter p xs
+      else x'' `lseq` filter p xs
+
+uncons :: [a] #-> Maybe (a, [a])
+uncons [] = Nothing
+uncons (x:xs) = Just (x, xs)
+
+reverse :: [a] #-> [a]
+reverse = Unsafe.toLinear NonLinear.reverse
```
Divesh-Otwani, line 116: Just write it out, the implementation isn't optimized.

src/Data/List/Linear.hs
```diff
@@ -0,0 +1,346 @@
...
+length :: [a] #-> ([a], Ur Int)
+length = Unsafe.toLinear $ \xs ->
+  (xs, Ur (NonLinear.length xs))
+-- We can only do this because of the fact that 'NonLinear.length'
+-- does not inspect the elements.
+
+--  'splitAt' @n xs@ returns a tuple where first element is @xs@ prefix of
+-- length @n@ and second element is the remainder of the list.
+splitAt :: Int -> [a] #-> ([a], [a])
+splitAt i = Unsafe.toLinear (Prelude.splitAt i)
```
Divesh-Otwani, line 128: Okay, I think you should just make an issue for writing out the implementation for anything from `Data.List` that isn't optimized and not worry about it in this PR.

src/Prelude/Linear/Internal.hs
```diff
@@ -53,13 +53,10 @@ uncurry f (x,y) = f x y
...
-foldl :: (b #-> a -> b) -> b #-> [a] -> b
-foldl _ b [] = b
-foldl f b (x:xs) = foldl f (f b x) xs
+-- XXX: temporary: with multiplicity polymorphism functions expecting a
+-- non-linear arrow would allow a linear arrow passed, so this would be
+-- redundant
+-- | Convenience operator when a higher-order function expects a non-linear
+-- arrow but we have a linear arrow.
+forget :: (a #-> b) #-> a -> b
+forget f a = f a
```
Divesh-Otwani, line 62: I'll allow this sneak.

src/Data/List/Linear.hs
```diff
@@ -0,0 +1,346 @@
...
+zip3 :: (Consumable a, Consumable b, Consumable c) => [a] #-> [b] #-> [c] #-> [(a, b, c)]
+zip3 = zipWith3 (,,)
+
+zipWith :: (Consumable a, Consumable b) => (a#->b#->c) -> [a] #-> [b] #-> [c]
+zipWith f xs ys =
+  zipWith' f xs ys & \(ret, leftovers) ->
+    leftovers `lseq` ret
+
+-- | Same as 'zipWith', but returns the leftovers instead of consuming them.
+zipWith' :: (a#->b#->c) -> [a] #-> [b] #-> ([c], Maybe (Either (NonEmpty a) (NonEmpty b)))
```
Divesh-Otwani, line 329: Very nice type.

Divesh-Otwani (approved):
Nice and clean code. Please make an issue to eventually remove the gratuitous uses of `Unsafe` in this module. We should only use `Unsafe` when there's a non-linear optimized implementation already in `base` that we want to copy.

src/Prelude/Linear/Internal.hs
Divesh-Otwani: Nevermind, I see it's for an import cycle.

src/Data/List/Linear.hs
utdemir: I agree that currently this module isn't look very safe since we can not ensure that none of those unsafe casts actually drop/duplicate values.

However, `Data.List` has a lot of tricks to make things fast; and implementing them here would be more work. I think this matches your sentence here:

> We should only use Unsafe when there's a non-linear optimized implementation already in base that we want to copy.

(Also, it is likely that the `unsafeCoerce`'s there probably get in the way of many optimisations so I do not think this will be as fast as the `Data.List` even in this shape).

Divesh-Otwani: I might have looked at the wrong files but when I looked at what's underneath `Data.List` for some of the coerced functions I didn't find optimized versions. In any case, not something we need to worry about right now. I'm not worried about safety, since most of these look like they could have linear implementations. Still, we should have an issue to double check.

utdemir: Oh yes, I did not make that choice per-function basis. I just looked `Data.List` overall, saw some gnarly bits and decided to implement this module by coerces. You are probably right that some of the functions would be easy to duplicate here. I'll create an issue.

PR #194, opened 2020-09-18
Instances for `Monoidal (,) ()`

aspiwack:
These are ported from #79 where they were needed for traversal-type
things. These instances can't hurt. Even though I don't know what we
are going to do with traversals yet.

### DIFF
--- a/src/Data/Profunctor/Kleisli/Linear.hs
+++ b/src/Data/Profunctor/Kleisli/Linear.hs
@@ -43,6 +43,10 @@ instance Control.Applicative f => Strong Either Void (Kleisli f) where
   first  (Kleisli f) = Kleisli (either (Data.fmap Left . f) (Control.pure . Right))
   second (Kleisli g) = Kleisli (either (Control.pure . Left) (Data.fmap Right . g))
 
+instance Control.Applicative f => Monoidal (,) () (Kleisli f) where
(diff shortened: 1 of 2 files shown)


src/Data/Profunctor/Kleisli/Linear.hs
```diff
@@ -43,6 +43,10 @@ instance Control.Applicative f => Strong Either Void (Kleisli f) where
   first  (Kleisli f) = Kleisli (either (Data.fmap Left . f) (Control.pure . Right))
   second (Kleisli g) = Kleisli (either (Control.pure . Left) (Data.fmap Right . g))
 
+instance Control.Applicative f => Monoidal (,) () (Kleisli f) where
+  Kleisli f *** Kleisli g = Kleisli $ \(x,y) -> (,) Control.<$> f x Control.<*> g y
+  unit = Kleisli Control.pure
```
Divesh-Otwani, line 48: Perhaps use `-XInstanceSigs` just for readability. It's a bit of a pause figuring out the type in my head. Same for instances below.

Divesh-Otwani (approved):

src/Data/Profunctor/Kleisli/Linear.hs
aspiwack: If we are to add instance signatures, we should go over the whole module. So this would be a separate PR.

PR #195, opened 2020-09-18
Linear lens combinators

aspiwack:
A handful of utilities for linear lens.

These are ported from #79 .

### DIFF
--- a/src/Control/Optics/Linear/Internal.hs
+++ b/src/Control/Optics/Linear/Internal.hs
@@ -23,11 +23,11 @@ module Control.Optics.Linear.Internal
   , _Just, _Nothing
   , traversed
     -- * Using optics
-  , get, set, gets
+  , get, set, gets, setSwap
   , match, build
   , over, over'
   , traverseOf, traverseOf'
-  , lengthOf
+  , toListOf, lengthOf
   , withIso, withPrism
     -- * Constructing optics
   , iso, prism
@@ -38,6 +38,7 @@ import qualified Control.Arrow as NonLinear
 import qualified Data.Bifunctor.Linear as Bifunctor
 import Data.Bifunctor.Linear (SymmetricMonoidal)
 import Data.Profunctor.Linear
+import Data.Functor.Compose hiding (getCompose)
 import Data.Functor.Linear
 import qualified Data.Profunctor.Kleisli.Linear as Linear
 import Data.Void
@@ -97,6 +98,9 @@ over (Optical l) f = getLA (l (LA f))
 traverseOf :: Optic_ (Linear.Kleisli f) a b s t -> (a #-> f b) -> s #-> f t
 traverseOf (Optical l) f = Linear.runKleisli (l (Linear.Kleisli f))
 
+toListOf :: Optic_ (NonLinear.Kleisli (Const [a])) a b s t -> s -> [a]
+toListOf l = gets l (\a -> [a])
+
 get :: Optic_ (NonLinear.Kleisli (Const a)) a b s t -> s -> a
 get l = gets l P.id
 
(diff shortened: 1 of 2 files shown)


src/Data/Functor/Linear/Internal.hs
```diff
@@ -97,6 +98,14 @@ instance Applicative Identity where
...
   Identity f <*> Identity x = Identity (f x)
 
+instance (Functor f, Functor g) => Functor (Compose f g) where
+  fmap f (Compose x) = Compose (fmap (fmap f) x)
+
+instance (Applicative f, Applicative g) => Applicative (Compose f g) where
+   pure x = Compose (pure (pure x))
+   (Compose f) <*> (Compose x) = Compose (liftA2 (<*>) f x)
+   liftA2 f (Compose x) (Compose y) = Compose (liftA2 (liftA2 f) x y)
+
```
Divesh-Otwani, line 108: I like these. I needed them for `linear-streams`.

src/Control/Optics/Linear/Internal.hs
```diff
@@ -67,6 +70,10 @@ assoc = iso Bifunctor.lassoc Bifunctor.rassoc
 (.>) :: Optic_ arr a b s t -> Optic_ arr x y a b -> Optic_ arr x y s t
 Optical f .> Optical g = Optical (f P.. g)
 
+
+lens :: (s #-> (a, b #-> t)) -> Lens a b s t
```
Divesh-Otwani, line 74: Very nice; me likey :smirk:

src/Control/Optics/Linear/Internal.hs
```diff
@@ -139,3 +152,10 @@ withIso (Optical l) f = f fro to
 withPrism :: Optic_ (Market a b) a b s t -> ((b #-> t) -> (s #-> Either t a) -> r) -> r
 withPrism (Optical l) f = f b m
   where Market b m = l (Market id Right)
+
+withLens :: Optic_ (Linear.Kleisli (Compose ((,) a) (FUN 'One b))) a b s t -> s #-> (a, b #-> t)
```
Divesh-Otwani, line 156: Given that this is just migration and that I'm going to take a pass through optics, this, although complicated, is fine right now. When I take a pass I'm going to comment that `withLens` is (kind of) the dual of `lens` and the `(Optic_ ...)` is basically `(a #-> (a, b -> b)) -> (s #-> (s, t -> t))`. 

One question: could we change `FUN 'One b` to `(->) b` which is much more readable to me?

Divesh-Otwani (approved):
A simple migration :+1:

utdemir:
> fix the CI before merging. No idea why it's failing. Maybe some cut-and-paste of strangely encoded characters?

It's a weird issue. I had it once before, I think there's a cache invalidation issue somewhere. Previously I got around that by randomly changing a parameter name on one of the upstream modules. Totally unsatisfactory "solution", I know.

src/Control/Optics/Linear/Internal.hs
aspiwack: > `withLens` is (kind of) the dual of `lens`

The better term is _inverse_

> could we change `FUN 'One b to (->) b` which is much more readable to me?

It would not work. `(->)` is `FUN 'Many`. There are many ways to make this simpler, but I don't think that they are much needed. The type after `Optic_` is not meant to be read. See the `lens` library for how to document such primitives.

aspiwack:
I can reproduce the stack failure locally. With a clean `.stack` and `.stack-work`.

Stack version
```
Version 2.1.3, Git revision 636e3a759d51127df2b62f90772def126cdf6d1f (7735 commits) x86_64 hpack-0.31.2
```

I don't really have time to figure this out, right now, though.

Divesh-Otwani:
Okay, I'll play with parameter names and then merge.

aspiwack:
It does make me very sad, if that's the solution we end up settling on.

utdemir:
> It does make me very sad, if that's the solution we end up settling on.

I agree. I spent some time trying to solve it but to no avail. So I gave up and opened https://github.com/tweag/linear-base/pull/199, it's still not a solution; but should be better than having to change random stuff.

PR #198, opened 2020-09-22
Expand linear mutable HashMap interface

### LINKED ISSUE #164: Expand the linear HashMap interface

utdemir:
Closes #164 .

This PR expands HashMap interface with following main functionalities:

* An `alter` function which can be used to delete, insert or update keys.
* A `mapMaybe` function which can transform the HashMap optionally throwing out the elements.

Most of the rest of the changes are implemented in terms of the above functions.

* `mapMaybe` provides the `Functor` instance and `filter`-like functions
* `alter` provides `union` and `intersection` functions, which in turn becomes the `Semigroup` instance.

While implementing above changes, I had to add a `map` function to `Array`'s; because the `Functor` instance was overly restrictive:

```haskell
fmap :: (a #-> b) -> Array a #-> Array b
map  :: (a  -> b) -> Array a #-> Array b
```

Since the `Array` does not store linear values, the new `map` function allows using `a` unrestrictedly. `fmap` can then be derived from `map` by `forget`-ing the linear arrow.

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -49,6 +49,7 @@ module Data.Array.Mutable.Linear
     write,
     unsafeWrite,
     resize,
+    map,
     -- * Accessors
     read,
     unsafeRead,
@@ -66,10 +67,10 @@ import qualified Data.Array.Mutable.Unlifted.Linear as Unlifted
 import qualified Data.Functor.Linear as Data
 import qualified Data.Vector as Vector
 import qualified Data.Vector.Mutable as MVector
-import Prelude.Linear ((&))
+import Prelude.Linear ((&), forget)
 import qualified Data.Primitive.Array as Prim
 import System.IO.Unsafe (unsafeDupablePerformIO)
-import Prelude hiding (read)
+import Prelude hiding (read, map)
 
 -- # Data types
 -------------------------------------------------------------------------------
@@ -234,6 +235,9 @@ freeze (Array arr) =
    -- Once it is exposed, we should be able to replace it with something
    -- safer like: `go arr = Vector 0 (sizeof arr) arr`
 
+map :: (a -> b) -> Array a #-> Array b
+map f (Array arr) = Array (Unlifted.map f arr)
+
 -- # Instances
 -------------------------------------------------------------------------------
 
@@ -249,7 +253,7 @@ instance Dupable (Array a) where
      wrap (# a1, a2 #) = (Array a1, Array a2)
 
 instance Data.Functor Array where
-  fmap f (Array arr) = Array (Unlifted.map f arr)
+  fmap f arr = map (forget f) arr
 
 -- # Internal library
 -------------------------------------------------------------------------------
--- a/src/Data/HashMap/Linear.hs
+++ b/src/Data/HashMap/Linear.hs
@@ -16,36 +16,48 @@
 -- |
 -- This module provides mutable hashmaps with a linear interface.
 --
--- To use mutable hashmaps, create a linear computation of type
--- @HashMap k v #-> Ur b@ and feed it to 'empty'.
---
--- This hashmap is implemented with robin hood hashing which has good average
--- case performance.
+-- It is implemented with Robin Hood hashing which has amortized
+-- constant time lookups and updates.
 module Data.HashMap.Linear
   ( -- * A mutable hashmap
     HashMap,
     Keyed,
-    -- * Run a computation using a 'HashMap'
+    -- * Constructors
     empty,
-    -- * Modifiers and Constructors
+    fromList,
+    -- * Modifiers
     insert,
-    delete,
     insertAll,
+    delete,
+    filter,
+    filterWithKey,
+    mapMaybe,
+    mapMaybeWithKey,
+    shrinkToFit,
     -- * Accessors
     size,
+    capacity,
     lookup,
-    member
+    member,
+    toList,
+    -- * Combining maps
+    union,
+    unionWith,
+    intersectionWith
   )
 where
 
 import Data.Array.Mutable.Linear (Array)
+import qualified Data.Functor.Linear as Data
 import qualified Data.Array.Mutable.Linear as Array
 import Data.Hashable
 import Data.Unrestricted.Linear
-import Prelude.Linear hiding ((+), lookup, read)
+import Prelude.Linear hiding ((+), lookup, read, filter, mapMaybe)
 import Prelude ((+))
+import qualified Data.Maybe as NonLinear
+import qualified Data.Function as NonLinear
 import qualified Prelude
-import GHC.Stack
+import qualified Unsafe.Linear as Unsafe
 
 -- # Implementation Notes
 -- This is a simple implementatation of robin hood hashing.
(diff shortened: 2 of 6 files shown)


Divesh-Otwani:
> Since the `Array` does not store linear values, the new `map` function allows using `a` unrestrictedly. `fmap` can then be derived from `map` by `forget`-ing the linear arrow.

The other way around, `map` derived from `fmap`?

src/Data/Array/Mutable/Linear.hs
```diff
@@ -223,7 +227,7 @@ instance Consumable (Array a) where
   consume (Array arr) = arr `Unlifted.lseq` ()
 
 instance Data.Functor Array where
-  fmap f (Array arr) = Array (Unlifted.map f arr)
+  fmap f arr = map (forget f) arr
```
Divesh-Otwani, line 230: This is ugly, but the implementation is correct and I don't know of a way around this. It's a tricky part of HOFs, where we technically should use the implementation of `Data.fmap` to implement `map` since it is less restrictive and has the more specific implantation but we can't.

test/Test/Data/Mutable/HashMap.hs
```diff
@@ -60,15 +68,16 @@ type HMTest = HMap #-> Ur Bool
 maxSize :: Int
 maxSize = 800
 
+-- HashMap's have lots of corner cases, so we try harder to find them.
```
Divesh-Otwani, line 71: Nice.

src/Data/HashMap/Linear.hs
```diff
@@ -280,6 +432,136 @@ instance Consumable (HashMap k v) where
...
   Array.toList robinArr & \(Ur xs) -> show xs
+
+-- | When using Robin-Hood hashing, the beginning of the array
+-- can contain elements "spilled over" from the end of the array.
+--
+-- This function finds the number of spillovers. The spillovers can
+-- be determined by the length of the prefix of the array where
+-- PSL > index.
+numSpillovers :: RobinArr k v #-> (RobinArr k v, Ur Int)
+numSpillovers arr =
```
Divesh-Otwani, line 465: Technically this implementation is correct because spillovers form a contiguous segment (which is easy to prove). Just comment that.

src/Data/HashMap/Linear.hs
```diff
@@ -120,141 +149,255 @@ data ProbeResult k v where
...
+          ((sz - 1 + spillovers) `mod` sz)
+          sz
+          0
+          arr'''
+          & \(arr'''', Ur b) -> HashMap b arr''''
+ where
+  go :: Int -- ^ Current index
+     -> Int -- ^ Last index to check
+     -> Int -- ^ Size of the array
+     -> Int -- ^ Number of resulting elements
```
Divesh-Otwani, line 264: This is an accumulated count of pairs that are present after the map. Just change the comment please.

src/Data/HashMap/Linear.hs
```diff
@@ -120,141 +149,255 @@ data ProbeResult k v where
...
-          capacity (HashMap ct arr') & \(HashMap ct' arr'', Ur cap) ->
-            probeFrom (k, p+1) ((ix+1)`mod` cap) (HashMap ct' arr'')
+-- | A version of 'fmap' which can throw out the elements.
+--
+-- Complexity: O(capacity hm)
+mapMaybe :: Keyed k => (v -> Maybe v') -> HashMap k v #-> HashMap k v'
+mapMaybe f = mapMaybeWithKey (\_k v -> f v)
+
+-- | Same as 'mapMaybe', but also has access to the keys.
+mapMaybeWithKey :: Keyed k => (k -> v -> Maybe v') -> HashMap k v #-> HashMap k v'
```
Divesh-Otwani, line 245: Very nice.

src/Data/HashMap/Linear.hs
```diff
@@ -120,141 +149,255 @@ data ProbeResult k v where
...
+--
+-- This is only useful after a lot of deletes.
+--
+-- Complexity: O(capacity hm)
+shrinkToFit :: Keyed k => HashMap k a #-> HashMap k a
+shrinkToFit hm =
+  size hm & \(hm', Ur size) ->
+    let targetSize = ceiling
+          (max 1 (fromIntegral size / constMaxLoadFactor))
+    in  resize targetSize hm'
```
Divesh-Otwani, line 381: The comment of resize says the `targetSize` should be larger. Just note here that the use is safe because you checked things will fit.

src/Data/HashMap/Linear.hs
```diff
@@ -266,12 +409,21 @@ lookup hm k =
...
+    (hm, Ur (Just _)) -> (hm, Ur True)
+
+-- | Converts a HashMap to a lazy list.
+toList :: HashMap k v #-> Ur [(k, v)]
+toList (HashMap _ arr) =
+  Array.toList arr & \(Ur elems) ->
+    elems
+      NonLinear.& NonLinear.catMaybes
+      NonLinear.& Prelude.map (\(RobinVal _ k v) -> (k, v))
+      NonLinear.& Ur
```
Divesh-Otwani, line 426: Feel free to leave it as is but I find using `$` more readable.

src/Data/HashMap/Linear.hs
```diff
@@ -280,6 +432,136 @@ instance Consumable (HashMap k v) where
...
+  Int -> RobinArr k v #-> Int -> RobinArr k v
+shiftSegmentBackward s arr ix = Array.read arr ix & \case
+  (arr', Ur Nothing) -> arr'
+  (arr', Ur (Just (RobinVal 0 _ _))) -> arr'
+  (arr', Ur (Just val)) ->
+    Array.write arr' ix Nothing & \arr'' ->
+      shiftSegmentBackward
+        s
+        (Array.write arr'' ((ix-1) `mod` s) (Just $ decRobinValPSL val))
+        ((ix+1) `mod` s)
```
Divesh-Otwani, line 535: I'm assuming nothing really changed here.

src/Data/HashMap/Linear.hs
```diff
@@ -120,141 +149,255 @@ data ProbeResult k v where
...
+          -- We need to insert a new key.
+          Ur (Just v)->
+            capacity hm & \(HashMap count arr, Ur cap) ->
+              tryInsertAtIndex
+                (HashMap
+                  count
+                  (Array.write arr ix (Just (RobinVal psl key v))))
+                ((ix + 1) `mod` cap)
+                (incRobinValPSL evicted)
+              & growMapIfNecessary
```
Divesh-Otwani, line 219: I (politely) think this is a terrible idea. It's much harder to maintain code like this. Implement `alter` in terms of `insert` and `delete` and not the other way around. I'm blocking this PR for this.

src/Data/HashMap/Linear.hs
```diff
@@ -57,25 +69,42 @@ import GHC.Stack
 -- * https://cs.uwaterloo.ca/research/tr/1986/CS-86-14.pdf
 --
 
+-- * Constants
+
```
Divesh-Otwani, line 73: Not needed, but it would be nice if you could use the `-- # Constants` followed by some `-----`s just for consistency.

src/Data/HashMap/Linear.hs
```diff
@@ -120,141 +149,255 @@ data ProbeResult k v where
...
+    (\k v -> if f k v then Just v else Nothing)
+
+-- | Complexity: O(capacity hm)
+filter :: Keyed k => (v -> Bool) -> HashMap k v #-> HashMap k v
+filter f = filterWithKey (\_k v -> f v)
+
+-- | Union of two maps using the provided function on conflicts.
+--
+-- Complexity: O(min(capacity hm1, capacity hm2)
+unionWith
```
Divesh-Otwani, line 310: Nice.

src/Data/HashMap/Linear.hs
```diff
@@ -120,141 +149,255 @@ data ProbeResult k v where
...
+  capacity hm1 & \(hm1', Ur cap1) ->
+    capacity hm2 & \(hm2', Ur cap2) ->
+      if cap1 > cap2
+      then go combine hm1' hm2'
+      else go (\v2 v1 -> combine v1 v2) hm2' hm1'
+ where
+   go :: (a -> b -> c) -> HashMap k a #-> HashMap k b #-> HashMap k c
+   go f hm1 hm2 = Unsafe.toLinear (go' f) hm1 hm2
+
+   -- This function is using 'hm1' unsafely, because 'mapMaybeWithKey'
```
Divesh-Otwani, line 356: If you haven't yet, make an issue.

Divesh-Otwani (changes requested):
Very nice. My one demand is that `alter` be implemented in terms of `insert` and `delete`.

Beyond the other minor comments, good for merging :smile:

src/Data/HashMap/Linear.hs
Divesh-Otwani: After some discussion, I'm reconsidering. We can't implement in terms of `insert` and `delete` without re-probing and I can't find a way to re-write `alter` to be as shorter but as efficient as this version.

Divesh-Otwani (approved):

src/Data/HashMap/Linear.hs
aspiwack: Actually, it's probably needed. Haddock will probably freak out if it sees a `-- *` comment outside of an export list.

src/Data/Array/Mutable/Linear.hs
aspiwack: Yes, `map` is less restrictive than `fmap`. As the implementation here demonstrates.

PR #200, opened 2020-09-22
Added hIsEOF function

Divesh-Otwani:


src/System/IO/Resource.hs
```diff
@@ -163,6 +164,9 @@ openFile path mode = Control.do
 hClose :: Handle #-> RIO ()
 hClose (Handle h) = unsafeRelease h
 
+hIsEOF :: Handle #-> RIO (Ur Bool, Handle)
+hIsEOF = coerce (unsafeFromSystemIOResource System.hIsEOF)
+
```
facundominguez, line 169: I'm not suggesting that this PR should be changed. I just wanted to comment that of late, I'm finding the church-like encoding of booleans more natural to do effects.
```Haskell
ifIsEOF :: Handle #-> (Handle #-> RIO a) -> (Handle #-> RIO a) -> RIO a
```
The code reads more like pseudocode:
```Haskell
ifIsEOF h
  hClose
  readSomeMore
```

Divesh-Otwani: I think that's a good thing to add. For longer functions in the EOF case, we might want the `hIsEOF` style.

utdemir (approved):

PR #204, opened 2020-09-23
Implement Dupable instances for mutable collections

### LINKED ISSUE #165: Expand the linear mutable vector interface

utdemir:
Depends on #203, closes #165.

This PR implements `Dupable` instances for mutable arrays, vectors, hashmaps and sets.

It also adds a `dup2` function to unlifted mutable arrays, since we can not implement `Dupable` for unlifted types.

All the work is done by `GHC.Exts.cloneMutableArray#` through our unlifted arrays, and pretty much all other instances are implemented trivially by using `Dupable` instances of their contents.

### DIFF
--- a/src/Data/Array/Mutable/Unlifted/Linear.hs
+++ b/src/Data/Array/Mutable/Unlifted/Linear.hs
@@ -29,9 +29,10 @@ module Data.Array.Mutable.Unlifted.Linear
   , copyInto
   , map
   , toList
+  , dup2
   ) where
 
-import Data.Unrestricted.Linear hiding (lseq)
+import Data.Unrestricted.Linear hiding (lseq, dup2)
 import Prelude (Int)
 import qualified Prelude as Prelude
 import qualified Unsafe.Linear as Unsafe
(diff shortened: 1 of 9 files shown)


src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -158,6 +159,17 @@ toList = unArray# Prelude.$ \arr ->
...
+-- | Clone an array.
+dup2 :: Array# a #-> (# Array# a, Array# a #)
+dup2 = Unsafe.toLinear go
+ where
+  go :: Array# a -> (# Array# a, Array# a #)
+  go (Array# arr) =
+    case GHC.runRW#
+           (GHC.cloneMutableArray# arr 0# (GHC.sizeofMutableArray# arr)) of
+      (# _, new #) -> (# Array# arr, Array# new #)
+{-# NOINLINE dup2 #-}
```
Divesh-Otwani, line 171: Nice.

Divesh-Otwani (approved):
Awesome.

Issue #185, opened 2020-09-11
What immutable array type should linear base use?


aspiwack:
In the `Array.Destination`, the `alloc` function has type

```haskell
alloc :: Int -> (DArray a #-> ()) #-> Vector a
```

This is saying that our type of immutable array is `Vector a`. There is nothing wrong with it, to be honest. The vector library is standard enough to be part of a standard library at this point. I just want to make sure that we all agree on this.

utdemir:
I do not have any concerns about having `vector` dependency or exposing it as part of the public interface.

However, I wonder if that specific function should use `Data.Array.Mutable.Linear.Array` instead. I do understand that there are many good reasons to use the `vector` library, since it likely is more performant and comes with more utilities. However, it also does seem a bit strange to use an array type from an external library instead of using one of our own. If it is because of a shortcoming of our linear Array library, we should fix that.

I am not sure about this, but it might be possible to have a function like `Data.Array.Mutable.Linear.Array #-> Unrestricted Data.Vector.Vector` which runs in constant time (I'm even less sure about the other way around). If that is possible, the case for returning an `Array` instead of ` Vector` is stronger.

I know that you mentioned "immutable array type" in the title. Does the result being "immutable" is an important point? Because if we were to use our Arrays, we could both do linear stuff with them, or still convert it to an immutable `Vector` when necessary.

---

I think another relevant papercut is about the naming of our vector-like collections on `linear-base`, since we also have:

* `Data.Vector.Linear`, which is not a linear interface to `Data.Vector`, but a length indexed "vector". (It happens to use `Data.Vector` underneath, but just as an implementation detail)
* `Data.Vector.Mutable.Linear`, which is arguably the most similar to `vector` library, and might be seen as a linear interface to `Data.Vector.Mutable`.

This all is a bit confusing. I will create another issue about this.

aspiwack:
Well, my question was about immutable vectors, which we can used unrestrictedly. It's unrelated to the question of whether we also want to eliminate destinations to a mutable type.

For things which can be unrestricted, I'd like, as much as possible, to use types from the normal Haskell ecosystem. So that linear code and normal code interoperates. `Vector` is fine for this, so is the `Array` type from base. Though I think I like `Vector` a little bit more, for the case at hand.

utdemir:
Just to note that, with #205 merged, freezing a mutable array results on a `Data.Vector.Vector`.

aspiwack:
[...]

PR #205, opened 2020-09-23
Add a constant time 'freeze' function to mutable arrays and vectors

utdemir:
As suggested in https://github.com/tweag/linear-base/pull/180#issuecomment-694047048, this PR adds a constant time `freeze` method to unlifted and lifted mutable arrays and mutable vectors.

As usual, it is implemented with unsafe primitives in unlifted arrays, and the rest of the collections make use of that implementation. I hope I got the `MVector`'s parameters right, as far as I could understand its first parameter is the starting index and the second index is the length. Because of that starting index, we can not have a constant time `thaw` operation (which would be unsafe anyway).

I am not entirely sure about the name `freeze`. I was also thinking of names like `toVector` or `toImmutableVector`, but they are likelier to be confused with either of our `Vector`'s. Let me know if you prefer something else.

Some simple tests pass, and I hope I'm not violating any invariant of the `vector` package.

Relevant: https://github.com/tweag/linear-base/issues/185

### DIFF
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -55,6 +55,7 @@ module Data.Array.Mutable.Linear
     size,
     slice,
     toList,
+    freeze,
   )
 where
 
@@ -63,6 +64,7 @@ import GHC.Stack
 import Data.Array.Mutable.Unlifted.Linear (Array#)
 import qualified Data.Array.Mutable.Unlifted.Linear as Unlifted
 import qualified Data.Functor.Linear as Data
+import Data.Vector (Vector)
 import Prelude.Linear ((&))
 import Prelude hiding (read)
 
@@ -214,6 +216,10 @@ slice from targetSize arr =
 
     wrap :: (# Array# a, Array# a  #) #-> (Array a, Array a)
     wrap (# old, new #) = (Array old, Array new)
+
+-- | /O(1)/ Convert an 'Array' to an immutable 'Vector' (from 'vector' package).
+freeze :: Array a #-> Ur (Vector a)
+freeze (Array arr) = Unlifted.freeze arr
 
 -- # Instances
 -------------------------------------------------------------------------------
--- a/src/Data/Array/Mutable/Linear.hs
+++ b/src/Data/Array/Mutable/Linear.hs
@@ -64,8 +64,11 @@ import GHC.Stack
 import Data.Array.Mutable.Unlifted.Linear (Array#)
 import qualified Data.Array.Mutable.Unlifted.Linear as Unlifted
 import qualified Data.Functor.Linear as Data
-import Data.Vector (Vector)
+import qualified Data.Vector as Vector
+import qualified Data.Vector.Mutable as MVector
 import Prelude.Linear ((&))
+import qualified Data.Primitive.Array as Prim
+import System.IO.Unsafe (unsafeDupablePerformIO)
 import Prelude hiding (read)
 
 -- # Data types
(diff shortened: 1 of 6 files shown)


src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -158,6 +163,16 @@ toList = unArray# Prelude.$ \arr ->
         case GHC.runRW# (GHC.readArray# arr i#) of
           (# _, ret #) -> ret : go (i Prelude.+ 1) len arr
 
+-- | /O(1)/ Convert an 'Array#' to an immutable 'Vector' (from 'vector' package).
+freeze :: forall a. Array# a #-> Ur (Vector.Vector a)
```
aspiwack, line 167: I think it would have been natural, here, to convert to an unlifted immutable array. And post-convert to `Vector` in the unlifted case instead.

See [`unsafeFreezeArray#`](https://hackage.haskell.org/package/base-4.12.0.0/docs/GHC-Exts.html#v:unsafeFreezeArray-35-)

aspiwack (commented):
You say that you have freezing primitives to mutable vectors (`freeze` would be a rather bad name for them, maybe `share`?) but I don't see any. It's fine, we don't need them yet, I can't say that I have a use case for such, so it can wait after the first release. You can make an issue if you think they may be important.

Let's not worry about `thaw` for the time being (I don't agree with your assessment that it can't be done in `O(1)` though, but we can discuss it elsewhere).

I have a slight change in interface in mind below, let's consider it.

utdemir:
> You say that you have freezing primitives to mutable vectors (`freeze` would be a rather bad name for them, maybe `share`?) but I don't see any.

It's here:

https://github.com/tweag/linear-base/blob/2963e17d1a0f19b7493c48beb59ff13d1d61c280/src/Data/Vector/Mutable/Linear.hs#L279-L283

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: MVector's constructor is as follows:

```haskell
data MVector s a = MVector {-# UNPACK #-} !Int
                           {-# UNPACK #-} !Int
                           {-# UNPACK #-} !(MutableArray s a)
        deriving ( Typeable )
```

That `MutableArray` is from the `primitive` package:

```haskell
-- | Mutable boxed arrays associated with a primitive state token.
data MutableArray s a = MutableArray
  { marray# :: MutableArray# s a }
  deriving ( Typeable )
```

So, that forces us to go through `MutableArray`. The `Vector` constructor work as you mentioned:

```
data Vector a = Vector {-# UNPACK #-} !Int
                       {-# UNPACK #-} !Int
                       {-# UNPACK #-} !(Array a)
        deriving ( Typeable )
```

And it'd be ideal to use it, however that constructor is not exported.

aspiwack:
You misread me. But I also misread you. So let's drop this line of conversation altogether.

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: I now understand why you are asking for this, and I agree. Freezing an unlifted mutable array to an unlifted immutable array makes sense. I did the change.

Which caused unlifted arrays freeze function to have a different signature:

```haskell
freeze :: (GHC.Array# a -> b) -> Array# a #-> Ur b
```

Because I couldn't put an unlifted type inside a `Ur`. We are using the same pattern on `unArray#` function too, so I think it is fine.

aspiwack: Yes, it's safe. `Ur a` is isomorphic to `forall k. (a -> k) #-> k` (exercise!).

Ultimately, we probably want to expose an unlifted unrestricted thing. There is no obstacle except time and syntax.

aspiwack: Here is the corresponding issue, by the way: https://gitlab.haskell.org/ghc/ghc/-/issues/18490 .

src/Data/Array/Mutable/Linear.hs
```diff
@@ -215,6 +220,20 @@ slice from targetSize arr =
...
+freeze (Array arr) =
+  Unlifted.freeze go arr
+ where
+   go arr = unsafeDupablePerformIO $ do
+     mut <- Prim.unsafeThawArray (Prim.Array arr)
+     let mv = MVector.MVector 0 (Prim.sizeofMutableArray mut) mut
+     Vector.unsafeFreeze mv
+   -- We only need to do above because 'Vector' constructor is hidden.
+   -- Once it is exposed, we should be able to replace it with something
+   -- safer like: `go arr = Vector 0 (sizeof arr) arr`
```
aspiwack, line 235: Let's make an issue in our tracker to fix this. And we'll contribute upstream some way. After milestone 1.

aspiwack (approved):

src/Data/Array/Mutable/Linear.hs
```diff
@@ -215,6 +220,20 @@ slice from targetSize arr =
...
 
+-- | /O(1)/ Convert an 'Array' to an immutable 'Vector.Vector' (from
+-- 'vector' package).
+freeze :: Array a #-> Ur (Vector.Vector a)
+freeze (Array arr) =
+  Unlifted.freeze go arr
+ where
+   go arr = unsafeDupablePerformIO $ do
+     mut <- Prim.unsafeThawArray (Prim.Array arr)
+     let mv = MVector.MVector 0 (Prim.sizeofMutableArray mut) mut
```
Divesh-Otwani, line 231: I think this is correct because of the source of the [MVector instance for MVector](https://hackage.haskell.org/package/vector-0.12.1.2/docs/src/Data.Vector.Mutable.html#line-86).

Divesh-Otwani (approved):
This is really cool :smiley:

src/Data/Array/Mutable/Linear.hs
utdemir: I opened: https://github.com/tweag/linear-base/issues/214

PR #209, opened 2020-09-24
Remove unsound Eq and Ord instances

utdemir:
I found these instances on `linear-base`:

```haskell
-- This is sound because we consume all parts of a data type when we inspect
-- for equality
instance Ur.Eq a => Eq a where
  (==) = Unsafe.toLinear2 (Ur.==)

-- Note: the reason we coerce the unrestricted implementation
-- is because it's optimized for many base types
instance (Dupable a, Ur.Ord a) => Ord a where
  (<) = Unsafe.toLinear2 (Ur.<)
  (>) = Unsafe.toLinear2 (Ur.>)
  (>=) = Unsafe.toLinear2 (Ur.>=)
  (<=) = Unsafe.toLinear2 (Ur.<=)
```
However, despite the comments, I fail to see how they are sound; we can not just assume that `Ur.Eq` and `Ur.Ord` instances linearly consume the values. For example, the `Eq` instance allows this:

```haskell
unsafeConsume :: a #-> a #-> ()
unsafeConsume a b = consume (AlwaysEqual a == AlwaysEqual b)

newtype AlwaysEqual a = AlwaysEqual a
instance Ur.Eq a where _ == _ = True  
``` 

I believe the `Ord` instance is also problematic in another way that it doesn't call the `Dupable` instance. Also, those instances do not play well with the typechecker for some reason (it couldn't infer `Eq [a]` from `Ur.Eq a` on my experiments).

So, if I'm not mistaken above; I think we should remove those instances. Unfortunately this means that we have to implement `Eq` and `Ord` for most types from `base`. I didn't want to do all the work before we agree on the solution, so this PR is just to show how the change will look like.

Let me know if you're happy with the idea and I can implement instances for other `base` types, following the examples on this PR. Alternatively, tell me why I am wrong :).

### DIFF
--- a/src/Data/Eq/Linear.hs
+++ b/src/Data/Eq/Linear.hs
@@ -2,18 +2,19 @@
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE StandaloneDeriving #-}
 
 -- | This module provides a linear 'Eq' class for testing equality between
 -- values, along with standard instances.
 module Data.Eq.Linear
-  (
-    Eq(..)
+  ( Eq(..)
   )
   where
 
 import Data.Bool.Linear
-import qualified Prelude as Ur
-import qualified Unsafe.Linear as Unsafe
+import qualified Prelude
+import Data.Unrestricted.Linear
 
 -- | Testing equality on values.
 --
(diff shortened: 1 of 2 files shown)


Divesh-Otwani (approved):
I feel a by quite foolish. You are obviously correct. Suppose I have `Ur.==` on something like 

```haskell
data HalfLinear a b where
  HalfLinear :: a #-> b -> HalfLinear a b
instance Ur.Eq a => Ur.Eq (HalfLinear a b) where
  (HalfLinear x _) == (HalfLinear y _) = (x == y)
```
This would be catastrophic if we allowed it, inferred a linear instance and used it in say the API for file handles like `HalfLinear Handle Bool` or something. Thank you for catching this!

PR #217, opened 2020-09-25
Explicit Prelude.Linear re-exports

### LINKED ISSUE #172: Make Prelude.Linear re-exports of Prelude explicit, or replaced with linear versions
Currently, `Prelude.Linear` re-exports `Prelude` save a set of functions we implemented linearly. This is not ideal when there are functions that could be implemented linearly but we've missed for some reason.

It would be better if we were to re-export Prelude functions and data types explicitly, considering if they should be implemented linearly for each one of them. When we choose to not implement something linearly, or omit an export; ideally there should be a comment explaining the reason.

See the dicussion at: https://github.com/tweag/linear-base/pull/133#issuecomment-679234955

utdemir:
Closes #172, relevant discussion: https://github.com/tweag/linear-base/pull/133#issuecomment-679234955

This PR modifies `Prelude.Linear` to not export `module Prelude`, but replaces it with explicit re-exports.

I tried to keep the interface the same; but I found some linear implementations we could export instead of the unrestricted counterparts:

* We now export `Data.Eq.Linear` and `Data.Ord.Linear`, instead of exporting unrestricted equality and comparision functions.
* We now export linear versions of `fst` and `snd` instead of unrestricted ones from the `base`.

Also:

* We don't export any `Functor`s, because it isn't clear whether data or control `Functor`'s should be exported. However, we do export unrestricted `IO` instead of linear `IO`.
* I added `Maybe`, `Either` and `Ordering` types and constructors to their relevant linear modules, so they don't need to be exported separately.

While implementing this issue, I also implementing a few simple functions which were missing:

```haskell
mconcat :: Monoid a => [a] #-> a
asTypeOf :: a #-> a -> a

($!) :: (a #-> b) #-> a #-> b
($!) f a = Unsafe.toLinear2 (Prelude.$!) (forget f) a
```

I think the implementation `$!` is sound; even though it uses `a` twice, first to evaluate it and then to feed it to the function. Can you confirm this, @aspiwack ?

Let me know if you prefer anything different.

### DIFF
--- a/examples/Spec.hs
+++ b/examples/Spec.hs
@@ -23,12 +23,12 @@ import Prelude.Linear
 import Test.Hspec
 import Test.QuickCheck
 
-eqList :: forall a. (Manual.Representable a, Movable a, Eq a) => List a #-> List a #-> Ur Bool
+eqList :: forall a. (Manual.Representable a, Movable a, P.Eq a) => List a #-> List a #-> Ur Bool
 eqList l1 l2 =
     eqUL (move (List.toList l1)) (move (List.toList l2))
   where
     eqUL :: Ur [a] #-> Ur [a] #-> Ur Bool
-    eqUL (Ur as1) (Ur as2) = Ur (as1 == as2)
+    eqUL (Ur as1) (Ur as2) = Ur (as1 P.== as2)
 
 data InjectedError = InjectedError
   deriving (Typeable, Show)
@@ -43,7 +43,7 @@ main = hspec P.$ do
         property (\(l :: [Int]) -> unur (Manual.withPool $ \pool ->
           let
             check :: Ur [Int] #-> Ur Bool
-            check (Ur l') = Ur P.$ l' == l
+            check (Ur l') = Ur P.$ l' P.== l
           in
             check $ move (List.toList $ List.ofList l pool)))
 
@@ -76,7 +76,7 @@ main = hspec P.$ do
   describe "Off-heap heaps" P.$ do
     describe "sort" P.$ do
       it "sorts" P.$
-        property (\(l :: [(Int, ())]) -> Heap.sort l == (L.reverse P.$ L.sort l))
+        property (\(l :: [(Int, ())]) -> Heap.sort l P.== (L.reverse P.$ L.sort l))
 
   describe "Linear arrays" P.$ do
     describe "destination arrays" P.$ do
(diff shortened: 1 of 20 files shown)


examples/Spec.hs
```diff
@@ -23,12 +23,12 @@ import Prelude.Linear
 import Test.Hspec
 import Test.QuickCheck
 
-eqList :: forall a. (Manual.Representable a, Movable a, Eq a) => List a #-> List a #-> Ur Bool
+eqList :: forall a. (Manual.Representable a, Movable a, P.Eq a) => List a #-> List a #-> Ur Bool
```
aspiwack, line 26: One-letter qualifier! I know it's not your fault. But let's slate this one for being fixed, shall we?

src/Data/Vector/Mutable/Linear.hs
```diff
@@ -237,7 +237,7 @@ mapMaybe vec (f :: a -> Maybe b) =
   go r w s vec'
     -- If we processed all elements, set the capacity after the last written
     -- index and coerce the result to the correct type.
-    | r == s =
+    | r Prelude.== s =
```
aspiwack, line 240: I'm questioning why all these `==` have been replaced by `Prelude.==`, the linear one is more general.

src/Prelude/Linear.hs
```diff
@@ -23,20 +23,98 @@
...
-  , id
-  , seq
+    -- ** Basic data types
+    module Data.Bool.Linear
+  , Prelude.Char
+  , module Data.Maybe.Linear
+  , module Data.Either.Linear
+    -- * Tuples
+  , fst
+  , snd
```
aspiwack, line 33: I think these functions are not very useful. We ought to export the bae ones.

src/Prelude/Linear.hs
```diff
@@ -23,20 +23,98 @@
...
+  , Prelude.Double
+  , Prelude.Rational
+  , Prelude.Word
+  , module Data.Num.Linear
+  , Prelude.Real (..)
+  , Prelude.Integral (..)
+  , Prelude.Floating (..)
+  , Prelude.Fractional (..)
+  , Prelude.RealFrac (..)
+  , Prelude.RealFloat (..)
```
aspiwack, line 54: These are all classes that we probably want to have a linear version of eventually. Let's keep track of this.

src/Prelude/Linear.hs
```diff
@@ -23,20 +23,98 @@
...
+    -- *** Numeric functions
+  , Prelude.subtract
+  , Prelude.even
+  , Prelude.odd
+  , Prelude.gcd
+  , Prelude.lcm
+  , (Prelude.^)
+  , (Prelude.^^)
```
aspiwack, line 62: Presumably some of these can be linear too. Let's keep track too.

src/Prelude/Linear/Internal.hs
```diff
@@ -33,12 +33,18 @@ id x = x
...
+asTypeOf = const
+
 -- | @seq x y@ only forces @x@ to head normal form, therefore is not guaranteed
 -- to consume @x@ when the resulting computation is consumed. Therefore, @seq@
 -- cannot be linear in it's first argument.
 seq :: a -> b #-> b
 seq x = Unsafe.toLinear (Prelude.seq x)
 
+($!) :: (a #-> b) #-> a #-> b
+($!) f a = Unsafe.toLinear2 (Prelude.$!) (forget f) a
```
aspiwack, line 46: Can we re-implement this? No need for an unsafe call, which can block optimisation, for a one-liner.

aspiwack (approved):
Aside from `fst` and `snd`: looks good to me.

src/Prelude/Linear/Internal.hs
utdemir: I agree; but if I didn't know I could :). I'd implement it like this:

```
f $! a = a `seq` f a
```

However, that wouldn't be linear since `a` is used twice. But now I figured something like:

```
($!) :: (a #-> b) #-> a #-> b
($!) f !a = f a
```

works. I'll update the PR.

src/Data/Vector/Mutable/Linear.hs
utdemir: I think that was because after we merged https://github.com/tweag/linear-base/pull/209, we don't have many `Eq` instances we could use (however, it seems like it has an instance of `Int`, I'll double check). I have a PR in the making which adds some more instances, after that we can use our `==` more often, I hope.

src/Prelude/Linear/Internal.hs
aspiwack: First rule of using `Unsafe` is: look at the original implementation. (which, in this case, doesn't work if you copy paste, because of the peskly unrestricted-let limitation, but yours is a solution)

src/Prelude/Linear.hs
```diff
@@ -23,20 +23,98 @@
...
+  , Prelude.until
+  , asTypeOf
+  , Prelude.error
+  , Prelude.errorWithoutStackTrace
+  , Prelude.undefined
+  , seq
+  , ($!)
+    -- * List operations
+  , module Data.List.Linear
+  --  -- * Functions on strings
```
Divesh-Otwani, line 85: Extra `--` messes up the haddock

Divesh-Otwani (approved):
Strongly approve :+1: 

(Assuming the `P.something`s change to `Prelude.something`s.)

utdemir:
> (Assuming the P.somethings change to Prelude.somethings.)

I'll change this one in a separate PR.

src/Data/Vector/Mutable/Linear.hs
utdemir: They were unnecessary as you said, I started using linear versions.

utdemir:
I opened https://github.com/tweag/linear-base/issues/229 for the single-letter qualified imports issue.

PR #219, opened 2020-09-25
Change the parameter order for optics

aspiwack:
For reasons unknown, and almost certainly my fault, the order of the
type arguments of optic types (lens, traversal, …) was not the same as
in the lens library.

Even if I had a reason at the time (though I suspect it's just a
mistake), it couldn't have been strong enough to overcome the
principle of least surprise of being compatible with a well-used
package.

This will probably imply something for #108 .


utdemir (approved):
I remember being confused about them too, thanks!

Divesh-Otwani (approved):
I second @utdemir.

PR #221, opened 2020-09-25
Export optics

### LINKED ISSUE #137: Have at least one non-internal optics module
There is no way currently to use optics without importing an internal module. There is a `Control.Optics.Linear.Internal` module but no `Control.Optics.Linear`.

Add a `Control.Optics.Linear` module. If it's fine to export everything from `Control.Optics.Linear.Internal`, then just rename that one.

aspiwack:
I followed the structure of the lens library: functions are
re-exported in modules which focus on one particular optics. So one
can see what the most relevant operations are for specific
optics. Everything is then re-exported in one big
Control.Optics.Linear module.

Closes #137 .

### DIFF
--- a/src/Control/Optics/Linear.hs
+++ b/src/Control/Optics/Linear.hs
@@ -1,0 +1,15 @@
+module Control.Optics.Linear
+  ( Optic_(..)
+  , Optic
+  , module Control.Optics.Linear.Iso
+  , module Control.Optics.Linear.Lens
+  , module Control.Optics.Linear.Prism
+  , module Control.Optics.Linear.Traversal
+  )
+where
+
+import Control.Optics.Linear.Internal (Optic_(..), Optic)
+import Control.Optics.Linear.Iso
+import Control.Optics.Linear.Lens
+import Control.Optics.Linear.Prism
+import Control.Optics.Linear.Traversal
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -19,7 +19,12 @@ library
     Control.Monad.IO.Class.Linear
     Control.Monad.Linear
     Control.Monad.Linear.Internal
+    Control.Optics.Linear
     Control.Optics.Linear.Internal
+    Control.Optics.Linear.Iso
+    Control.Optics.Linear.Lens
+    Control.Optics.Linear.Prism
+    Control.Optics.Linear.Traversal
     Data.Array.Destination
     Data.Array.Mutable.Linear
     Data.Array.Mutable.Unlifted.Linear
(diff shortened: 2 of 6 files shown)


utdemir (approved):
I have mixed feelings about this. I also do like seeing things in conceptually manageable chunks, but just for the sake of grouping them, creating 4 different modules seem costly (given that the implementations are still all in `Internal`. Another alternative would be to just add sections to `Control.Optics.Linear` module, eg:

```haskell
  ( Optic_(..)
  , Optic
  
    -- * Iso
  , Iso, Iso'
    -- ** Composing optics
  , (.>)
    -- ** Common optics
  , swap, assoc
    -- ** Using optics
  , withIso
    -- ** Constructing optics
  , iso
  
    -- * Lens
  . Lens, Lens'
    -- ** Composing lens
  , (.>)
    -- ** Common optics
  , _1, _2
    -- ** Using optics
  , get, set, gets, setSwap
  , over, over'
  , withLens
    -- ** Constructing optics
  , lens
  
    -- * Prism
    ...
  )
```

But I do not feel strongly about this, so I'll leave it to you and @Divesh-Otwani's decision. I'd be happy with either.

src/Control/Optics/Linear.hs
```diff
@@ -0,0 +1,15 @@
+module Control.Optics.Linear
+  ( Optic_(..)
+  , Optic
+  , module Control.Optics.Linear.Iso
+  , module Control.Optics.Linear.Lens
+  , module Control.Optics.Linear.Prism
+  , module Control.Optics.Linear.Traversal
```
Divesh-Otwani, line 7: Please put in haddock headings here and make sure it looks nice with the inner headings having two stars `-- **`

Divesh-Otwani (approved):
I think the structure is great, particularly as optics grows.

src/Control/Optics/Linear.hs
Divesh-Otwani: Actually, nevermind, I'll do that in my documentation PR anyway.

aspiwack: Yes, there was no documentation here.

In this particular setup, though, the idea is that the main module simply contains links to the various other modules (the Haddock will just say, `module Iso` etc…)

Each of these module, on the other hand, is supposed to have a well-sectioned organisation.

I used this organisation, which is the same as that of the lens library, because optics module tend to get large, hence separating the exports into file may make sense.

But if we feel that we prefer having a single module with sections instead, I'm fine with it too. I don't have a strong opinion either way.

PR #222, opened 2020-09-25
Rename over'/traverseOf' -> overU/traverseOfU

aspiwack:
It's more in line with the rest of linear-base. In an ideal word we
don't even need these function and over & traverse are multiplicity
polymorphic. But we are not there yet.

Depends on #219 #221 .


aspiwack:

utdemir (approved):

Divesh-Otwani (approved):

Issue #220, opened 2020-09-25
The Traversal story


aspiwack:
Let me write some thought on the design of optics, and especially
traversals, so that I don't forget again. All this should make its way
into a document 

### Profunctor optics

When designing the optics library, I settled on profunctor optics
rather than a Van Laarhoven encoding for two reasons:
- Profunctor optics compose better (mostly because prisms always
  require some profunctoriness, which the lens library has to work around)
- It makes very clear how to make a `newtype` for `Optic_` that
  everything else is a restriction on.

As I'm writing this, I realise that we could have done `newtype Optic_
p f s t a b = Optical (p a (f b)) -> p s (f t))` and defined, say,
lens as `type Lens s t a b = forall f. Functor f => Optic (FUN 'One) s
t a b`), which counter the second point, but as I'm writing this
counter, another benefit of profunctors is coming back:

- Profunctors abstract over `FUN 'One` vs `(->) = FUN 'Many`.

The latter is used in lens for instance so that `over` can be linear
while `get` and `set` aren't. If we forced `FUN 'One`, then `set` and
`get` would still be definable, but would require additional
allocations, which would be a bit unfortunate.

So, anyway, profunctor it is. Not that it makes the problem of
`Traversal` any harder really, so whatever.

### Traversals

The first difficulty with traversals is that they can be defined in
two ways. Let me define a bunch of classes on profunctors:

```haskell
class Strong (,) () p where
  second :: p a b -> p (c, a) (c, b)
class Strong Either Void p where
  second :: p a b -> p (Either c a) (Either c b)
class Monoidal (,) () p where
  (***) :: p a b -> p c d -> p (a, c) (b, d)
  unit :: p () ()
class (Strong (,) () p, Strong Either Void p) => Wandering p where
  -- | Equivalently but less efficient in general:
  --
  -- > wander' :: Data.Traversable t => p a b -> p (t a) (t b)
  --
  -- It would be less efficient because `wander'` is trivially
  -- implemented as `wander traverse`, implementing `wander` in terms
  -- of `wander'` actually requires allocating some intermediate traversable
  -- structure.
  wander :: forall s t a b. (forall f. Control.Applicative f => (a %1 -> f b) -> s %1 -> f t) -> p a b -> p s t
```

Both `Strong` super-classes are implied by `wander'` since both `(c,)`
and `Either c` are traversable.

#### Dilemma

The first difficulty that we face is that there are two different
definitions for profunctor traversable in the literature.
[...]

PR #223, opened 2020-09-25
Add a function to create traversals

aspiwack:
I'm assuming, here, that traversals are built from the same type as
`Traverse`. I believe this to be the right type, which I explain in
some details in #220.

Depends on  #219 #221 .

### DIFF
--- a/src/Data/Profunctor/Linear.hs
+++ b/src/Data/Profunctor/Linear.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeOperators #-}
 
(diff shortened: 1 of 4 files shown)


aspiwack:

src/Data/Profunctor/Linear.hs
```diff
@@ -58,7 +59,10 @@ class (SymmetricMonoidal m u, Profunctor arr) => Strong m u arr where
   {-# INLINE second #-}
 
 class (Strong (,) () arr, Strong Either Void arr) => Wandering arr where
-  wander :: Data.Traversable f => a `arr` b -> f a `arr` f b
+  -- | Equivalently but less efficient in general:
+  --
+  -- > wander :: Data.Traversable f => a `arr` b -> f a `arr` f b
+  wander :: forall s t a b. (forall f. Control.Applicative f => (a #-> f b) -> s #-> f t) -> a `arr` b -> s `arr` t
```
Divesh-Otwani, line 65: I like this much better actually.

Divesh-Otwani (approved):

src/Data/Profunctor/Linear.hs
aspiwack: In principle, I like the previous one better. Like, it's theoretically less ad hoc. But on the other hand, it's also less general in a sense.

PR #224, opened 2020-09-25
withLens & reifyLens

aspiwack:
The `withX` family of function is continuation-passing style. Probably
to avoid some allocations. The withLens function was not of this
shape, so I think it deserves another name. I like `reifyLens` for it,
so I used it.

I also added a quick implementation of a `withLens` function in terms
of `reifyLens`. Most certainly we can do a more direct implementation,
but I didn't take the time to think about it.

The getter/setter split of lenses doesn't really make sense for linear
lens. However, there is a presentation using an existential type which
does have such a shape. I decided to use it for `withLens`, by
symmetry with `withIso` and `withPrism`.

Depends on #219 #221 .

### DIFF
--- a/src/Control/Optics/Linear/Internal.hs
+++ b/src/Control/Optics/Linear/Internal.hs
@@ -29,6 +29,7 @@ module Control.Optics.Linear.Internal
   , over, over'
   , traverseOf, traverseOf'
   , toListOf, lengthOf
+  , reifyLens
   , withIso, withLens, withPrism
     -- * Constructing optics
   , iso, lens, prism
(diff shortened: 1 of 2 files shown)


aspiwack:

utdemir (approved):

src/Control/Optics/Linear/Internal.hs
```diff
@@ -132,30 +133,37 @@ build (Optical l) x = Linear.runCoKleisli (l (Linear.CoKleisli getConst')) (Cons
...
   where Market b m = l (Market id Right)
 
-withLens :: Optic_ (Linear.Kleisli (Compose ((,) a) (FUN 'One b))) a b s t -> s #-> (a, b #-> t)
-withLens (Optical l) s = getCompose (Linear.runKleisli (l (Linear.Kleisli (\a -> Compose (a, id)))) s)
+-- XXX: probably a direct implementation would be better
+withLens
+  :: Optic_ (Linear.Kleisli (Compose ((,) a) (FUN 'One b))) s t a b
+  -> (forall c. (s #-> (c, a)) -> ((c, b) #-> t) -> r)
+  -> r
+withLens l k = k (Bifunctor.swap . (reifyLens l)) (uncurry ($))
```
Divesh-Otwani, line 163: This is dense but ok.

Divesh-Otwani (approved):

src/Control/Optics/Linear/Internal.hs
aspiwack: There is probably a better definition. One that doesn't indirect through `reifyLens`. I just went for the fastest implementation I could do (though I do find this particular definition illuminating, lenses are a deep and fascinating subject).

PR #226, opened 2020-09-25
Implement Ord in terms of `compare`

utdemir:
Previously the `Ord` typeclass was implemented in terms of comparison operators (`<`, `>=` eg). And since a linear `compare` does not follow from them, it was implemented as:

```haskell
compare :: (Dupable a, Ur.Ord a) => a #-> a #-> Ordering
compare = Unsafe.toLinear2 Ur.compare
```

But, as established at #209, this approach is unsound. So that would force us to remove the `compare` function. However, that is unfortunate, because in the linear setting `compare` is strictly more powerful than the others. As another point, when implementing `Ord` instances for composite types, `compare` is a lot more useful since it saves us from implementing two separate, but very similar looking functions (`<` and `>`, for example).

So, the natural conclusion is to reverse the current state of affairs, with implementing `Ord` typeclass in terms of only `compare`, and implementing the comparison operators outside of the typeclass. This way:

* We can have both comparison operators and `compare` function.
* Implementing instances is easier with `compare`. This can be seen on the instances in this patch.

Another solution would be to add `compare` to the typeclass definition alongside with the others. But that will make the comparison functions redundant since `compare` would be in each minimal complete definition anyway.

This PR implements the above change, and it adds some `Ord` (and therefore `Eq`) instances necessary for commonly used data types.

### DIFF
--- a/src/Data/Eq/Linear.hs
+++ b/src/Data/Eq/Linear.hs
@@ -1,7 +1,5 @@
-{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE NoImplicitPrelude #-}
-{-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE StandaloneDeriving #-}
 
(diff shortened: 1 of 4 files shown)


Divesh-Otwani:
Hey I don't think this PR is sound on principle.

> Previously the `Ord` typeclass was implemented in terms of comparison operators (`<`, `>=` eg). And since a linear `compare` does not follow from them, it was implemented as:
> 
> ```haskell
> compare :: (Dupable a, Ur.Ord a) => a #-> a #-> Ordering
> compare = Unsafe.toLinear2 Ur.compare
> ```
> 
> But, as established at #209, this approach is unsound. 

Yes, but if you use `Ord` instead of `Ur.Ord` it's not unsound. What was unsound was inferring linear orderings from unrestricted implementations.

> So that would force us to remove the `compare` function. However, that is unfortunate, because in the linear setting `compare` is strictly more powerful than the others. As another point, when implementing `Ord` instances for composite types, `compare` is a lot more useful since it saves us from implementing two separate, but very similar looking functions (`<` and `>`, for example).
> 
> So, the natural conclusion is to reverse the current state of affairs, with implementing `Ord` typeclass in terms of only `compare`, and implementing the comparison operators outside of the typeclass. This way:
> 
>     * We can have both comparison operators and `compare` function.

Don't need to do this to have both.

> 
>     * Implementing instances is easier with `compare`. This can be seen on the instances in this patch.

That doesn't matter because we can't eliminate what might be a useful concept in some linear applications.

It's not a question of easier instances. A linear comparison might have a use case that only `<=` is allowed because it consumes the arguments. Maybe adding other class like `Comparable` would be nice but `<` and `compare` are strictly different concepts in the linear setting. The more fundamental one is `<=` since `compare` is more powerful.

Perhaps a re-working of this PR renaming `Ord` to `StrictOrd` and adding `Ord` that implies `StrictOrd` but has a linear compare instance would work nicely. However, as presented this isn't sound on principle.

aspiwack:
@Divesh-Otwani I don't follow.

It is quite clear that you can define all the boolean comparison operator from `compare`.

It is also the case that `compare` is the right concept as far `Ord` is concerned.

However, I don't think it'd hurt having the boolean comparisons in the `Ord` class. And, since they are in the class in base, I'd say we should have everything in. And yes, it does imply that there is a single minimal set, namely `compare`, but it's fine.

Divesh-Otwani:
> It is also the case that `compare `is the right concept as far `Ord` is concerned.

Oh, yes, what I'm thinking of is a partially ordered set and we want `Ord` to be a linear ordering, for which I agree we should use `compare` as the minimal definition. This is fine as is. In the event that we want a linear partial order, we can just make one as @utdemir suggested.

utdemir:
> However, I don't think it'd hurt having the boolean comparisons in the Ord class. And, since they are in the class in base, I'd say we should have everything in. And yes, it does imply that there is a single minimal set, namely compare, but it's fine.

This makes sense, I updated the PR.

src/Data/Eq/Linear.hs
```diff
@@ -34,9 +32,46 @@ class Eq a where
   x /= y = not (x == y)
   infix 4 ==, /=
 
+-- * Instances
+
+instance Prelude.Eq a => Eq (Ur a) where
```
Divesh-Otwani, line 37: Clever.

src/Data/Ord/Linear.hs
```diff
@@ -61,46 +34,99 @@ point. (And thus there are 4 minimal definitions.)
...
+min x y =
+  dup2 x & \(x', x'') ->
+    dup2 y & \(y', y'') ->
+      if x' <= y'
+      then y'' `lseq` x''
+      else x'' `lseq` y''
+
+-- * Instances
+
+instance Prelude.Ord a => Ord (Ur a) where
```
Divesh-Otwani, line 86: Clever, again.

Divesh-Otwani (approved):

src/Data/Ord/Linear.hs
```diff
@@ -61,46 +34,99 @@ point. (And thus there are 4 minimal definitions.)
 -- * @x <= y@ = @not (y > x)@
 -- * @x >= y@ = @not (y < x)@
 --
+-- Unlike in the non-linear setting, a linear @compare@ doesn't follow from
+-- @<=@ since it requires calls: one to @<=@ and one to @==@. However,
```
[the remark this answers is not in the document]
Divesh-Otwani: I'll merge anyway since it's clear enough.

PR #228, opened 2020-09-26
Consistent ordering of linear return values

### LINKED ISSUE #191: Order of the returned tuple when threading linear values varies across different modules
Pretty pointless bikeshedding issue ahead, I apologize.

Working with linear values usually require threading the linear value from function to function, and when functions also return another result, they return a tuple of the linear value and a (usually unrestricted) result.

However, the order of the linear value and the result at the returned tuple varies from module to module in `linear-base`. Here're some examples:

```haskell
-- System.IO.Resource and polarized array modules return the threaded value on `snd`
hGetChar :: Handle #-> RIO (Ur Char, Handle)
findLength :: Pull.Array a #-> (Int, Array a)

-- Linear mutable collections (arrays, vectors, hashmaps and sets) return the thread value on `fst`
size :: Vector a #-> (Vector a, Ur Int)
read :: HasCallStack => Vector a #-> Int -> (Vector a, Ur a)
```

As a user, although it is not a huge deal because of the types, it would still be annoying to having to remember where to expect the return value; I think there is value in consistency. So, we should make a decision here.

In my opinion, there is no clear advantage of either direction. So, let's pick one and use it across the codebase, if there is no objection.

utdemir:
Closes #191 .

This PR flips the returned tuples at linear mutable collections, so the "actual" return value is at the `fst`, and the collection to thread linearly is at `snd`.

No other changes. If there are any (not type-checked) documentation referring to the type signatures, they need to be fixed too.

See the discussion at #191 for more details.

### DIFF
--- a/examples/Simple/TopSort.hs
+++ b/examples/Simple/TopSort.hs
@@ -67,16 +67,16 @@ postOrderHM nodes dag = findSources nodes (computeInDeg nodes dag) & \case
   -- Increment in-degree of all neighbors
   incChildren :: InDegGraph #-> Ur Node #-> InDegGraph
   incChildren dag (Ur node) = HMap.lookup dag node & \case
-     (dag, Ur Nothing) -> dag
-     (dag, Ur (Just (xs,i))) -> incNodes (move xs) dag
+     (Ur Nothing, dag) -> dag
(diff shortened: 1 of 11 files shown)


Divesh-Otwani (approved):
Awesome :smiley:

PR #232, opened 2020-09-28
Rename single-letter qualified imports

### LINKED ISSUE #229: Rename one-letter qualified imports in linear-base
Some of the modules are imported as single-letter acronyms, which is not readable. eg at: `examples/Spec.hs`:

```haskell
import qualified Prelude as P
```

We should use more descriptive names here.

utdemir:
First stab at #229, depends on #226 .

It depends on #226 since it changes some `Prelude.==` invocations to `Prelude.Linear.==`. It's probably easier to wait until that is merged before reviewing this.

Omissions: 

* There is one `Data.Vector.Linear as V` re-export. However it is already going to be fixed by #227 .
* There is a single-letter qualified import on optics module, but I'm not touching it since there's a few PR's open and I do not want to create a conflict.

At the same time, I found some functions that were missing from #189 , and I sneakily re-exported them in this PR.


src/Streaming/Prelude/Linear.hs
```diff
@@ -24,9 +24,9 @@
 
    Other libraries that come up in passing are
 
-> import qualified Control.Foldl as L -- cabal install foldl
-> import qualified Pipes as P
-> import qualified Pipes.Prelude as P
+> import qualified Control.Foldl as Foldl -- cabal install foldl
+> import qualified Pipes
+> import qualified Pipes.Prelude as Pipes
```
Divesh-Otwani, line 29: Don't do it here. You can do this but then you should change it consistently throughout the haddock which will take forever.

aspiwack: I do really want single-letter imports gone everywhere. They are bad practice, and I don't want our examples to use encourage bad practices.

Divesh-Otwani: Okay, but doing it throughout all the streaming docs that took me more than a day to read though isn't a good use of time right now. I think we should make an issues for post-release.

utdemir: Then I will undo this change, and keep the issue open, no need to make the Haddock changes hold up this PR.

src/Data/List/Linear.hs
```diff
@@ -64,6 +64,10 @@ module Data.List.Linear
   , cycle
   , iterate
   , unfoldr
+  -- * Ordered lists
+  , NonLinear.sort
+  , NonLinear.sortOn
+  , NonLinear.insert
```
Divesh-Otwani, line 70: Sneaks are a bit annoying but ok.

Divesh-Otwani (approved):

PR #237, opened 2020-09-29
Add doctest

### LINKED ISSUE #235: Integrate `doctests` to type-check inline examples
Since `linear-base` codebase is changing a lot these days, inline code snippets in our Haddock's can get out of sync, as seen on #234.

We should integrate `doctest`, so our CI can catch those errors. This will change the overall style of the examples a bit, since they currently demonstrate entire modules; but with `doctests` they will look more like REPL sessions. eg:

```haskell
>>> :set -XLinearTypes
>>> :set -XQualifiedDo
>>> import qualified System.IO.Resource as Linear
>>> import qualified Control.Monad.Linear as Control
>>> import qualified Data.Text as Text
>>> import Prelude.Linear
>>> import qualified Prelude
>>> :{
 linearWriteToFile :: IO ()
 linearWriteToFile = Linear.run Prelude.$ Control.do
   handle1 <- Linear.openFile "/home/user/test.txt" WriteMode
   handle2 <- Linear.hPutStrLn handle1 (Text.pack "hello there")
   () <- Linear.hClose handle2
   Control.return (Ur ())
:}
```

To me, this is unfortunate. But I couldn't think of a better way to keep those examples correct; and the benefits outweigh the cost, I think.

utdemir:
Closes #235.

This PR integrates `doctest` library to type-check and test the code examples in our Haddock's.

* It modifies our examples to look like REPL sessions instead of entire modules.
* There is an annoying `doctest` issue where we need to put an empty line after a multiline declaration, which doesn't look great. I'll file an upstream issue.
* As discussed on #232; `Streaming` modules are pretty big, so refactoring their examples would take a long time. So, this PR escapes their examples so they are not recognized by `doctest`.
* We're adding two new overrides for `syb` and `doctest`, both overrides have upstream PR's.

In order to try the tests, run `cabal test linear-base:doctests`, or an equivalent `stack` command. It is also a part of the longer `cabal test` command.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -19,3 +19,15 @@ source-repository-package
   location: https://github.com/utdemir/haskell-hedgehog.git
   tag: c98aa9e33bf6871098d6f4ac94eeaac10383d696
   subdir: hedgehog
+
+source-repository-package
+  -- https://github.com/dreixel/syb/pull/25
+  type: git
+  location: https://github.com/utdemir/syb.git
+  tag: a072e20d7a5806779486646418fa20c5ffd99a4f
+
+source-repository-package
+  -- https://github.com/sol/doctest/pull/272
+  type: git
+  location: https://github.com/utdemir/doctest.git
+  tag: 090cccaf12c5643ca80f8c2afe693a488277d365
--- a/Setup.hs
+++ b/Setup.hs
@@ -1,2 +1,6 @@
-import Distribution.Simple
-main = defaultMain
+module Main where
+
+import Distribution.Extra.Doctest (defaultMainWithDoctests)
+
+main :: IO ()
+main = defaultMainWithDoctests "doctests"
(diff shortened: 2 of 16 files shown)


cabal.project
```diff
@@ -19,3 +19,15 @@ source-repository-package
...
+  -- https://github.com/dreixel/syb/pull/25
+  type: git
+  location: https://github.com/utdemir/syb.git
+  tag: a072e20d7a5806779486646418fa20c5ffd99a4f
+
+source-repository-package
+  -- https://github.com/sol/doctest/pull/272
+  type: git
+  location: https://github.com/utdemir/doctest.git
+  tag: 090cccaf12c5643ca80f8c2afe693a488277d365
```
Divesh-Otwani, line 33: Make an issue to make these tweag forks and not your forks.

Divesh-Otwani (approved):

PR #238, opened 2020-09-29
Documenting destination arrays

Divesh-Otwani:
This replaces #110 but takes much of the comments from there. The examples all typecheck but doctest isn't working with me so I'm leaving that to @utdemir to choose if he wants to fix as he reviews.

### DIFF
--- a/src/Data/Array/Destination.hs
+++ b/src/Data/Array/Destination.hs
@@ -2,9 +2,95 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
+-- | This module provides destination arrays
+--
+-- == What are destination arrays? What are they good for?
+--
+-- Destination arrays are write-only arrays that are only allocated once,
+-- thereby avoiding your reliance on GHC's fusion mechanisms to remove
+-- unneccessary allocations.
+--
+-- The current status-quo for computations that have a write-only array
+-- threaded along is to rely on fusion. While the optimizations in say,
+-- `Data.Vector` are quite good at ensuring GHC fuses, they aren't
+-- foolproof and can sometimes break by simple refactorings.
+--
+-- Avoiding extra allocations of a write-only array is easy in C, with
+-- something the functional programming world calls destination passing style,
+-- or DPS for short.
+--
+-- Here is a C function that manipulates an array written in DPS style; it
+-- takes in the destiniation array @res@ and writes to it:
+--
+-- @
+-- // ((a + b) * c) for vectors a,b and scalar c
+-- void apbxc(int size, int *a, int *b, int c, int *res){
+--   for (int i=0; i<size;++i){res[i]=a[i]+b[i];}
+--   mult(size, c, res);
+-- }
+--
+-- void mult(int size, int scalar, int* vec){
+--   for (int i=0; i<size; ++i){vec[i] *= scalar;}
+-- }
+-- @
+--
+-- We could write this with @Data.Vector@ in haskell and rely on GHC to fuse
+-- (but this isn't great in complex uses where some simple changes can
+-- break the fusion):
+--
+-- @
+-- import Data.Vector (Vector)
+-- import qualified Data.Vector as Vector
+-- import qualified Prelude
+--
+-- apbxc :: Vector Int -> Vector Int -> Int -> Vector Int
+-- apbxc a b c = mult c Prelude.$ Vector.zipWith (Prelude.+) a b
+--
+-- mult :: Int -> Vector Int -> Vector Int
+-- mult c = Vector.map (Prelude.* c)
+-- @
+--
+-- Doing this with destination arrays we are guarenteed only one allocation
+-- of the resulting array:
+--
+-- @
+--  apbxcDest :: Vector Int -> Vector Int -> Int -> Vector Int
+--  apbxcDest a b c = alloc (Vector.length a) writeResult
+--    where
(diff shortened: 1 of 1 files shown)


src/Data/Array/Destination.hs
```diff
@@ -2,9 +2,95 @@
...
+-- Doing this with destination arrays we are guarenteed only one allocation
+-- of the resulting array:
+--
+-- @
+--  apbxcDest :: Vector Int -> Vector Int -> Int -> Vector Int
+--  apbxcDest a b c = alloc (Vector.length a) writeResult
+--    where
+--      writeResult :: DArray Int #-> ()
+--      writeResult = fromFunction (\i -> ((a ! i) + (b ! i)) * c)
+-- @
```
utdemir, line 62: I think this is not a great example, since the destination array examples does a single operation, hence there is nothing to fuse. However, in `vector` case, the intermediate vector can appear after `zipWith`, but before `mult`. An eqivalent `vector` example would be something like

```
apbxc a b c = Vector.zipWith (\i j -> (i + j) * c)
```

which doesn't have the fusion problem either. So, I think in this example destination array doesn't provide much value. But, I don't know how to write the current `vector` example with destination arrays since they're write-only (but it is doable with push-pull arrays).

I think an example to demonstrate the advantage of destination arrays would include `split` in some way, so the two halves can be filled on different functions which doesn't allocate.

src/Data/Array/Destination.hs
```diff
@@ -2,9 +2,95 @@
...
+-- array is written to exactly once. This is because the only way to create and
+-- use a destination array is via
+--
+-- @
+-- alloc :: Int -> (DArray a #-> ()) #-> Vector a
+-- @
+--
+-- and the only way to really consume a @DArray@ is via our API
+-- which requires you to completely fill the array.
+--
```
utdemir, line 86: In here, we can also mention that they are performant since they are implemented with in-place updates on a mutable vector; which can be done safely because of linear-types.

src/Data/Array/Destination.hs
```diff
@@ -70,22 +148,21 @@ fill = Unsafe.toLinear2 unsafeFill
...
   where
     unsafeSplit (DArray ds) =
       let (dsl, dsr) = MVector.splitAt n ds in
         (DArray dsl, DArray dsr)
 
--- | Convert a Vector into a destination array to be filled, possibly with
--- a conversion function.
--- Assumes both arrays have the same size.
+-- | Convert a Vector into a destination array to be filled, possibly with a
+-- conversion function.  Assumes both arrays have the same size.
```
utdemir, line 161: Wouldn't it be more straightforward to say something like:

> Fills the destination array with the contents of given vector

Because, to me, the result is not "a destination array to be filled".

utdemir (approved):
I really like this new introduction. My only major comment is about the example, otherwise I'm happy with it.

src/Data/Array/Destination.hs
aspiwack: You could also define destination-passing versions of `zip` and `map`, and show that they chain well. It would illustrate fusion nicely.

```haskell
map :: (a -> b) -> Vector a -> DArray b #-> ()
map f as = fromFunction $ \i -> f (a ! i)
```

etc…

Then you probably want to link to the polarised fusion story for a higher level abstraction of this pattern.

src/Data/Array/Destination.hs
```diff
@@ -2,9 +2,95 @@
...
+-- mult :: Int -> Vector Int -> Vector Int
+-- mult c = Vector.map (Prelude.* c)
+-- @
+--
+-- Doing this with destination arrays we are guarenteed only one allocation
+-- of the resulting array:
+--
+-- @
+--  apbxcDest :: Vector Int -> Vector Int -> Int -> Vector Int
+--  apbxcDest a b c = alloc (Vector.length a) writeResult
```
aspiwack, line 58: It should be
```suggestion
--  apbxcDest a b c = alloc (min (Vector.length a) (Vector.length b)) writeResult
```

src/Data/Array/Destination.hs
aspiwack: Agreed.

src/Data/Array/Destination.hs
```diff
@@ -2,9 +2,95 @@
...
+-- import qualified Prelude
+--
+-- apbxc :: Vector Int -> Vector Int -> Int -> Vector Int
+-- apbxc a b c = mult c Prelude.$ Vector.zipWith (Prelude.+) a b
+--
+-- mult :: Int -> Vector Int -> Vector Int
+-- mult c = Vector.map (Prelude.* c)
+-- @
+--
+-- Doing this with destination arrays we are guarenteed only one allocation
```
monoidal, line 53:
```suggestion
-- Doing this with destination arrays we are guaranteed only one allocation
```

src/Data/Array/Destination.hs
```diff
@@ -2,9 +2,95 @@
...
+--  apbxcDest :: Vector Int -> Vector Int -> Int -> Vector Int
+--  apbxcDest a b c = alloc (Vector.length a) writeResult
+--    where
+--      writeResult :: DArray Int #-> ()
+--      writeResult = fromFunction (\i -> ((a ! i) + (b ! i)) * c)
+-- @
+--
+-- == Aside: Why do we need linear types?
+--
+-- Linear types avoids ambigious writes to the destination array.
```
monoidal, line 66:
```suggestion
-- Linear types avoids ambiguous writes to the destination array.
```

PR #242, opened 2020-10-01
Consistent parameter ordering of linear collections

### LINKED ISSUE #147: Parameter order of linear collections
Currently, most of our linear collections (`Array`, `Vector.Mutable` and `HashMap`) take the collection as a first argument. Here are some examples from linear base:

```haskell
Data.Array.Mutable.Linear.write :: HasCallStack => Array a #-> Int -> a -> Array a
Data.Vector.Mutable.Linear.write :: HasCallStack => Vector a #-> Int -> a -> Vector a
```

This is also what `vector` library does.

But there are also places where we do the opposite:

```haskell
Data.Array.Mutable.Linear.resize :: HasCallStack => Int -> Array a #-> Array a
Data.Array.Polarized.Pull.foldr :: (a #-> b #-> b) -> b #-> Array a #-> b
Data.Array.Destination.split :: Int -> DArray a #-> (DArray a, DArray a)
```

(Also, as expected, the functions we re-implemented from `base` also tend to take the collection as the last argument)

---

I think, taking the collection as the last parameter has some benefits:

* I believe most libraries are using the collection-last argument order.
* It makes point-free usage nicer (`vec & write 0 'a' & write 1 'c'`)
* Usually our functions are linear on the collection, and not on the rest. So, a type signature like:

```haskell
write :: Int -> a -> Array a #-> Array a
```

Makes it clear that "given an `Int` and an `a` that we're linearly transforming an Array to another". The alternative:

```haskell
write :: Array a #-> Int -> a -> Array a
```

In my opinion does not read as well, the unrestricted arrow before the result gives me the wrong impression.

The disadvantages I can think of:

* Differing from the `vector` library.
* We have to change some amount of code.

---

No matter which order we choose, it would be good to stick to it across the codebase.

This issue is broken up from the original PR: https://github.com/tweag/linear-base/pull/146

utdemir:
Closes #147 .

This PR updates the mutable collections API slightly to have consistency on order of parameters. The issue #147 has more reasoning behind the changes, but in short:

* The pure-looking names take the collection as the last parameter (`set`, `get`, `lookup`)
* Imperative-looking names take the collection as the first parameter (`read`, `write`)

For arrays and vectors, we both `set` and `get` methods alongside with the flipped versions as `read` and `write` (and their unsafe variants). However, unlifted arrays, maps and sets only have an entirely pure-looking interface. The only reason is that I did not find much value duplicating the entire API and making up new names for imperative-looking counterparts.

Let me know if you have another preference.

### DIFF
--- a/docs/DESIGN.md
+++ b/docs/DESIGN.md
@@ -77,10 +77,10 @@ We have established the following conventions in this project:
   instance, import `Data.Functor.Linear` as `Linear` and not as `F`
   for functor.
 * All public modules have an export list.
-* For things which is meant te be used in pure contexts, take the
-thing as the last parameter (similar to `Data.Map`). If it is meant
-to be used in monadic contexts, take the thing as the first parameter
-(eg. `Control.Concurrent.MVar`). See [issue #147][issue-147] for some
+* Pure functions which modify a container take the
+container as the last parameter (similar to functions in `Data.Map`). Monadic functions on containers
+take the containers as the first parameter
+(similar to functions in `Control.Concurrent.MVar`). See [issue #147][issue-147] for some
 more details.
 
 [functors]: https://www.tweag.io/posts/2020-01-16-data-vs-control.html
--- a/docs/DESIGN.md
+++ b/docs/DESIGN.md
@@ -77,6 +77,11 @@ We have established the following conventions in this project:
   instance, import `Data.Functor.Linear` as `Linear` and not as `F`
   for functor.
 * All public modules have an export list.
+* For things which is meant te be used in pure contexts, take the
+thing as the last parameter (similar to `Data.Map`). If it is meant
+to be used in monadic contexts, take the thing as the first parameter
+(eg. `Control.Concurrent.MVar`). See [issue #147][issue-147] for some
+more details.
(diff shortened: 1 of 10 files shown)


docs/DESIGN.md
```diff
@@ -77,6 +77,11 @@ We have established the following conventions in this project:
   instance, import `Data.Functor.Linear` as `Linear` and not as `F`
   for functor.
 * All public modules have an export list.
+* For things which is meant te be used in pure contexts, take the
+thing as the last parameter (similar to `Data.Map`). If it is meant
+to be used in monadic contexts, take the thing as the first parameter
+(eg. `Control.Concurrent.MVar`). See [issue #147][issue-147] for some
+more details.
```
aspiwack, line 84: “things” is not super clear.

```suggestion
* Pure functions which modify a container take the
container as the last parameter (similar to functions in `Data.Map`). Monadic functions on containers
take the containers as the first parameter
(similar to functions in `Control.Concurrent.MVar`). See [issue #147][issue-147] for some
more details.
```

aspiwack (approved):

PR #243, opened 2020-10-01
Add call stacks to partial functions of destination arrays

### LINKED ISSUE #239: Destination array: mirror should have `HasCallStack`
The `mirror` function in `Data.Array.Destination` is partial (due to its calling `Vector.(!)`). It should have a `HasCallStack` constraint.

There is a tiny bit of work to do there, since the error is never raised by `mirror` itself, but by `(!)` which doesn't have a `HasCallStack` constraint. So we need to wrap `(!)` with an `HasCallStack` constraint one way or the other. If we want the errors to be located.

utdemir:
Closes #239.

This PR adds a `HasCallStack` constraints to `fill` and `mirror` functions of destination arrays. 

Instead of having variant of `Vector.!` with `HasCallStack` (which'd probably require bounds checking), we do an initial check and panic if the given vector is smaller than the destination array.

I also added a `size` function of `DArray`'s, and made it a part of the public interface; because I did not see a reason not to.

### DIFF
--- a/src/Data/Array/Destination.hs
+++ b/src/Data/Array/Destination.hs
@@ -116,6 +116,7 @@ module Data.Array.Destination
     DArray
   -- * Create and use a @DArray@
   , alloc
+  , size
   -- * Ways to write to a @DArray@
   , replicate
   , split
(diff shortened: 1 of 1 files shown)


aspiwack (approved):
This was not urgent, but fair enough :slightly_smiling_face:

Issue #197, opened 2020-09-18
Is there a notion of lens-to-something-unrestricted?


aspiwack:
Consider:

```haskell
data T a b where
  MkT :: a #-> b -> MkT a b
```

(what matters here is that the second field is unrestricted)

If I have a lens `l2` into the second field, then I can actually `set` the value at `l2` linearly (currently, `set l2 :: b' -> T a b -> T b'`, no linearity anywhere).

In #79 , there was

```haskell
set'' :: Optic_ (NonLinear.Kleisli (Control.Reader b)) a b s t -> b #-> s -> t
set'' (Optical l) b s = Control.runReader (NonLinear.runKleisli (l (NonLinear.Kleisli (const (Control.reader id)))) s) b
```

to this specific effect.

### Unrestricted lenses

But what is the type of `l2`? Certainly it's not going to be `Optic_ (NonLinear.Kleisli (Control.Reader b))`, it doesn't compose with anything. A reasonable requirement for the type of `l2` is that it is a super type of `Lens`.

We can probably work this out. In a sense `T` is the prototypical example. An unrestricted-field-lens should define an isomorphism between a type, and `T x b` for some `x`. We basically need a variant of

```haskell
second :: a `arr` b -> (c,a) `arr` (c, b)
```

Replacing `(,)` with `T` (or something equivalent).

Something to this effect anyway. Maybe we can see it as two properties, one relating to the pair, and one to the unrestrictedness. So that we could deal with unrestricted fields in prisms and traversals as well.

### Replace `set`?

What I'm sort of wondering too, is whether this `set''` could subsume `set` some way.  Probably not, to be honest. Though it's a bit sad to have two different `set` functions which do the same thing, just in slightly different circumstances.

utdemir:
+1, it seems like this situation is becoming pretty common. I had this issue before while writing lenses for mutable collections, where the collection itself is linear but the elements inside are unrestricted. So I had ended up writing something like:

```haskell
lensAt :: Int -> Lens' (Unrestricted a) (Rec b)
```

aspiwack:
I'm on this.

I know how to fix. But, of course, it duplicates every Optics type. In my head, I'm suffixing them with a `U`, so we would have

```haskell
type IsoU
type IsoU'
type LensU
type LensU'
type PrismU
type PrismU'
type TraversalU
type TraversalU'
```

Do we like this? (cc @mboes , if you have an opinion)

aspiwack:
[...]

PR #244, opened 2020-10-01
Add an alterF function to the HashMap module

aspiwack:
If anything, `alterF` is part of the optics story (it's a lens in the
style of #197). But it's also generally useful, if rarely so.

Also exposes `alter` (which has been mysteriously forgotten) and
`alterF` in the export list.

Depends on #147


utdemir (approved):

utdemir:
(Later on we can do the same on vectors and arrays `modify` functions too, but this PR is good to go)

PR #245, opened 2020-10-04
Update GHC and use new linear arrow syntax

utdemir:
This PR updates GHC to a newer revision, which comes with the new linear arrow syntax (`%p ->`).

* Uses https://github.com/tweag/nixpkgs/commit/60a06c5a2ec88392c1eb5e252367dde4e0ead16c to update GHC. I haven't sent a PR upstream because it has an annoying locale fix which probably won't work on all GHC variants (I'm thinking about `musl`); and I couldn't find time to test them.
* I replaced most uses of `#->` with crude search & replace (`sed -i 's/#->/%1->/g' {src,test,examples}/**/*.hs`), so it's possible that it missed something.
* There is one behaviour change that I don't know the reason of; it seems like now the pattern guards does not bind linearly, so I had to apply this kind of change on a few places:

```haskell
instance (Prelude.Eq a, Movable a) => Eq (MovableEq a) where
  MovableEq ar == MovableEq br
-    | Ur a <- move ar , Ur b <- move br
-    = a Prelude.== b
+    = move (ar, br) & \(Ur (a, b)) ->
+        a Prelude.== b
```

The most important one is that, I had to disable our doctests. I added a comment explaining why to `linear-base.cabal` , quoted below:

> -- cabal-install has a piece of code[1] which injects a Cabal upper bound to
-- packages with custom Setup.hs's. And this happens after the overrides,
-- so the usual mechanisms of overriding upper bounds does not work.
--
-- GHC 9 comes with Cabal 3.4, which is above that bound. So, when using
-- GHC 9 with cabal-install 3.2; `build-type: Custom` causes another Cabal
-- library to be built, and that causes a strange type error ("expecting IO,
-- but got IO"), which I suspect because it conflicts with the existing boot
-- packages.
--
-- [1]: https://github.com/haskell/cabal/blob/d28c80acc69b9e7fa992a0b2b7fced937734b238/cabal-install/src/Distribution/Client/ProjectPlanning.hs#L1132-L1149

I don't like it, but it is the easiest workaround until GHC 9 is released hopefully within a few weeks.

### DIFF
--- a/src/Prelude/Linear.hs
+++ b/src/Prelude/Linear.hs
@@ -145,7 +145,6 @@ import Data.Ord.Linear
 import Data.Tuple.Linear
 import Data.List.Linear
 import qualified Prelude
-import GHC.Exts (FUN)
 import Prelude.Linear.Internal
 import Data.Eq.Linear
 import Data.String
--- a/src/Prelude/Linear.hs
+++ b/src/Prelude/Linear.hs
@@ -11,13 +11,13 @@
 -- >>> :set -XNoImplicitPrelude
 -- >>> import Prelude.Linear
 -- >>> :{
---   boolToInt :: Bool #-> Int
+--   boolToInt :: Bool %1-> Int
 --   boolToInt False = 0
 --   boolToInt True = 1
 -- :}
 --
 -- >>> :{
---   makeInt :: Either Int Bool #-> Int
+--   makeInt :: Either Int Bool %1-> Int
 --   makeInt = either id boolToInt
 -- :}
 --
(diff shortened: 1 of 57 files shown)


aspiwack:
The reasons why: https://gitlab.haskell.org/ghc/ghc/-/issues/18439

Thanks for this PR. I'm fine with the temporary workarounds.

src/Prelude/Linear.hs
```diff
@@ -145,15 +145,14 @@ import Data.Ord.Linear
...
 import Data.List.Linear
 import qualified Prelude
-import GHC.Exts (FUN)
 import Prelude.Linear.Internal
 import Data.Eq.Linear
 import Data.String
 
 -- | Replacement for the flip function with generalized multiplicities.
-flip :: FUN r (FUN p a (FUN q b c)) (FUN q b (FUN p a c))
+flip :: (a %p -> b %q -> c) %r -> b %q -> a %p -> c
```
aspiwack, line 153: Aha! Nice :slightly_smiling_face:

aspiwack (approved):
Merge asap. Apparently there are already conflicts.

Issue #254, opened 2020-10-12
Update readme post release


Divesh-Otwani:
The "Getting started" section should mention the release on hackage or maybe stackage. We should also mention the haddock documentation.

utdemir:
I think there's a link now to the Hackage page which includes the haddocks, so I am closing this issue.

When Stackage releases a resolver with GHC 9 we can mention that too.

PR #256, opened 2020-10-12
Mention haddock in readme

Divesh-Otwani:


aspiwack (commented):
This is a bit pointless without a link to something. Like the Hackage package. In that these word don't add any information.

Divesh-Otwani:
Ok, I'll do it when we get to #254.

PR #258, opened 2020-10-12
Made toLinear parametric in input functions multiplicity

Divesh-Otwani:
Sometimes it's annoying to use `Unsafe.toLinear` or variants if a function is linear on some later argument and currying is tricky. This just makes that easier:

Before:
```haskell
toLinear2 :: (a -> b -> c) %1-> (a %1-> b %1-> c)
```

After:
```haskell
toLinear2 :: (a %p-> b %q-> c) %1-> (a %1-> b %1-> c)
```


aspiwack (approved):
Fair enough.

PR #259, opened 2020-10-12
minor whitespace edits

Divesh-Otwani:


utdemir (approved):
Great!

While we're at it, can you also remove the trailing whitespaces on various files? This command should suffice:

```
sed -i 's/[ \t]*$//' {src,test,examples}/**/*.hs
```

PR #261, opened 2020-10-12
Created a Data.Semigroup.Linear module

Divesh-Otwani:
I put the `Data.Semigroup` stuff into a separate module. I had to allow orphan instances since `Endo` would be defined in the semigroup module. I hope that's not a dealbreaker.


utdemir (approved):
I don't see the orphan instance as a problem, because in this case there is no other sensible implementation.

Although, I personally wouldn't split the files because they're conceptually vey similar; but the change LGTM.

PR #262, opened 2020-10-14
quicksort example

Divesh-Otwani:
The quicksort example used in the blog post. Will update the blog post after this is in.

### DIFF
--- a/examples/Simple/Quicksort.hs
+++ b/examples/Simple/Quicksort.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 
+-- | This module implements quicksort with mutable arrays from linear-base
 module Simple.Quicksort
   ( testQuicksort
   , quickSort
@@ -40,6 +41,12 @@ go lo hi arr
         \arr3 -> go lo (ix-1) arr3 &
           \arr4 -> go (ix+1) hi arr4
 
+-- | @partition arr pivot lo hi = (arr', Ur ix)@ such that
+-- @arr'[i] <= pivot@ for @lo <= i <= ix@,
+-- @arr'[j] > pivot@ for @ix < j <= hi@,
+-- @arr'[k] = arr[k]@ for @k < lo@ and @k > hi@, and
+-- @arr'@ is a permutation of @arr@ (which means basically only the slice
+-- from indicies @lo@ to @hi@ are permuted).
 partition :: Array Int %1-> Int -> Int -> Int -> (Array Int, Ur Int)
 partition arr pivot lx rx
   | (rx < lx) = (arr, Ur (lx-1))
@@ -52,6 +59,7 @@ partition arr pivot lx rx
           (False, False) -> swap arr2 lx rx &
             \arr3 -> partition arr3 pivot (lx+1) (rx-1)
 
+-- | @swap a i j@ exchanges the positions of values at @i@ and @j@ of @a@.
 swap :: HasCallStack => Array Int %1-> Int -> Int -> Array Int
 swap arr i j = Array.read arr i &
   \(Ur ival, arr1) -> Array.read arr1 j &
--- a/examples/Simple/Quicksort.hs
+++ b/examples/Simple/Quicksort.hs
@@ -45,8 +45,7 @@ go lo hi arr
 -- @arr'[i] <= pivot@ for @lo <= i <= ix@,
 -- @arr'[j] > pivot@ for @ix < j <= hi@,
 -- @arr'[k] = arr[k]@ for @k < lo@ and @k > hi@, and
--- @arr'@ is a permutation of @arr@ (which means basically only the slice
--- from indicies @lo@ to @hi@ are permuted).
+-- @arr'@ is a permutation of @arr@.
 partition :: Array Int %1-> Int -> Int -> Int -> (Array Int, Ur Int)
 partition arr pivot lx rx
   | (rx < lx) = (arr, Ur (lx-1))
--- a/examples/Simple/Quicksort.hs
+++ b/examples/Simple/Quicksort.hs
@@ -1,0 +1,58 @@
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Simple.Quicksort
+  ( testQuicksort
+  , quickSort
+  )
+where
+
+import GHC.Stack
+import qualified Data.Array.Mutable.Linear as Array
+import Data.Array.Mutable.Linear (Array)
+import Data.Unrestricted.Linear
+import Prelude.Linear hiding (partition)
+
+
+-- # Testing
+-------------------------------------------------------------------------------
+
+testQuicksort :: [Int] -> Bool
+testQuicksort xs = sort xs == quickSort xs
+
+
+-- # Quicksort
+-------------------------------------------------------------------------------
+
+quickSort :: [Int] -> [Int]
+quickSort xs = unur $ Array.fromList xs $ Array.toList . arrQuicksort
+
+arrQuicksort :: Array Int %1-> Array Int
+arrQuicksort arr = Array.size arr &
+  \(Ur len, arr1) -> go 0 (len-1) arr1
+
+go :: Int -> Int -> Array Int %1-> Array Int
+go lo hi arr
+  | lo >= hi = arr
+  | otherwise = Array.read arr lo &
+    \(Ur pivot, arr1) -> partition arr1 pivot lo hi &
+      \(arr2, Ur ix) -> swap arr2 lo ix &
+        \arr3 -> go lo (ix-1) arr3 &
+          \arr4 -> go (ix+1) hi arr4
+
+partition :: Array Int %1-> Int -> Int -> Int -> (Array Int, Ur Int)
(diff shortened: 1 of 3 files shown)


examples/Simple/Quicksort.hs
```diff
@@ -0,0 +1,61 @@
...
+        (Ur rVal, arr2) -> case (lVal <= pivot, pivot < rVal) of
+          (True, True) -> partition arr2 pivot (lx+1) (rx-1)
+          (True, False) -> partition arr2 pivot (lx+1) rx
+          (False, True) -> partition arr2 pivot lx (rx-1)
+          (False, False) -> swap arr2 lx rx & \case
+            arr3 -> partition arr3 pivot (lx+1) (rx-1)
+
+swap :: HasCallStack => Array Int %1-> Int -> Int -> Array Int
+swap arr i j = Array.read arr i & \case
+  (Ur ival, arr1) -> Array.read arr1 j & \case
```
utdemir, line 60: `LambdaCase`'s on this file seems unnecessary.

examples/Simple/Quicksort.hs
```diff
@@ -0,0 +1,61 @@
...
+quickSort xs = unur $ Array.fromList xs $ Array.toList . arrQuicksort
+
+arrQuicksort :: Array Int %1-> Array Int
+arrQuicksort arr = Array.size arr & \case
+  (Ur len, arr') -> go 0 (len-1) arr'
+
+go :: Int -> Int -> Array Int %1-> Array Int
+go lo hi arr = case lo >= hi of
+  True -> arr
+  False -> Array.read arr lo & \case
```
utdemir, line 40: Using guards here (or at least if-then-else) should be nicer:

```suggestion
go lo hi arr
  | lo >= hi = arr
  | otherwise = 
    Array.read arr lo & \case
```

examples/Simple/Quicksort.hs
```diff
@@ -0,0 +1,61 @@
...
+partition arr pivot lx rx
+  | (rx < lx) = (arr, Ur (lx-1))
+  | otherwise = Array.read arr lx & \case
+      (Ur lVal, arr1) -> Array.read arr1 rx & \case
+        (Ur rVal, arr2) -> case (lVal <= pivot, pivot < rVal) of
+          (True, True) -> partition arr2 pivot (lx+1) (rx-1)
+          (True, False) -> partition arr2 pivot (lx+1) rx
+          (False, True) -> partition arr2 pivot lx (rx-1)
+          (False, False) -> swap arr2 lx rx & \case
+            arr3 -> partition arr3 pivot (lx+1) (rx-1)
```
utdemir, line 56: Same here, would look like:

```haskell
swap lx rx arr2 
  & partition pivot (lx+1) (rx-1)
```

examples/Simple/Quicksort.hs
```diff
@@ -0,0 +1,61 @@
...
+
+swap :: HasCallStack => Array Int %1-> Int -> Int -> Array Int
+swap arr i j = Array.read arr i & \case
+  (Ur ival, arr1) -> Array.read arr1 j & \case
+    (Ur jval, arr2) -> Array.write (Array.write arr2 i jval) j ival
```
utdemir, line 61: This is optional, but inlining function calls like this makes the code less readable to me, I'd prefer something like:

```haskell
arr
  & Array.set i jval
  & Array.set j ival
```

examples/Simple/Quicksort.hs
```diff
@@ -0,0 +1,61 @@
...
+
+-- # Quicksort
+-------------------------------------------------------------------------------
+
+quickSort :: [Int] -> [Int]
+quickSort xs = unur $ Array.fromList xs $ Array.toList . arrQuicksort
+
+arrQuicksort :: Array Int %1-> Array Int
+arrQuicksort arr = Array.size arr & \case
+  (Ur len, arr') -> go 0 (len-1) arr'
```
utdemir, line 35: Some places we call this `arr'`, some other we call this `arr1`. We should set a precedent somewhere.

utdemir (changes requested):
I copied a bunch of my stylistic comments from the original PR to here.

All of them are non-critical style improvements, but I think there is value keeping the example pretty, since it might be the first time someone seeing linear code.

examples/Simple/Quicksort.hs
```diff
@@ -0,0 +1,61 @@
...
+
+go :: Int -> Int -> Array Int %1-> Array Int
+go lo hi arr = case lo >= hi of
+  True -> arr
+  False -> Array.read arr lo & \case
+    (Ur pivot, arr0) -> partition arr0 pivot lo hi & \case
+      (arr1, Ur ix) -> swap arr1 lo ix & \case
+        arr2 -> go lo (ix-1) arr2 & \case
+          arr3 -> go (ix+1) hi arr3
```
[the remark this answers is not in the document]
Divesh-Otwani: I think that's less readable than a lambda that names the next state with `arr0`, `arr1`, etc.

examples/Simple/Quicksort.hs
```diff
@@ -0,0 +1,61 @@
...
+-------------------------------------------------------------------------------
+
+quickSort :: [Int] -> [Int]
+quickSort xs = unur $ Array.fromList xs $ Array.toList . arrQuicksort
+
+arrQuicksort :: Array Int %1-> Array Int
+arrQuicksort arr = Array.size arr & \case
+  (Ur len, arr') -> go 0 (len-1) arr'
+
+go :: Int -> Int -> Array Int %1-> Array Int
```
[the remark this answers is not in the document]
Divesh-Otwani: It does, but it's written this way to match the blog post.

examples/Simple/Quicksort.hs
Divesh-Otwani: I just composed the functions.

examples/Simple/Quicksort.hs
Divesh-Otwani: Yes, true. I will keep the lambda notation instead of just using currying. I think it's more readable as we pass the result of the left of the `&` and bind it to the argument of the lambda, which is the `x` in `\x -> ...`.

examples/Simple/Quicksort.hs
```diff
@@ -0,0 +1,58 @@
...
+go lo hi arr
+  | lo >= hi = arr
+  | otherwise = Array.read arr lo &
+    \(Ur pivot, arr1) -> partition arr1 pivot lo hi &
+      \(arr2, Ur ix) -> swap arr2 lo ix &
+        \arr3 -> go lo (ix-1) arr3 &
+          \arr4 -> go (ix+1) hi arr4
+
+partition :: Array Int %1-> Int -> Int -> Int -> (Array Int, Ur Int)
```
facundominguez, line 43:
```suggestion
-- | @partition a pivot i j@ permutes the values of @a@ in the range @[i, j]@.
-- The values greater than the pivot come last, preceded by the remaining values.
--
-- More formally, when @partition a pivot i j == (p, s)@, for all @k@ such that
-- @i <= k <= j@:
-- If @k <= s@ then @p[k] <= pivot@, and
-- if @s < k@ then @pivot < p[k]@
partition :: Array Int %1-> Int -> Int -> Int -> (Array Int, Ur Int)
```

I'm finding partition a mouthful to explain. Improving it would maybe require introducing a notion of slice.
```
partition :: Slice Int -> (Slice Int, Slice Int)
```

examples/Simple/Quicksort.hs
```diff
@@ -0,0 +1,58 @@
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Simple.Quicksort
```
facundominguez, line 4: I'm missing some context. If this example is discussed in a post, a link to the post would be in order in the module documentation.

Divesh-Otwani: The post isn't out yet. I added some doc at the top.

examples/Simple/Quicksort.hs
Divesh-Otwani: I think I put down a reasonable explanation. Quicksort is tricky and we kind of have to expect some familiarity with the algorithm.

examples/Simple/Quicksort.hs
```diff
@@ -0,0 +1,66 @@
...
+      \(arr2, Ur ix) -> swap arr2 lo ix &
+        \arr3 -> go lo (ix-1) arr3 &
+          \arr4 -> go (ix+1) hi arr4
+
+-- | @partition arr pivot lo hi = (arr', Ur ix)@ such that
+-- @arr'[i] <= pivot@ for @lo <= i <= ix@,
+-- @arr'[j] > pivot@ for @ix < j <= hi@,
+-- @arr'[k] = arr[k]@ for @k < lo@ and @k > hi@, and
+-- @arr'@ is a permutation of @arr@ (which means basically only the slice
+-- from indicies @lo@ to @hi@ are permuted).
```
aspiwack, line 49:
```suggestion
-- @arr'@ is a permutation of @arr@.
```

aspiwack (approved):

aspiwack:
I really don't know why the remaining “change requested” is blocking the merge. It is not a rule (the only rule I put was: at least one approval, which the PR has).

I'm using my admin privilege and merging.

Issue #93, opened 2020-04-17
Elegance of Tests


Divesh-Otwani:
- Make the HashMap tests as elegant as the Data.Array.Mutable.Linear tests. That is, use `===` from hedgehog and have the allocation functions take `HashMap k v #-> Unrestricted (TestT IO ())`. Also, have as systematic an approach to testing. (There are currently too few tests.)
- Rename the list generators non_empty_list and so on. Make sure that anything that can be tested with empty lists is. In general, make sure the domain, the X in "forall a in X. axiom(a)" is fully covered.

PR #263, opened 2020-11-06
Testing set intersect and union

Divesh-Otwani:
Working towards closing #93 I've added tests for intersections and unions for sets.

[Remark: We should probably add tests for `empty` and maybe I should just do that in the PR but it looks obviously correct to me. (Saying this, I feel like I'm going to end up having to add it.)]

Issue #93 explains that the mutable data test suite is meant to provide properties which serve as axioms that define the behavior of how to interact with the data structure. That is, if all the properties hold, the behavior of the data structure is mathematically specified. For each accessor function `f` and each modifier or constructor `g` we have one axiom to specify the behavior of `f (g (...)) = ...`. See for instance [this](https://softwarefoundations.cis.upenn.edu/vfa-current/ADT.html) link. 

In this PR, since we added two modifiers (intersection and union) and have two basic accessors (member and size), we need 2 X 2 = 4 axioms, which are precisely the 4 tests I've added.

### DIFF
--- a/test/Test/Data/Mutable/Set.hs
+++ b/test/Test/Data/Mutable/Set.hs
@@ -7,8 +7,8 @@
 --
 -- = How we designed the tests
 --
--- We use hedgehog to test properties that are axioms that fully define the
--- behavior of sets. There is at least one axiom per each combination of
+-- We use hedgehog to test properties that are axioms that funtionally define
+-- the behavior of sets. There is at least one axiom per each combination of
 -- accessor and constructor/mutator. So for the accessor @f@, and
 -- constructor/mutator @g@, we have one axiom of the form @f (g (...)) = ...@.
 --
@@ -20,10 +20,22 @@
 -- the union as sets, converting back with @Set.toList@ and then sorting
 -- the output list.
 --
--- To have such a test replace an axiom, however, we need to have "homomorphisms"
--- for each accessor. E.g., for all x and l, @member x (fromList l) = elem x l@.
+-- To have such a test replace an axiom, however, we need to have
+-- "homomorphisms" for each accessor or modifier. In general, we want to be
+-- able to say for any modifier or accessor @f@,
 --
--- So now, for instance, we can prove
+-- >  toList (f set) = f' (toList set)
+--
+-- where @f'@ is some reference implementation we know. The key idea is a kind
+-- of "homomorphism" between @f@ and a reference implementation @f'@ that
+-- holds across the conversion function @toList@.
+--
+-- Note: We could also formulate this in terms of @fromList@.
+--
+-- For example, we'd want to have @member x (fromList l) = elem x l@
+-- for any @x@ and @l@.
+--
+-- With this we can prove
 --
 -- >  member x (intersect set1 set2) = member x set1 && member x set2
 --
@@ -77,14 +89,14 @@ group =
   , testProperty "∀ s, x \\in s. size (delete s x) = size s - 1" sizeDelete1
   , testProperty "∀ s, x \\notin s. size (delete s x) = size s" sizeDelete2
   -- Homomorphism tests
-  , testProperty "sort . nub = sort . toList . fromList" toListFromList
-  , testProperty "member x (fromList l) = elem x l" memberHomomorphism
-  , testProperty "size . fromList = length . nub" sizeHomomorphism
+  , testProperty "sort . nub = sort . toList" toListFromList
+  , testProperty "member x s = elem x (toList s)" memberHomomorphism
+  , testProperty "size = length . toList" sizeHomomorphism
   , testProperty
-      "sort . nub (l ∪ m) = sort . toList (fromList l ∪ fromList m)"
+      "sort . nub ((toList s) ∪ (toList s')) = sort . toList (s ∪ s')"
       unionHomomorphism
   , testProperty
-      "sort . nub (l ∩ m) = sort . toList (fromList l ∩ fromList m)"
+      "sort . nub ((toList s) ∩ (toList s')) = sort . toList (s ∩ s')"
       intersectHomomorphism
   ]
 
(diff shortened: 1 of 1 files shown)


Divesh-Otwani:
Also, no idea why the CI is failing since `nix-shell --pure --run 'cabal update; cabal test'` worked on my machine.

facundominguez:
Retrying the CI job succeeds. Maybe there is a full disk in the machine that ran the first attempt.

test/Test/Data/Mutable/Set.hs
```diff
@@ -129,9 +141,39 @@ memberDelete2Test val1 val2 set = fromRead (Set.member val2 set)
...
+  v <- forAll value
+  let valInBoth = elem v l && elem v l'
+  test $ unur Linear.$
+    Set.fromList l
+      (\s -> Set.fromList l'
+        (\s' -> memberIntersectTest v valInBoth (Set.intersection s s')))
+
+memberIntersectTest :: Int -> Bool -> Set.Set Int %1-> Ur (TestT IO ())
+memberIntersectTest val result inter = Set.member val inter Linear.&
+  \(Ur b, s') -> lseq s' (Ur (b === result))
```
aspiwack, line 172: This test is correct. But I contend that it doesn't test much, as I assume that, almost always, `v` will not be a member of either `l` or `l'` (let alone both). Therefore, this test will almost always be testing that `False === False`.

Instead, I'd suggest writing a `toList` function of sort, and testing that the output list is the expected one. Individual tests will be a bit more expensive, but much more informative.

test/Test/Data/Mutable/Set.hs
```diff
@@ -192,3 +234,31 @@ sizeDelete2Test val set = fromRead (Set.size set)
...
+
+sizeIntersect :: Property
+sizeIntersect = property $ do
+  l <- forAll list
+  l' <- forAll list
+  let isectListLen = List.length $ List.nub $ List.intersect l l'
+  let f :: Set.Set Int %1-> Ur Int
+      f = \s -> Set.fromList l' (\s' -> dropSnd (Set.size (Set.intersection s s')))
+  let isectInSetsLen = unur Linear.$ Set.fromList l f
+  test (isectInSetsLen === isectListLen)
```
aspiwack, line 256: Same comment.

test/Test/Data/Mutable/Set.hs
Divesh-Otwani: Ah, so pseudo-homomorphism tests. It's not trivial to see, but these are sufficient to replace defining axioms. Done below.

test/Test/Data/Mutable/Set.hs
```diff
@@ -192,3 +221,38 @@ sizeDelete2Test val set = fromRead (Set.size set)
...
+toListFromList = property $ do
+  l <- forAll list
+  let outsideSet = List.nub . List.sort $ l
+  List.sort (unur (Set.fromList l Set.toList)) === outsideSet
+
+unionHomomorphism :: Property
+unionHomomorphism = property $ do
+  l <- forAll list
+  l' <- forAll list
+  let listUnion = List.nub $ List.sort $ List.union l l'
```
utdemir, line 235: `nub` is pretty inefficient (`O(n^2)`), but we should be able to replace it with `Data.Containers.ListUtils.nubOrd` which runs in `O(nlogn)`. It doesn't change much, but will probably make the tests a bit faster.

Divesh-Otwani: Good idea :+1:

Divesh-Otwani:
I cleaned up those tests and convinced myself of a way to still test the axioms if we are using "homomorphism" tests against reference list implementations.

test/Test/Data/Mutable/Set.hs
```diff
@@ -2,18 +2,56 @@
...
+-- If however, this is cumbersome, we can often just test against an existing
+-- implementation. This is like a homomorphism test (but it's not quite a
+-- homomorphism). For unions, this is like saying if we have two lists A and B,
+-- and we take their union as lists, then sort and nub (remove duplicates)
+-- this is the same as using @Set.fromList@ to make them into sets, taking
+-- the union as sets, converting back with @Set.toList@ and then sorting
+-- the output list.
+--
+-- To have such a test replace an axiom, however, we need to have "homomorphisms"
+-- for each accessor. E.g., for all x and l, @member x (fromList l) = elem x l@.
```
aspiwack, line 24: Like in #265: use `toList`, rather than `fromList`.

And probably phrase this in terms of semantics, rather than homomorphism.

You can spare yourself the cumbersomeness of using lists in your semantics if you use `Set` instead.

Divesh-Otwani: I like homomorphisms better though :cry: It's like saying the semantics of operations on sets can be expressed in terms of the semantics of operations on lists because they can carry through the `fromList`. I'll add a comment stating this though.

Divesh-Otwani: Also, I want to use lists in my semantics because we don't have a `fromSet` constructor and it won't correspond as well. (Though we easily could have such a constructor.)

aspiwack (approved):
I still think this homorphism thing is a bit misleading.

The semantics of a `Set` is, well, a (finite) set. And the semantic of `union` on `Set` is the union on the semantics. Then we get to claim that the semantics is an homomorphism for the `union` structure, which, technically, is true. But we are not testing the semantics, we are testing the union.

That being said, there is a way in which we can legitimately speak of homomorphism. Let's imagine that we have an abstract data type. It has a semantics and all. But, in this view, the semantics is not executable. However we use `[]` (or `Data.Set`) as a _reference implementation_. Then, we establish correctness of our implementation of interest by exhibiting a (semantic-preserving) homomorphism from our structure to the reference implementation.

PR #265, opened 2020-11-11
Axiomatic tests/vec arr

Divesh-Otwani:
In the style of #263 these are the remaining tests we need to fully specify mutable arrays and vectors.

### DIFF
--- a/test/Test/Data/Mutable/Array.hs
+++ b/test/Test/Data/Mutable/Array.hs
@@ -6,31 +6,10 @@
 -- |
 -- Tests for mutable arrays.
 --
--- The API is basically 4 mutators or constructors: alloc, write, resize,
--- resizeSeed and 2 accessors: length, read.
+-- See the testing framework explained in Test.Data.Mutable.Set.
 --
--- We test arrays behave properly with properties that axiomatically define
--- arrays.  Basically, using these properties, the behavior of any use of
--- arrays should be fully specified without having to look at the
--- implementation.
---
--- An example of such a property would be the following:
---
--- for all i, j, arr, s.t. i /= j and 0 <= i,j < (length arr),
---             read (write arr j x) i == read arr i
---
--- In general, most uses of arrays are of the form
--- \[  accessor (constructor (...)) \].
---
--- Hence, we define properties that allow you to "simplyfy" or "rewrite"
--- any such combination into a simpler form. (More formally,
--- we're defining term re-write rules.) For now, constructor-constructor
--- re-writes are skipped (e.g., write (write a i x) i y = write a i y).
---
--- TODO:
---  * Test failures for out of bound access
---  * Constructor - constructor rules, like
---                write (write a i x) i y = write a i y
+-- Remarks:
+--  * We don't test for failure on out-of-bound access (it's trivial)
 module Test.Data.Mutable.Array
   ( mutArrTests,
   )
--- a/test/Test/Data/Mutable/Array.hs
+++ b/test/Test/Data/Mutable/Array.hs
@@ -8,6 +8,9 @@
 --
 -- See the testing framework explained in Test.Data.Mutable.Set.
 --
+-- The combiniation of axioms and homormorphisms provided fully specify
+-- the behavior of arrays.
+--
 -- Remarks:
 --  * We don't test for failure on out-of-bound access (it's trivial)
 module Test.Data.Mutable.Array
@@ -20,6 +23,7 @@ import Data.Unrestricted.Linear
 import qualified Data.Functor.Linear as Data
 import qualified Data.Ord.Linear as Linear
 import Hedgehog
+import qualified Data.List as List
 import qualified Hedgehog.Gen as Gen
 import qualified Hedgehog.Range as Range
 import qualified Prelude.Linear as Linear hiding ((>))
@@ -46,6 +50,9 @@ group =
   , testProperty "∀ a,i,x. len (write a i x) = len a" lenWrite
   , testProperty "∀ a,s,x. len (resize s x a) = s" lenResizeSeed
   -- Tests against a reference implementation
+  , testProperty "∀ a,ix. write a ix . fromList = fromList . write a ix" writeRef
+  , testProperty "∀ ix. read ix (fromList l) = l !! i" readRef
+  , testProperty "size . fromList = length" sizeRef
   , testProperty "∀ a,s,x. resize s x a = take s (toList a ++ repeat x)" resizeRef
   , testProperty "∀ s,n. slice s n = take s . drop n" sliceRef
   , testProperty "f <$> fromList xs == fromList (f <$> xs)" refFmap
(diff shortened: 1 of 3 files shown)


test/Test/Data/Mutable/Vector.hs
```diff
@@ -7,10 +7,10 @@
 -- |
 -- Tests for mutable vectors.
 --
--- TODO:
---  * Test failures for out of bound access
---  * Constructor - constructor rules, like
---                write (write v i x) i y = write v i y
+-- See the testing framework explained in Test.Data.Mutable.Set.
+--
+-- The combiniation of axioms and homormorphisms provided fully specify
```
utdemir, line 12: Typos on `combination` and `homomorphism`.

test/Test/Data/Mutable/Array.hs
```diff
@@ -6,31 +6,13 @@
...
 --
--- The API is basically 4 mutators or constructors: alloc, write, resize,
--- resizeSeed and 2 accessors: length, read.
+-- See the testing framework explained in Test.Data.Mutable.Set.
 --
--- We test arrays behave properly with properties that axiomatically define
--- arrays.  Basically, using these properties, the behavior of any use of
--- arrays should be fully specified without having to look at the
--- implementation.
+-- The combiniation of axioms and homormorphisms provided fully specify
```
utdemir, line 11: Same typos here.

utdemir, line 11: I find this sentence

> fully specify the behavior of arrays.

very ambitious. I'm sure we can find some operations outside `read` and `write` that we don't "fully" specify, I'm thinking about things like `freeze`. And whether something specifies it "fully" is also open, because then we should also test the out of bounds conditions etc. I'd honestly just remove the sentence.

Divesh-Otwani: I completely disagree. It's the whole point of #93. 

Right now the sentence means "fully specify" minus the remarks of things we skip. It's clear that that is true.

If you have a reference implementation and define homomorphisms for every constructor/modifer/accessor `f` of the form `f (fromList l) = fromList (f' l)` or `f (fromList l) = f' l`, then you can always move computations inside the `fromList` and have axioms that define everything.

Divesh-Otwani: The explanation in Test.Data.Mutable.Set in #263 covers this with an example proof.

test/Test/Data/Mutable/Array.hs
```diff
@@ -6,31 +6,14 @@
...
--- The API is basically 4 mutators or constructors: alloc, write, resize,
--- resizeSeed and 2 accessors: length, read.
+-- See the testing framework explained in Test.Data.Mutable.Set.
 --
--- We test arrays behave properly with properties that axiomatically define
--- arrays.  Basically, using these properties, the behavior of any use of
--- arrays should be fully specified without having to look at the
--- implementation.
+-- The combination of axioms and homomorphisms provided fully specify
+-- the behavior of arrays.
```
aspiwack, line 12: I think it's a bit of a misnomer to call these things homorphisms (there is a sense in which they are, but I don't think it's really a meaningful sense). What we are doing is testing operations with respect to their semantics (which happens to be given as a reference implementation).

test/Test/Data/Mutable/Array.hs
```diff
@@ -67,6 +51,9 @@ group =
   , testProperty "∀ a,i,x. len (write a i x) = len a" lenWrite
   , testProperty "∀ a,s,x. len (resize s x a) = s" lenResizeSeed
   -- Tests against a reference implementation
+  , testProperty "∀ a,ix. write a ix . fromList = fromList . write a ix" writeRef
+  , testProperty "∀ ix. read ix (fromList l) = l !! i" readRef
+  , testProperty "size . fromList = length" sizeRef
```
aspiwack, line 56: These properties are better specified in terms to `toList` than in term of `fromList`.

Despite the fact that we actually use `fromList` for the test, because we don't have a choice. Using `fromList` in the specification implies that arrays which are not of the form `fromList l` are exempt from the rule. There are, of course, non, or at least so we claim. But it should make it clear that `toList` is the more economical direction.

Another way to think about it (the right way if you are me), is that this is a property of arrays, so should quantify over the array, it is written in terms of the semantics of arrays, which is given by `toList`.

Just to make sure that there is no misunderstanding: I'm only asking you to rewrite the test names in the strings. The test themselves are fine.

test/Test/Data/Mutable/Vector.hs
```diff
@@ -48,6 +53,16 @@ group =
...
+  , testProperty "write ix a (fromList l) = fromList (update l ix a)" refWrite
+  , testProperty "fst $ modify f ix (fromList l) = snd $ f (l !! ix)" refModify1
+  , testProperty
+      "snd (modify f i (fromList l)) = fromList (update l i (fst (f (l !! i))))"
+      refModify2
+  , testProperty "push x . fromList = fromList . snoc x" refPush
+  , testProperty "pop . fromList = fromList . init" refPop
+  , testProperty "read ix (fromList l) = l !! ix" refRead
+  , testProperty "size . fromList = length" refSize
+  , testProperty "toList . shrinkToFit . fromList = id" refShrinkToFit
```
aspiwack, line 65: Same remark: reformulate in terms of `toList`.

test/Test/Data/Mutable/Vector.hs
```diff
@@ -220,6 +238,79 @@ lenPushTest val vec = fromLen Linear.$ Vector.size vec
...
+    listWrite _ _ [] = error "Index too big"
+    listWrite 0 a (_:xs) = a:xs
+    listWrite n a (x:xs) = x : listWrite (n-1) a xs
+
+refModify1 :: Property
+refModify1 = property $ do
+  l <- forAll nonEmptyList
+  let f x = (mod x 5, (mod x 5) Prelude.< 3)
+  ix <- forAll $ Gen.element [0..(length l - 1)]
+  let (_,result) = listApp ix f l
```
aspiwack, line 261: You should do as in your test specification and write `snd (f (l !! ix))`.

test/Test/Data/Mutable/Array.hs
aspiwack: “fully specifies” is a strong word. First, this is only functional specification. None of these properties specify anything about, say, the cost semantics of arrays. Second, of course, it does mean that we have a semantics for every single operation.

It's a rather good idea to tone it down. Something like “functionally specifies each operation”, maybe?

Or really, nothing at all, since you already point to the Set module which has less terse explanation.

test/Test/Data/Mutable/Array.hs
Divesh-Otwani: I just assumed that `fromList` was onto it's domain (which it is, though for vectors, it isn't). I'll re-write the test strings :+1:

test/Test/Data/Mutable/Array.hs
Divesh-Otwani: I think they are mostly like homomorphisms: we carry through an operation `f (fromList l) = fromList (f' l)` and have a correspondence between `f` and `f'` subject to the map `fromList`. As a result of this, because we have a semantics for `f'`, we specify `f` because any computation done with an `f` can be translated to one done with `f'`. This will be explained in the tests for `Set`.

test/Test/Data/Mutable/Vector.hs
Divesh-Otwani: Forgot to do this in my last commit. Just force pushed now.

test/Test/Data/Mutable/Array.hs
```diff
@@ -6,31 +6,14 @@
...
--- we're defining term re-write rules.) For now, constructor-constructor
--- re-writes are skipped (e.g., write (write a i x) i y = write a i y).
---
--- TODO:
---  * Test failures for out of bound access
---  * Constructor - constructor rules, like
---                write (write a i x) i y = write a i y
+-- Remarks:
+--  * We don't test for failure on out-of-bound access (it's trivial)
+--  * We don't test the empty constructor because it's trivial
```
aspiwack, line 16:
```suggestion
--  * We don't test for failure on out-of-bound access
--  * We don't test the empty constructor
```

test/Test/Data/Mutable/Vector.hs
```diff
@@ -7,10 +7,14 @@
...
---  * Constructor - constructor rules, like
---                write (write v i x) i y = write v i y
+-- See the testing framework explained in Test.Data.Mutable.Set.
+--
+-- The combination of axioms and homomorphisms provided functionally specify
+-- the behavior of vectors.
+--
+-- Remarks:
+--  * We don't test for failure on out-of-bound access (it's trivial)
+--  * We don't test the empty constructor because it's trivial
```
aspiwack, line 17:
```suggestion
--  * We don't test for failure on out-of-bound access
--  * We don't test the empty constructor
```

aspiwack (approved):

PR #266, opened 2020-11-17
Axiomatic hashmap tests

### LINKED ISSUE #93: Elegance of Tests

Divesh-Otwani:
In the style of #265, here are the other tests for hashmaps.

I found a bug in `mapMaybeWithKey` and rewrote it hopefully with the same complexity and more readability. The bug was identified with the test `refMap`; it appears `mapMaybeWithKey` lets the map be applied twice to elements with key 0.

It's easy to see why my algorithm is correct and terminates but not so easy to see why it's O(n), though I conjecture that it is and with some hand-waving, it seems probable.

Also, along with the other two PRs, this closes #93.

### DIFF
--- a/src/Data/HashMap/Mutable/Linear.hs
+++ b/src/Data/HashMap/Mutable/Linear.hs
@@ -61,6 +61,8 @@ import Prelude ((+))
 import qualified Data.Maybe as NonLinear
 import qualified Data.Function as NonLinear
 import qualified Prelude
+import Unsafe.Coerce (unsafeCoerce)
+import qualified Unsafe.Linear as Unsafe
 
 -- # Implementation Notes
 -- This is a simple implementatation of robin hood hashing.
@@ -204,7 +206,7 @@ alterF f key hm =
             -- We need to delete it.
             Ur Nothing ->
               Array.write arr ix Nothing & \arr' ->
-                shiftSegmentBackward cap arr' ((ix + 1) `mod` cap) & \arr'' ->
+                shiftSegmentBackward 1 cap arr' ((ix + 1) `mod` cap) & \arr'' ->
                   HashMap
                     (count - 1)
                     arr''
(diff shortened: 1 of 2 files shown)


src/Data/HashMap/Mutable/Linear.hs
```diff
@@ -270,56 +269,77 @@ mapMaybe f = mapMaybeWithKey (\_k v -> f v)
...
-      , Ur ret
-      )
-    | otherwise =
-      go
-        (if incr then (curr + 1) `mod` sz else curr)
-        end sz ret arr
+mapMaybeWithKey (f :: k -> v -> Maybe v') (HashMap _ arr) = Array.size arr &
+  \(Ur size, arr1) -> countAliveCells 0 (size-1) 0 (Array.map f' arr1) & \case
+    (Ur 0, arr2) -> HashMap 0 arr2
+    (Ur n, arr2) -> HashMap n (cleanUpDeletes arr2)
```
utdemir, line 275: The issue with this approach is that you pass through the input array multiple times (at least twice), and on each one of them you make multiple reads and writes. However, the original algorithm does a single pass, and only a single read/write pair for each element. So, even if we can prove the new algorithm to be `O(n)`, I'd still think that it would have bigger constant factors.

So, I would suggest to either find the error on the old algorithm, or implement the new algorithm with the same performance characteristics as the previous one.

test/Test/Data/Mutable/HashMap.hs
```diff
@@ -135,14 +158,14 @@ val = do
...
--- | Random pairs
+-- | Random pairs with no duplicate keys
 keyVals :: Gen [(Int, String)]
 keyVals = do
   size <- Gen.int $ Range.linear 0 maxSize
   let sizeGen = Range.singleton size
   keys <- Gen.list sizeGen key
   vals <- Gen.list sizeGen val
-  return $ zip keys vals
+  return $ zip (nubOrd keys) vals
```
utdemir, line 168: I think it might be good to test it using duplicate keys, because the behaviour on key collisions (on `fromList` eg.) is also significant.

aspiwack:
I think @utdemir has more opinion on this than I am, so I'll let you make the call on this.

test/Test/Data/Mutable/HashMap.hs
Divesh-Otwani: Sure -- I just did it to avoid making one test more complicated to write.

src/Data/HashMap/Mutable/Linear.hs
Divesh-Otwani: First and foremost we need an implementation that we can _reason_ is correct. Then after that we can worry about performance. *It's a bad idea to get a performant algorithm that passes all our tests and because it passes our tests, we tell ourselves that it's correct.*

The old algorithm isn't easy to reason about and though this one is better, I think it isn't ideal. I will try two things. First, I will try re-writing the old algorithm so we can reason it is correct (and hopefully find the bug in this manner). I'll rename and re-structure the sub functions and put precise pre- and post- conditions on all functions. Second, I'll try re-writing my algorithm to make it fully correct. I will then aim to get it the performance of the old one. However, if we can't do that, correctness is more important than performance so I'll stick with a correct but less efficient version.

Divesh-Otwani: Actually, looking at the old algorithm I'm actually going to say it's O(n^2) if we have n key-value pairs in the array. Consider something like `[Nothing, Nothing, ..., Nothing, Just pair_1, ..., Just pair_n, Nothing, ..., Nothing]` where `f k v = Nothing`. In the original algorithm, provided `pair_1` is not a spillover, you would delete that element, call `shiftSegmentBackward` and that would have to shift back `n-1` elements. Similarly, deleting `pair_2` shifts back `n-2` cells and this sum `1+2+...+n-1` is O(n^2).

So, I'm just going to come up with a better argument for the correctness and complexity of my algorithm.

utdemir: I don't think it counts as `O(n^2)`, since that `shiftSegmentBackwards` only process things as far as the hash collisions force us to. We assume the hash function to be uniform enough that hash collisions are uncommon. You are describing an extreme case where many elements corresponded to the same hash, and if that is the case _all_ HashMap functions will have to move all the elements.

As per your previous comment, I agree that there is little point of performance if the implementation is incorrect. And there are definitely things to do on the first approach to make them easier to reason about.

What I am trying to say is that, the second algorithm _inherently_ does multiple passes over the HashMap, which very likely is not as performant than doing it in a single pass (my ballpark estimation would be something like 4x slower for big enough HashMap's; given if I'm not wrong about the algorithm being `O(n)`).

Don't get me wrong, I'm not trying to defend the initial implementation. Do make it easier to reason about. My suggestion is that, make sure that your implementation also does a single pass over the array, an one read per cell and one write per resulting element (or maybe a few more in case of hash collisions, but my gut feeling is that even they can be avoided, which is indeed not the case in current implementation).

Divesh-Otwani: It's not rare for several elements to hash into a range where they overlap and have a contiguous segment. They don't all have to hash into the same slot to get the picture I drew. In the worst case, assuming uniform hashing, we (probably) don't have O(n^2) but do have several contiguous segments and I'd bet this is definitely worse than O(n).

I have an algorithm that's easier to reason about that does at most 2*n inspections where n is the size of the underlying array. I seriously doubt that it's possible to do this in a single pass with looking at each cell at most once. I'll commit that in a bit.

Divesh-Otwani: Ok, committed and ready.

utdemir (approved):
I haven't paid much attention to the correctness of the algorithm, but it definitely looks like it's doing minimal reads writes, so I am happy with it.

PR #272, opened 2020-11-27
Renamed single-letter imports

### LINKED ISSUE #229: Rename one-letter qualified imports in linear-base

Divesh-Otwani:
Closes #229 

I ran `grep -nr ' as .$' .` in the base and got 

```bash
./src/Streaming/Prelude/Linear.hs:18:> import qualified Streaming.Prelude as S
./src/Streaming/Prelude/Linear.hs:28:> import qualified Pipes as P
./src/Streaming/Prelude/Linear.hs:29:> import qualified Pipes.Prelude as P
./src/Data/V/Linear.hs:31:-- >>> import qualified Data.V.Linear as V
./src/Data/Unrestricted/Linear.hs:91:import qualified Data.V.Linear as V
```

and the last two are OK because of the module name `Data.V.Linear` and the other ones are comments from the old streaming documentation.

### DIFF
--- a/src/Control/Optics/Linear/Internal.hs
+++ b/src/Control/Optics/Linear/Internal.hs
@@ -48,7 +48,7 @@ import Data.Void
 import GHC.Exts (FUN)
 import GHC.Types
 import Prelude.Linear
-import qualified Prelude as P
+import qualified Prelude
 
 newtype Optic_ arr s t a b = Optical (a `arr` b -> s `arr` t)
 
@@ -71,7 +71,7 @@ assoc :: SymmetricMonoidal m u => Iso (a `m` (b `m` c)) (d `m` (e `m` f)) ((a `m
 assoc = iso Bifunctor.lassoc Bifunctor.rassoc
 
 (.>) :: Optic_ arr s t a b -> Optic_ arr a b x y -> Optic_ arr s t x y
-Optical f .> Optical g = Optical (f P.. g)
+Optical f .> Optical g = Optical (f Prelude.. g)
 
 
 lens :: (s %1-> (a, b %1-> t)) -> Lens s t a b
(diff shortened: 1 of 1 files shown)


src/Control/Optics/Linear/Internal.hs
```diff
@@ -48,7 +48,7 @@ import Data.Void
 import GHC.Exts (FUN)
 import GHC.Types
 import Prelude.Linear
-import qualified Prelude as P
+import qualified Prelude as Ur
```
utdemir, line 51: This probably are not in scope of this PR, but I was wondering if we should have some consistency here. I think I saw all three:

```haskell
import qualified Prelude
import qualified Prelude as NonLinear
import qualified Prelude as Ur
```

It might be good to decide on one. (My favourite is the first one, I guess, since we don't want single letter imports).

Divesh-Otwani: `Ur` is two letters one. Would you be OK with that?

utdemir: My problem is not about it being two letters, but more about we already having a `Data.Unrestricted.Linear` module and a `Ur` data type. So, when I see an `Ur` qualifier in a codebase, I am more likely to think of `Data.Unrestricted.Linear` instead of `Prelude`.

Divesh-Otwani: Very fair ... I'll go with `import qualified Prelude` then.

mboes: There would be nothing wrong with never creating an alias for modules when importing them, except that it would be unbearably verbose. Because at least that way, when seeing an occurrence of the module in source code, readers know exactly what module it is. So shortening long hierarchical module names while still retaining the most significant bit of their name (usually the final component) strikes the right balance.

utdemir (approved):

PR #273, opened 2020-12-02
Functor & Monad Module Organization

### LINKED ISSUE #271: Messy module structure of Functors
* We should have a file where all `Data` functors and the like are exported and all `Control` functors and the like are exported
* However, we should also have the traditional `Data.Applicative.Linear`, `Data.Traversable.Linear` and `Control.Applicative.Linear` modules.
* This will be hard to get right with the module dependencies so this should be thought out and mentioned in the `DESIGN.md`.

Divesh-Otwani:
This is very much a draft and not ready for any substantial review. When finished, it will close #271.

We need to rethink our module organization, because of problems laid out in #274  . So far, I've just done a minimal amount to get the src/Data/Functor/* stuff separated but there's a ton more to just break things into a clean structure. Lots more to say as I continue. Unfortunately since this stuff is too tightly coupled, this won't be a small PR; it can't be more modular than something that touches a lot of files.

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -25,6 +25,7 @@ library
     Control.Optics.Linear.Lens
     Control.Optics.Linear.Prism
     Control.Optics.Linear.Traversal
+    Data.Applicative.Linear
     Data.Array.Destination
     Data.Array.Mutable.Linear
     Data.Array.Mutable.Unlifted.Linear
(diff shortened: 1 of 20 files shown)


Divesh-Otwani:
I'm breaking this up into atomic PRs.

Issue #274, opened 2020-12-02
Design: Consistent & Non-Abitrary Module Structure


Divesh-Otwani:
The internal module structure has some major (not completely disjoint) problems:

- Arbitrariness
  - We shouldn't make arbitrary decisions as to what is or isn't in an internal module.
  - Example: why is `flip` defined in `Prelude.Linear` but not in `Prelude.Linear.Internal`? What is the rule by which we determine what goes in an internal module and what goes together?
  - Example: the same question applies to `(<$)` in `Data.Functor.Linear`?
  - What exactly is the reason we put `ReaderT` in `Control.Monad.Linear` but not in the internal modules?
  - What is the rule by which we decided to put `Consumable` and `Dupable` and `Movable` in `Data.Unrestricted`? With the exception of the last, these seem like separate concepts. 
- Cohesion
  - Internal modules should have clear **boundaries** that ideally are reflected in their name. `Internal` doesn't communicate what is internal. I should have some idea of what ought to be in this module and what ought not to be -- this way, if I see it imported somewhere I know basically what is imported. This ideally comes from each module having a clearly defined and self contained **purpose**, usually about a singular _concept_.  
  - Example: when I look at `Data.Unrestricted.Linear` and see the import `Data.Functor.Linear.Internal` I have no idea what this is importing. Is it importing all of `Data.Functor.Linear` essentially? It seems the purpose is to avoid a cycle of module dependencies but this still tells me very little. It seems like the original programmer has a clear idea of what that module is and what it does but as a new reader, I do not. This is a problem: there shouldn't be this silent knowledge or "internal wisdom" that makes it hard to do things unless you know all this background that isn't explicitly written down. [Cf., [normalization of deviance](https://danluu.com/wat/).] If stuff like this grows, this becomes unsustainable.
  - Example: When I look at imports of `Control.Monad.Linear.Internal` I have no idea what they are really importing. I don't know the boundaries of what is and is not in this module, nor do I understand it's purpose.
- Consistency
  - The template by which we make internal modules should be consistent. If one part of the codebase just makes internal modules called `Internal` which just holds anything and everything core that requires no internal dependencies and other parts like `Streaming` have internal modules like `Streaming.Type` this is inconsistent.
 
I propose the following system and subsequent changes:

- System
  - We write this system down in the `Design.md` document.
  - Internal modules are not named 'Internal' and instead have the name of a single concept, which defines clear boundaries of what one can expect or not expect to be in them.
[...]

PR #275, opened 2020-12-04
Reorganized Data.Unrestricted

Divesh-Otwani:
Working towards solving #274, this PR reorganizes `Data.Unrestricted.Linear` by breaking it up into these self-explanatory modules:

- `Data.Unrestricted.Internal.Ur`
- `Data.Unrestricted.Internal.Consumable`
- `Data.Unrestricted.Internal.Dupable`
- `Data.Unrestricted.Internal.Movable`
- `Data.Unrestricted.Internal.Instances`

I also reduced one dependency in `Data.Functor.Linear` which will be useful in later PRs and is a direct consequence of the changes in this one.


src/Data/Unrestricted/Internal/Consumable.hs
```diff
@@ -0,0 +1,33 @@
...
+{-# LANGUAGE LinearTypes #-}
+
+module Data.Unrestricted.Internal.Consumable
+  (
+  -- * Consumable
+    Consumable(..)
+  , void
+  , lseq
+  , seqUnit
+  )
```
aspiwack, line 11: Since these are internal modules, maybe it's better to export everything (and not have an export list at all)

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -0,0 +1,232 @@
...
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Data.Unrestricted.Internal.Instances where
```
aspiwack, line 15: Why do you prefer this to defining these instances in the files where the classes are defined?

src/Data/Unrestricted/Internal/Movable.hs
```diff
@@ -0,0 +1,19 @@
...
+
+import Data.Unrestricted.Internal.Ur
+import Data.Unrestricted.Internal.Dupable
+
+-- | The laws of the @Movable@ class mean that @move@ is compatible with
+-- @consume@ and @dup@.
+--
+-- * @case move x of {Ur _ -> ()} = consume x@
+-- * @case move x of {Ur x -> x} = x@
+-- * @case move x of {Ur x -> (x, x)} = dup2 x@
```
aspiwack, line 16: Let's take the opportunity to add to the laws that `move` represents coalgebras of the `Ur` comonads.
```suggestion
-- | Use @'Movable' a@ to represent a type which can be used many times even when given linearly. Simple data types such as 'Bool' or @[]@ are 'Movable'. Though, bear in mind that this typically induces a deep copy of the value.
--
-- Formally, @'Movable' a@ is the class of [coalgebras](https://ncatlab.org/nlab/show/coalgebra+over+a+comonad) of the 'Ur' comonad. That is
--
-- * @unur (move x) = x@
-- * @move \@(Ur a) (move \@a x) = fmap (move \@a) $ move \@a x  
--
-- Additionally, a 'Movable' instance must be compatible with its 'Dupable' parent instance. That is:
--
-- * @case move x of {Ur _ -> ()} = consume x@
-- * @case move x of {Ur x -> (x, x)} = dup2 x@
```

aspiwack (approved):
Wibbles. Then you can merge.

src/Data/Unrestricted/Internal/Consumable.hs
Divesh-Otwani: I put it there because it helps give a glance at what's defined in the module. I think it improves readability (at least for me, which might just be me) and can't hurt.

src/Data/Unrestricted/Internal/Instances.hs
Divesh-Otwani: Oh, this should be documented in a comment: it avoids a circular dependence with `Ur` and `Movable`. Since many classic `Movable` instances depend on `Ur` and the definition of `Movable` depends on `Ur`. 

After separating the `Movable` instances, I thought it clean to separate the rest too.

PR #276, opened 2020-12-08
Module org/data.functor

Divesh-Otwani:
Depends on #275.

This PR moves Data.Functor.Internal.Traversable to it's own module and breaks up Data.Functor into internal modules.

- I thought that since `Traversable` depended on `Control` functors it made sense to make it its own module. I can move it back or we could move it to `Control.Functor` in the future if that makes more sense.
- I moved `void` from the `Prelude` to `Data.Functor.Linear` since it resolved a dependency and seemed to fit better there. Should I re-export it in the Prelude? 
- I made this depend on #275 since we need to be able to only have `Consumable` imported in the internal functor module to break an import loop.

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -37,8 +37,8 @@ library
     Data.Either.Linear
     Data.Eq.Linear
     Data.Functor.Linear
-    Data.Functor.Internal.Functor
-    Data.Functor.Internal.Applicative
+    Data.Functor.Internal.Linear.Functor
+    Data.Functor.Internal.Linear.Applicative
     Data.HashMap.Mutable.Linear
     Data.List.Linear
     Data.Maybe.Linear
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -39,6 +39,7 @@ library
     Data.Functor.Linear
     Data.Functor.Internal.Linear.Functor
     Data.Functor.Internal.Linear.Applicative
+    Data.Functor.Internal.Linear.Traversable
     Data.HashMap.Mutable.Linear
     Data.List.Linear
     Data.Maybe.Linear
@@ -49,7 +50,6 @@ library
     Data.Profunctor.Linear
     Data.Semigroup.Linear
     Data.Set.Mutable.Linear
-    Data.Traversable.Linear
     Data.Tuple.Linear
     Data.Unrestricted.Internal.Consumable
     Data.Unrestricted.Internal.Dupable
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -38,7 +38,6 @@ library
     Data.Eq.Linear
     Data.Functor.Linear
     Data.Functor.Linear.Internal
-    Data.Functor.Linear.Internal.Traversable
     Data.HashMap.Mutable.Linear
     Data.List.Linear
     Data.Maybe.Linear
@@ -49,6 +48,7 @@ library
     Data.Profunctor.Linear
     Data.Semigroup.Linear
     Data.Set.Mutable.Linear
+    Data.Traversable.Linear
     Data.Tuple.Linear
     Data.Unrestricted.Internal.Consumable
     Data.Unrestricted.Internal.Dupable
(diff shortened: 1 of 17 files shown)


linear-base.cabal
```diff
@@ -37,8 +37,8 @@ library
     Data.Either.Linear
     Data.Eq.Linear
     Data.Functor.Linear
-    Data.Functor.Linear.Internal
-    Data.Functor.Linear.Internal.Traversable
```
utdemir, line 41: I don't think we should lose the word `Linear`. Although I am not sure if it should be `Data.Functor.Internal.Linear` or `Data.Functor.Linear.Internal`.

Divesh-Otwani: This may seem a bit long, but I''ll change it to these:
- `Data.Functor.Linear.Internal.Functor`
- `Data.Functor.Linear.Internal.Applicative`

... since the most specific thing about these internal modules is the _concept_ they contain and the concept of `Applicative` is qualified, say, by being data, being a functor, being linear, and the module being internal.

In terms of interfaces, things like`Data.Functor.Linear` make sense as linear flavors of familiar concepts. But I think internal modules shouldn't follow the same pattern and instead have the last name be one central concept so that it's clear when we import them internally what chiefly we're importing.

src/Control/Monad/Linear.hs
```diff
@@ -148,7 +149,7 @@ get :: (Applicative m, Dupable s) => StateT s m s
 get = state dup
 
 put :: (Applicative m, Consumable s) => s %1-> StateT s m ()
-put = void . replace
+put = Data.void . replace
```
aspiwack, line 152: It's completely unrelated, but do we have in our README, a documentation of our recommended qualifying conventions? That `Data.Functor.Linear` is typically qualified as `Data`?

src/Data/Functor/Internal/Linear/Applicative.hs
```diff
@@ -0,0 +1,79 @@
...
+
+import Data.Functor.Internal.Linear.Functor
+import Prelude.Linear.Internal
+import qualified Control.Monad.Trans.Reader as NonLinear
+import Data.Monoid.Linear hiding (Sum)
+import Data.Functor.Compose
+import Data.Functor.Const
+import Data.Functor.Identity
+
+-- # Applicative Defintion
```
aspiwack, line 18:
```suggestion
-- # Applicative definition
```

(mostly for the missing 'i', but I think that the capital 'D' is unwarranted as well)

src/Data/Traversable/Linear.hs
```diff
@@ -9,7 +9,7 @@
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE TypeApplications #-}
 
-module Data.Functor.Linear.Internal.Traversable
+module Data.Traversable.Linear
```
aspiwack, line 12: I think I will disagree with this. `Traversable` is still a subclass of `Data.Functor`. A very natural one to boot. The fact that it mentions `Control.Functor` is completely irrelevant to the problem. The whole point of having internal modules is that we don't have to care about compilation order when we make interfaces. So I say: move `Traversable` back into `Data.Functor.Linear`.

src/Data/Functor/Linear.hs
```diff
@@ -19,23 +19,14 @@
 module Data.Functor.Linear
   ( -- * Data Functor Hierarchy
     Functor(..)
-  , Applicative(..)
   , (<$>)
   , (<$)
+  , void
```
aspiwack, line 24: It's a reasonable place for this. No need to re-export it in the prelude, since there can be a `Control.void` as well, and there is little reason to favour one over the other.

In an ideal world, the prelude would re-export the `Data` and `Control` namespace, but Haskell can't do that yet. So be it.

aspiwack (approved):
I want `Traversal` back into `Data.Functor.Linear` (see below). Apart from that, it does look pretty good. So no need for a second review round from me. Merge when ready.

src/Control/Monad/Linear.hs
Divesh-Otwani: No, I"ll add this in a later PR.

PR #278, opened 2020-12-10
Control.Monad => Control.Functor and sub divisions; Data.V divided up

Divesh-Otwani:
Working towards #274 this PR 

- renames `Control.Monad` into `Control.Functor` to parallel `Data.Functor`
- divides up the contents of `Control.Monad` into internal modules postfixed with `Class`, `Instances`, `MonadTrans`, `Reader`, `State`
- divides up `Data.V.Linear` to finally put to bed a rather annoying module dependency circle, explained in a comment there

Also, this looks like a lot of changes because of the line counts and such but is really just moving around a handful of files.


src/Data/V/Internal/Linear/V.hs
```diff
@@ -0,0 +1,166 @@
...
+  , fromIntegral
+  , error
+  , (-))
+import qualified Prelude as Prelude
+import Prelude.Linear.Internal
+import qualified Unsafe.Linear as Unsafe
+
+{- Developers Note
+
+See Data.V.Linear to see why we created this module structure.
```
aspiwack, line 50:
```suggestion
See “Developers Note” in Data.V.Linear to see why we created this module structure.
```

src/Data/V/Linear.hs
```diff
@@ -59,40 +48,33 @@ module Data.V.Linear
...
 
-newtype V (n :: Nat) (a :: Type) = V (Vector a)
-  deriving (Eq, Ord, Prelude.Functor)
-  -- Using vector rather than, say, 'Array' (or directly 'Array#') because it
-  -- offers many convenience function. Since all these unsafeCoerces probably
-  -- kill the fusion rules, it may be worth it going lower level since I
-  -- probably have to write my own fusion anyway. Therefore, starting from
-  -- Vectors at the moment.
+Remark: We tried to create Data.V.Internal.Linear.Instances but for some
+reason GHC couldn't export the instances from there.
```
aspiwack, line 72: This sounds very unlikely. What did you actually tried to do, and what didn't work?

src/Data/V/Internal/Linear/V.hs
```diff
@@ -0,0 +1,166 @@
...
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UnboxedTuples #-}
+{-# LANGUAGE UndecidableInstances #-}
+module Data.V.Internal.Linear.V
```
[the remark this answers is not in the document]
Divesh-Otwani: I've actually been doing `Internal.Linear` in all my module organization PRs. It made sense to me as a way to put all internal modules in one place, even if they aren't `Linear`.

src/Data/V/Linear.hs
Divesh-Otwani: I tried to do exactly what's done in `Data.Unrestricted` and that failed.

src/Data/V/Internal/Linear/V.hs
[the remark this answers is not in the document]
aspiwack: Name hierarchies in Haskell are rather unintuitive to me. But I think that `XXX.Linear.Internal` is more consistent with the rest of Haskell. Since what we really have is a module named `Foo.Bar` and its internal module behind `Foo.Bar.Internal`.

Worth changing for consistency.

aspiwack (approved):

Issue #268, opened 2020-11-24
nix error: cannot coerce null to a string


chikeabuah:
**Describe the bug**
When I run `stack build` I get the nix coercion error shown in the attached image.

**To Reproduce**
I downloaded the linear-base repo and nix, then ran stack build.

**Environment**
 - OS name + version: macOS Mojave Version 10.14.3
 - Version of the code: master branch (unchanged)

**Additional context**
<img width="1271" alt="nixerror" src="https://user-images.githubusercontent.com/2280440/100105981-f7c59e80-2e35-11eb-9b86-642394faaa90.png">

joprice:
Looks like this is caused by the `glibcLocales` modification here 

https://github.com/utdemir/nixpkgs/commit/647a878e061b1b63efd2094eb9e546c2a7aa0d58#diff-71cf0d63355f78e3ae66dff9521cb419b983e123c1c1d090fffa499c6a716068R143

which is only available on linux:

```nix
glibcLocales = if stdenv.hostPlatform.libc == "glibc" then callPackage ../development/libraries/glibc/locales.nix { } else null;
```

Making it optional gets you past this:

``` nix
    export LANG="en_US.UTF-8";
  '' + stdenv.lib.optionalString (glibcLocales != null) ''
    export LOCALE_ARCHIVE=${glibcLocales}/lib/locale/locale-archive
  '' + ''
```

But you might hit other issues with newer versions of OSX (I'm on 11.0.1, not sure what version it was when the "libsystem" errors started). For that, I think the branch nix fork will need to be rebased against a newer nix nightly commit. I forked it and did so and it started building.

utdemir:
Thank you for finding the culprit @joprice! It definitely make sense to me.

I'll rebase our fork, apply the fix and try to find a Mac to confirm if it works.

joprice:
No problem. This should be obvious from the error, but I forgot that I also had to remove `happy = self.happy_1_20_0;`, since the latest master has 1.20.

PR #279, opened 2020-12-14
Fix builds on MacOS

utdemir:
Closes https://github.com/tweag/linear-base/issues/268.

I applied the fix mentioned at #268 to our fork of nixpkgs, rebased it to the current nixpkgs master (appearently it contains another fix for darwin builds), and pushed the branch to tweag's nixpkgs instead of my own. See the last two commits on https://github.com/tweag/nixpkgs/tree/78d9696f5d1524fcf2fd947f012ffc08fcba4dd1 for the nixpkgs changes.

I also added a parameter to `shell.nix` which allows passing `system` as a parameter (so I could test this change on a LInux machine with a darwin build machine.).

I tested that the `shell.nix` builds on both Linux and MacOS Mojave

### DIFF
--- a/nixpkgs.nix
+++ b/nixpkgs.nix
@@ -1,9 +1,10 @@
 let
-  # https://github.com/tweag/nixpkgs/tree/update-ghchead-20201001
-  rev = "60a06c5a2ec88392c1eb5e252367dde4e0ead16c";
-  sha256 = "0952kxb7zlw81vwk72dm4cxs01ygqgbxsy0ibqsj7khr7xp115jh";
+  # https://github.com/tweag/nixpkgs/tree/ud/update-ghchead-20201001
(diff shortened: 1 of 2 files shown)


aspiwack (approved):
It's a bit unclear to me what's happening here. But as long as it works, I'm fine. In the relatively short term, now, we should be able to get rid of any fork anyway (I don't even remember why we have one)

utdemir:
> I don't even remember why we have one

It think it is because `ghcHEAD` in upstream nixpkgs was not new enough at one point. I don't remember what it was missing, but it might be the `%1 ->` syntax.

Probably we can check if the newly merged `ghc901` attribute contains all the changes we want, or we can open a PR to upstream. I was postponing those since hopefully GHC 9 release is soon enough that, as you said, we won't need those fixes.

PR #283, opened 2020-12-29
Instances module for Data.V.Linear

Divesh-Otwani:
We weren't able to create a `Instances` module for `Data.V.Linear` as we did for `Data.Functor.Linear` but this PR fixes that.

The issue: to be able to export a module, GHC requires that module to have something to export. 

Our other `Instances` modules exported something.

To get this to work I had `Data.V.Linear.Internal.Instances` re-export `Data.fmap` which is harmless since we have Data.Functor instances in it. [To be more precise, by "harmless", I mean for any use of `Data.V.Linear`, it will be indistinguishable from if the `Instances` module did not export anything.]

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -65,6 +65,7 @@ library
     Data.Unrestricted.Linear
     Data.V.Linear
     Data.V.Linear.Internal.V
+    Data.V.Linear.Internal.Instances
     Data.Vector.Mutable.Linear
     Debug.Trace.Linear
     Foreign.Marshal.Pure
--- a/src/Data/V/Linear.hs
+++ b/src/Data/V/Linear.hs
@@ -1,10 +1,5 @@
-{-# OPTIONS -Wno-orphans #-}
-{-# LANGUAGE FlexibleInstances #-}
+{-# OPTIONS_GHC -Wno-dodgy-exports #-}
 {-# LANGUAGE NoImplicitPrelude #-}
-{-# LANGUAGE UndecidableInstances #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications #-}
-
 -- | This module defines vectors of known length which can hold linear values.
 --
 -- Having a known length matters with linear types, because many common vector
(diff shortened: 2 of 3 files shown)


utdemir:
> The issue: to be able to export a module, GHC requires that module to have something to export.

I don't think this is the case. It triggers a warning (`-Wdodgy-exports`) to have an empty exports list (which makes sense because the only reason we're in this state is because we're using `-Wno-orphans`). We can fix it by either using `-Wno-dodgy-exports` and using an empty exports list, or just not having an exports list as below:

```
index 930d06a..1cae74a 100644
--- a/src/Data/V/Linear.hs
+++ b/src/Data/V/Linear.hs
@@ -40,11 +40,10 @@ module Data.V.Linear
   , iterate
   -- * Type-level utilities
   , caseNat
-  , module Data.V.Linear.Internal.Instances
   ) where
 
 import Data.V.Linear.Internal.V
-import Data.V.Linear.Internal.Instances
+import Data.V.Linear.Internal.Instances ()
 
 {- Developers Note
 
diff --git a/src/Data/V/Linear/Internal/Instances.hs b/src/Data/V/Linear/Internal/Instances.hs
index 8d4693a..d4501e1 100644
--- a/src/Data/V/Linear/Internal/Instances.hs
+++ b/src/Data/V/Linear/Internal/Instances.hs
@@ -9,7 +9,7 @@
 --
 -- To get this to compile, we need this module to export something
 -- so it can be exported in Data.V.Linear
-module Data.V.Linear.Internal.Instances (Data.fmap) where
+module Data.V.Linear.Internal.Instances where
 
 import Data.V.Linear.Internal.V
 ```

Divesh-Otwani:
Yup -- it's just that warning; missed that.

Have you checked that this actually exports the instances? (For some reason I think I recall trying that and seeing that it didn't.) In any case, I'll try that and see if it works, and if not I'll just use `-Wno-dodgy-exports`.

utdemir:
> Have you checked that this actually exports the instances? (For some reason I think I recall trying that and seeing that it didn't.) 

I haven't checked it, but I am pretty sure as soon as the instances module is imported somewhere in the dependency chain (as we do from `Data.V.Linear`) the instances will be in scope; as GHC tries very hard to keep the instances coherent.

Divesh-Otwani:
So:
- We can remove the export and instances are exported.
- I'm not doing that since I think this is a bit clearer to anyone reading it and we've (really, _I've_) done it elsewhere. 
- We use `-Wno-dodgy-exports`.

aspiwack (approved):

aspiwack (commented):
Argl. I failed to send the comment below this morning. Try and check if it works after having merge then. Let me know :slightly_smiling_face:

PR #285, opened 2020-12-29
Import conventions for Data, Control

Divesh-Otwani:
This addresses a comment in #278:

> It's completely unrelated, but do we have in our README, a documentation of our recommended qualifying conventions? That Data.Functor.Linear is typically qualified as Data?

### DIFF
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -52,10 +52,17 @@ Interfacing linear code with regular Haskell is done, for instance, through the 
 The data type `Ur`, short for _unrestricted_ lets you store an
 unrestricted value inside a linear value.
 
+### Import Conventions
+
+We've designed `linear-base` to work nicely with the following import conventions:
+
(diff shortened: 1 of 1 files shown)


docs/USER_GUIDE.md
```diff
@@ -52,10 +52,17 @@ Interfacing linear code with regular Haskell is done, for instance, through the
 The data type `Ur`, short for _unrestricted_ lets you store an
 unrestricted value inside a linear value.
 
+### Import Conventions
+
+We've designed `linear-base` to work nicely with the following import conventions:
+
+1. `import qualified Data.Functor.Linear as Data`
+2. `import qualified Control.Functor.Linear as Control`
```
aspiwack, line 60: Since it's not actually an ordered list, it may be preferable to use bullets instead

```suggestion
- `import qualified Data.Functor.Linear as Data`
- `import qualified Control.Functor.Linear as Control`
```

aspiwack (approved):

PR #287, opened 2021-01-04
More generic push arrays 1/2

Divesh-Otwani:
This PR flushes out the design laid out in the `better-push-array` branch. [In an effort to avoid plagiarism: note that all of these are @aspiwack's ideas.] We represent push arrays with this clever little rank-2 thing:

```haskell
data Array a where
  Array :: (forall m. Monoid m => (a -> m) -> m) %1-> Array a
```
that takes a polymorphic conversion to a monoid for some element, and creates a monoid. This way it represents the monoidal concatenation of a certain natural (including zero) number of `a`s, in some unknown order.

We instantiate this with a function that takes an `a` and makes it an `ArrayWriter a`:

```haskell
data ArrayWriter a where
  ArrayWriter :: (DArray a %1-> ()) %1-> !Int -> ArrayWriter a
```

which holds the ingredients needed to write some number of elements, without holding the space to do so.

This is part 1/2 because I still have to add the fold functions.

### DIFF
--- a/src/Data/Array/Destination.hs
+++ b/src/Data/Array/Destination.hs
@@ -123,6 +123,7 @@ module Data.Array.Destination
   , mirror
   , fromFunction
   , fill
+  , dropEmpty
   )
   where
 
@@ -178,6 +179,14 @@ fill = Unsafe.toLinear2 unsafeFill
         error "Destination.fill: requires a destination of size 1"
       else
         unsafeDupablePerformIO Prelude.$ MVector.write ds 0 a
+
+-- | @dropEmpty dest@ consumes and empty array and fails otherwise.
+dropEmpty :: HasCallStack => DArray a %1-> ()
+dropEmpty = Unsafe.toLinear unsafeDrop where
+  unsafeDrop :: DArray a -> ()
+  unsafeDrop (DArray ds)
+    | MVector.length ds > 0 = error "Destination.dropEmpty on non-empty array."
+    | otherwise = ()
 
 -- | @'split' n dest = (destl, destr)@ such as @destl@ has length @n@.
 --
--- a/src/Data/Array/Destination.hs
+++ b/src/Data/Array/Destination.hs
@@ -186,7 +186,7 @@ dropEmpty = Unsafe.toLinear unsafeDrop where
   unsafeDrop :: DArray a -> ()
   unsafeDrop (DArray ds)
     | MVector.length ds > 0 = error "Destination.dropEmpty on non-empty array."
-    | otherwise = ()
+    | otherwise = ds `seq` ()
 
 -- | @'split' n dest = (destl, destr)@ such as @destl@ has length @n@.
 --
--- a/src/Data/Array/Polarized.hs
+++ b/src/Data/Array/Polarized.hs
@@ -102,8 +102,8 @@ module Data.Array.Polarized
 import qualified Data.Array.Polarized.Pull.Internal as Pull
 import qualified Data.Array.Polarized.Pull as Pull
 import qualified Data.Array.Polarized.Push as Push
+import qualified Data.Foldable as NonLinear
 import Prelude.Linear
-import qualified Prelude
 import Data.Vector (Vector)
 
 -- DEVELOPER NOTE:
@@ -130,8 +130,7 @@ import Data.Vector (Vector)
 -- NOTE: this does NOT require allocation and can be in-lined.
 transfer :: Pull.Array a %1-> Push.Array a
 transfer (Pull.Array f n) =
-  Push.Array
-    (\k -> Prelude.foldl (\m i -> m <> (k (f i))) mempty [0..(n-1)])
+  Push.Array (\k -> NonLinear.foldMap' (\x -> k (f x)) [0..(n-1)])
 
 -- | This is a shortcut convenience function
 -- for @transfer . Pull.fromVector@.
--- a/src/Data/Array/Polarized.hs
+++ b/src/Data/Array/Polarized.hs
@@ -99,11 +99,11 @@ module Data.Array.Polarized
   )
   where
 
-import qualified Data.Array.Destination as DArray
 import qualified Data.Array.Polarized.Pull.Internal as Pull
 import qualified Data.Array.Polarized.Pull as Pull
 import qualified Data.Array.Polarized.Push as Push
 import Prelude.Linear
+import qualified Prelude
 import Data.Vector (Vector)
 
 -- DEVELOPER NOTE:
@@ -129,7 +129,9 @@ import Data.Vector (Vector)
 -- | Convert a pull array into a push array.
 -- NOTE: this does NOT require allocation and can be in-lined.
 transfer :: Pull.Array a %1-> Push.Array a
-transfer (Pull.Array f n) = Push.Array (\g -> DArray.fromFunction (\i -> g (f i))) n
+transfer (Pull.Array f n) =
+  Push.Array
+    (\k -> Prelude.foldl (\m i -> m <> (k (f i))) mempty [0..(n-1)])
 
 -- | This is a shortcut convenience function
 -- for @transfer . Pull.fromVector@.
--- a/src/Data/Array/Polarized/Push.hs
+++ b/src/Data/Array/Polarized/Push.hs
@@ -22,7 +22,6 @@ import qualified Data.Array.Destination as DArray
 import Data.Array.Destination (DArray)
 import Data.Vector (Vector)
 import qualified Prelude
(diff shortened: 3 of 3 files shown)


src/Data/Array/Polarized.hs
```diff
@@ -129,7 +129,9 @@ import Data.Vector (Vector)
 -- | Convert a pull array into a push array.
 -- NOTE: this does NOT require allocation and can be in-lined.
 transfer :: Pull.Array a %1-> Push.Array a
-transfer (Pull.Array f n) = Push.Array (\g -> DArray.fromFunction (\i -> g (f i))) n
+transfer (Pull.Array f n) =
+  Push.Array
+    (\k -> Prelude.foldl (\m i -> m <> (k (f i))) mempty [0..(n-1)])
```
utdemir, line 134: I'm not sure, but in my experience `foldl` is rarely the most efficient fold. Probably I'd go with `foldl'`.

src/Data/Array/Polarized/Push.hs
```diff
@@ -11,55 +10,113 @@
...
+      (DArray.split l1 darr) & \(darr1,darr2) -> consume (k1 darr1, k2 darr2))
+    (l1+l2)
+-- Remark. In order for the function above to work, consume must forcibly
+-- evaluate both tuples. If it was lazy, then we might not actually perform
+-- @k1@ or @k2@ and the unsafe IO won't get done. In general, this makes me
+-- think we haven't spelled out the careful rules of what consuming functions
+-- must follow so that we can release a safe API.
+
+emptyWriter :: ArrayWriter a
+emptyWriter = ArrayWriter (Unsafe.toLinear (const ())) 0
```
utdemir, line 120: Unsafe call here assumes that we can just forget about the `DArray`. I think this is the case; but it might worth a comment here. (I think every unsafe call deserves a comment unless incredibly obvious.)

src/Data/Array/Polarized/Push.hs
```diff
@@ -11,55 +10,113 @@
...
+-------------------------------------------------------------------------------
+
+instance Data.Functor Array where
+  fmap f (Array k) = Array (\g -> k (\x -> (g (f x))))
+
+instance Prelude.Semigroup (Array a) where
+  (<>) x y = append x y
+
+instance Semigroup (Array a) where
+  (<>) = append
```
utdemir, line 81: This is a purely stylistic choice, so feel free to ignore this; but I'd just inline the `append` here, I don't see a much need for a separate function and the fewer names the better. 

Same goes with `mempty` and the variants for `ArrayWriter`.

src/Data/Array/Polarized/Push.hs
```diff
@@ -11,55 +10,113 @@
...
+  allocArrayWriter (ArrayWriter writer len) = DArray.alloc len writer
 
 -- | @`make` x n@ creates a constant push array of length @n@ in which every
 -- element is @x@.
-make :: a -> Int -> Array a
-make x n = Array (\k -> DArray.replicate (k x)) n
+make :: HasCallStack => a -> Int -> Array a
+make x n
+  | n < 0 = error "Making negative length push array"
+  | otherwise = Array (\makeA -> mconcat $ Prelude.replicate n (makeA x))
```
aspiwack, line 68: Base has an [`stimes`](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Semigroup.html#v:stimes) function for this. Maybe we should have one too?

src/Data/Array/Polarized/Push.hs
```diff
@@ -11,55 +10,113 @@
...
+addWriters (ArrayWriter k1 l1) (ArrayWriter k2 l2) =
+  ArrayWriter
+    (\darr ->
+      (DArray.split l1 darr) & \(darr1,darr2) -> consume (k1 darr1, k2 darr2))
+    (l1+l2)
+-- Remark. In order for the function above to work, consume must forcibly
+-- evaluate both tuples. If it was lazy, then we might not actually perform
+-- @k1@ or @k2@ and the unsafe IO won't get done. In general, this makes me
+-- think we haven't spelled out the careful rules of what consuming functions
+-- must follow so that we can release a safe API.
```
aspiwack, line 117: Sure, but, on the other hand, you cannot not consume them (try to forget `consume` or something). That's what linearity does for you. The unsafe IO is (if I haven't messed up) fully encapsulated in the `DArray` abstraction. So types guarantee that you can't get this wrong. Therefore, this comment is superfluous

```suggestion
```

src/Data/Array/Polarized/Push.hs
aspiwack: There shouldn't be an `Unsafe.toLinear` in this module. If we are missing a way to consume an empty destination array, then let's add it to the destination array module.

src/Data/Array/Polarized.hs
Divesh-Otwani: We can't use a linear `foldMap` without changing the type of `Push` and `Pull` to hold `Int %1-> a` and `(a %1-> m) -> m` respectively. I used `Prelude.foldMap`.

src/Data/Array/Polarized/Push.hs
Divesh-Otwani: I'll make an unsafe empty `DArray` consumer.

src/Data/Array/Polarized/Push.hs
Divesh-Otwani: I like it because it's a tiny bit more DRY even though the chance of changing the implementation is close to zero.

src/Data/Array/Polarized.hs
Divesh-Otwani: Or did you mean a `foldMap` on `Pull` arrays? @aspiwack

src/Data/Array/Polarized/Push.hs
Divesh-Otwani: Yes, I see. There's no way to write `((),()) %1-> ()` without using something unsafe that doesn't force the evaluation of both tuples.

src/Data/Array/Polarized/Push.hs
Divesh-Otwani: `Array $ \makeA -> stimes (makeA x)` might be more efficient if the `stimes` implementation for `ArrayWriter` could be more efficient.  However, I don't want to change `Monoid` in this PR.

src/Data/Array/Polarized.hs
aspiwack: I meant `foldMap` on lists. But you are right that there is some tension in the types. Let's forget about this for now.

src/Data/Array/Polarized/Push.hs
aspiwack: As you wish, but I think it's fine to lazily add stuff to the library (though I'm sensitive to the argument that changing a type class is not a cheap action, and may need to be done carefully).

It was not really about efficiency (though it may be), but about clarity. We can do this in a separate PR.

src/Data/Array/Destination.hs
```diff
@@ -179,6 +180,14 @@ fill = Unsafe.toLinear2 unsafeFill
       else
         unsafeDupablePerformIO Prelude.$ MVector.write ds 0 a
 
+-- | @dropEmpty dest@ consumes and empty array and fails otherwise.
+dropEmpty :: HasCallStack => DArray a %1-> ()
+dropEmpty = Unsafe.toLinear unsafeDrop where
+  unsafeDrop :: DArray a -> ()
+  unsafeDrop (DArray ds)
+    | MVector.length ds > 0 = error "Destination.dropEmpty on non-empty array."
+    | otherwise = ()
```
aspiwack, line 189: You know what? I don't think it's safe. We at least need to `seq` ds like this:
```suggestion
dropEmpty :: HasCallStack => DArray a %1-> ()
dropEmpty = Unsafe.toLinear unsafeDrop where
  unsafeDrop :: DArray a -> ()
  unsafeDrop (DArray ds)
    | MVector.length ds > 0 = error "Destination.dropEmpty on non-empty array."
    | otherwise = ds `seq` ()
```

But generally speaking, I'm wondering if we shouldn't replace `newtype DArray a = DArray (MVector …)` with `data DArray a where { DArray :: MVector … -> DArray a }`. It adds an indirection. But it will save some `Unsafe.toLinear` presumably. Maybe most. This may considerably reduce the kernel of trust, here.

Divesh-Otwani: I'll make that ^ an issue for later.

aspiwack (approved):

PR #288, opened 2021-01-05
Push arrays/fold

### LINKED ISSUE #62: Construction of Push Arrays
Currently, the only way to construct a (non-constant) Push Array is via a Pull Array. This is undesirable if the elements cannot be conveniently given in an arbitrary order, as a Pull Array demands. Similarly, the intuition for Push Arrays is that they should be friendly to construct, but this is not reflected in the API.

### LINKED ISSUE #80: Push Arrays
The current type of Push Arrays does not allow some operations which ought to be possible for a Push Array: for instance folding into a monoid or unzipping. It seems that the `DArray ->. ()` ought to be generalised somehow; at the moment, the only way to consume a Push Array is to allocate.

Divesh-Otwani:
### Summary
This PR adds to push arrays:

- folds and unzip and closes #80 
- more flexible construction and closes #62 

Depends on #287.

### Things I don't fully understand

- I don't fully understand the role of linear types here. Is it just to have the linear monoid so that the type necessarily writes each element once (and that in the allocation, we use destination arrays). Can we come up with a concise statement of how we apply linear types?
- Aren't we still missing the "use each index only once" property of pull arrays?

### DIFF
--- a/src/Data/Array/Polarized/Push.hs
+++ b/src/Data/Array/Polarized/Push.hs
@@ -11,9 +11,17 @@
 --
 -- This module is designed to be imported qualified as @Push@.
 module Data.Array.Polarized.Push
-  ( Array(..)
-  , alloc
+  (
+  -- * Construction
+    Array(..)
   , make
+  , singleton
+  , cons
+  , snoc
+  -- * Operations
+  , alloc
+  , foldMap
+  , unzip
   )
 where
 
@@ -22,7 +30,7 @@ import qualified Data.Array.Destination as DArray
 import Data.Array.Destination (DArray)
 import Data.Vector (Vector)
 import qualified Prelude
-import Prelude.Linear
+import Prelude.Linear hiding (unzip, foldMap)
 import GHC.Stack
 
 
@@ -65,6 +73,21 @@ make :: HasCallStack => a -> Int -> Array a
 make x n
   | n < 0 = error "Making a negative length push array"
   | otherwise = Array (\makeA -> mconcat $ Prelude.replicate n (makeA x))
+
+singleton :: a -> Array a
+singleton x = Array (\writeA -> writeA x)
+
+snoc :: a -> Array a %1-> Array a
+snoc x (Array k) = Array (\writeA -> (k writeA) <> (writeA x))
+
+cons :: a -> Array a %1-> Array a
+cons x (Array k) = Array (\writeA -> (writeA x) <> (k writeA))
+
+foldMap :: Monoid b => (a -> b) -> Array a %1-> b
+foldMap f (Array k) = k f
+
+unzip :: Array (a,b) %1-> (Array a, Array b)
+unzip (Array k) = k (\(a,b) -> (singleton a, singleton b))
 
 
 -- # Instances


b-mehta:
This and part 1 looks promising to me! I think `foldWith` looks sensible, and if `unzip` can be made to work then I'd be much more convinced. Nice stuff!

src/Data/Array/Polarized/Push.hs
```diff
@@ -67,6 +74,21 @@ make x n
...
+snoc x (Array k) = Array (\writeA -> (k writeA) <> (writeA x))
+
+cons :: a -> Array a %1-> Array a
+cons x (Array k) = Array (\writeA -> (writeA x) <> (k writeA))
+
+foldWith :: Monoid b => (a -> b) -> Array a %1-> b
+foldWith f (Array k) = k f
+
+unzip :: Array (a,b) %1-> (Array a, Array b)
+unzip = undefined
```
utdemir, line 90:
```suggestion
unzip (Array k) = k (\(a, b) -> (singleton a, singleton b))
```

This kind of make sense to me, but I'm unsure (both semantically and performance-wise).

src/Data/Array/Polarized/Push.hs
```diff
@@ -67,6 +74,21 @@ make x n
...
+singleton x = Array (\writeA -> writeA x)
+
+snoc :: a -> Array a %1-> Array a
+snoc x (Array k) = Array (\writeA -> (k writeA) <> (writeA x))
+
+cons :: a -> Array a %1-> Array a
+cons x (Array k) = Array (\writeA -> (writeA x) <> (k writeA))
+
+foldWith :: Monoid b => (a -> b) -> Array a %1-> b
+foldWith f (Array k) = k f
```
utdemir, line 87: This function is `foldMap` (specialized to `Array` and with a linear arrow), so I think we should call it that.

The `Foldable`/`Traversable` issues are likely relevant, but those assume the first parameter to be linear too. Bit unfortunate that we don't have a typeclass for this.

src/Data/Array/Polarized/Push.hs
```diff
@@ -1,5 +1,7 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE TypeInType #-}
```
utdemir, line 4: I don't think we need this.

src/Data/Array/Polarized/Push.hs
```diff
@@ -1,5 +1,7 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
```
utdemir, line 1: This also doesn't seem to be necessary.

src/Data/Array/Polarized/Push.hs
Divesh-Otwani: That's *very* clever and I feel dumb for not coming up with it after messing with it for a while, lol :smile_cat: 

It does make sense: `k` is something that takes something that writes an `a` to a monoid that represents a composable way to write an `a` to a cell of an array. Here, `k` takes something that changes `(a,b)` to a monoidal-computation that writes an `(a,b)` to a cell of an array. 

This implementation in words says "a way to write each pair of (a,b) into a monoidal writing-computation is to have a pair of monoidal writing-computations".  If you give this to `k` it will make a bunch of pairs of singletons and concatenate them.

src/Data/Array/Polarized/Push.hs
Divesh-Otwani: Remnants from my testing.

src/Data/Array/Polarized/Push.hs
Divesh-Otwani: Remnants from my testing.

aspiwack (approved):

PR #292, opened 2021-01-18
Adding Deques

Divesh-Otwani:
I wanted to build out a few data structures that are relatively easy to get out. This is one of them.

### DIFF
--- a/src/Data/Deque/Mutable/Linear.hs
+++ b/src/Data/Deque/Mutable/Linear.hs
@@ -7,7 +7,8 @@
 
 -- | Mutable Linear Deque
 --
--- This module provides a pure interface to a mutable deque.
+-- This module provides a pure interface to a bounded mutable deque.
+-- The deque has a maxiumum size and is represented with an array underneath.
 --
 -- It is designed to be imported qualfied:
 --
--- a/src/Data/Deque/Mutable/Linear.hs
+++ b/src/Data/Deque/Mutable/Linear.hs
@@ -1,0 +1,203 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+-- | Mutable Linear Deque
+--
+-- This module provides a pure interface to a mutable deque.
+--
+-- It is designed to be imported qualfied:
+--
+-- > import qualfied Data.Deque.Mutable.Linear as Deque
+module Data.Deque.Mutable.Linear
+  (
+  -- * Allocation
+    Deque
+  , alloc
+  , fromList
+  -- * Querying
+  , size
+  , length
+  , isFull
+  , peekFront
+  , peekBack
+  -- * Modification
+  , pushFront
+  , pushBack
+  , popFront
+  , popBack
+  , map
+  -- * Consumption
+  , toList
+  )
+where
+
+import qualified Data.Array.Mutable.Linear as Array
+import Data.Unrestricted.Linear
+import Prelude.Linear hiding (length, map)
+import qualified Prelude
+import GHC.Stack
+
+
+-- # Types
+-------------------------------------------------------------------------------
+
+data Deque a where
+  Deque :: !Int -> !Ptr -> !(Array.Array a) %1-> Deque a
+  -- This is: Deque length ptr array
+  --
+  -- The length is the number of elements stored.
+  -- The ptr is the starting pointer to the front end, and the deque
+  -- continues forward, wrapping the end if needed. Example:
+  --
+  -- [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+  --  ....^                 ^..........
+  --      |                 |
+  --   ptr+len             ptr
+  --
+  --
+  -- So the deque is: 7--8--9--10--0--1
+
+newtype Ptr = Ptr Int deriving Prelude.Num
+
+-- | The two faces of a deque
+data Face = Front | Back
+
+
+-- # Internal Helpers
+-------------------------------------------------------------------------------
+
+-- @backPtr offset len size ptr = ptr'@ where @ptr'@ is the pointer
+-- to the back of the deque + the offset
+-- Must have: len >= 1
+backPtr :: Int -> Int -> Int -> Ptr -> Int
+backPtr off len sz (Ptr p) = (off + p + len - 1) `mod` sz
+
+-- @prevPtr size ptr@ will be the previous pointer
+prevPtr :: Int -> Ptr -> Int
+prevPtr sz (Ptr p) = (p - 1 + sz) `mod` sz
+
+-- @nextPtr size ptr@ will be the next pointer
+nextPtr :: Int -> Ptr -> Int
+nextPtr sz (Ptr p) = (p + 1) `mod` sz
+
+
+-- # Allocation
+-------------------------------------------------------------------------------
+
+-- | Run a computation of an empty Deque with a given size
+alloc :: HasCallStack => Int -> (Deque a %1-> Ur b) %1-> Ur b
+alloc k f = Array.alloc k err $ \arr -> f (Deque 0 0 arr) where
+  err = Prelude.error "Accessing error element of a collection!"
+
+-- | Run a computation on a Deque that is deterimined by the given the list
+-- where we treat the start and end of the list as the left and right pointers,
+-- with the total capacity as the length of the list.
+fromList :: HasCallStack => [a] -> (Deque a %1-> Ur b) %1-> Ur b
+fromList xs f =
+  Array.fromList xs $ \arr -> f (Deque (Prelude.length xs) 0 arr)
+
+
+-- # Querying
+-------------------------------------------------------------------------------
+
+-- | The total capacity of the Deque
+size :: Deque a %1-> (Ur Int, Deque a)
+size (Deque len ptr arr) = Array.size arr &
+  \(sz, arr0) -> (sz, Deque len ptr arr0)
+
+-- | The number of elements currently stored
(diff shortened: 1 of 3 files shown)


src/Data/Deque/Mutable/Linear.hs
```diff
@@ -0,0 +1,182 @@
...
+import qualified Prelude
+import GHC.Stack
+
+
+-- # Types
+-------------------------------------------------------------------------------
+
+data Deque a where
+  Deque :: Int -> Maybe (Int,Int) -> !(Array.Array a) %1-> Deque a
+  -- This is: Deque size (Maybe (left ptr,right ptr)) array
```
utdemir, line 48: There is a slightly simpler representation, where you only store the number of elements in the queue and the starting index. It always wraps around to the beginning at the end of the array, so `(start + count) % size`  would give the position of the next element.

So only `Int -> Int -> Array a` should suffice, I think.

src/Data/Deque/Mutable/Linear.hs
```diff
@@ -0,0 +1,182 @@
...
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+-- | Mutable Linear Deque
+--
+-- This module provides a pure interface to a mutable deque.
+--
+-- It is designed to be imported qualfied:
+--
+-- > import qualfied Data.Deque.Mutable.Linear as Deque
```
utdemir, line 13: I think it is crucial to mention that this is a **bounded** deque.  I would even call it `Data.Dequeue.Bounded.Mutable.Linear` instad. It would also be good to mention:

* That the underlying storage is an array
* When full, newer elements overwrite the older elements

src/Data/Deque/Mutable/Linear.hs
```diff
@@ -0,0 +1,182 @@
...
+addRight :: HasCallStack => Deque a %1-> a -> Deque a
+addRight = add R
+
+pop :: HasCallStack => Dir -> Deque a %1-> Deque a
+pop _ (Deque _ Nothing arr) = error "Popping from empty deque." $ arr
+pop dir (Deque sz (Just (l,r)) arr)
+  | l==r = Deque sz Nothing arr
+  | otherwise = case dir of
+    L -> Deque sz (Just (nextIx sz l, r)) arr
+    R -> Deque sz (Just (l, prevIx sz r)) arr
```
utdemir, line 152: Should we also overwrite the removed elements(with something like `undefined`), so they can be garbage collected?

This is something I just thought of; and also applies to `Vector`; there too removed elements might still be pointed to from the underlying array and even though the interface does not expose them they would still be counted as alive. Until they are overwritten with another value or the vector shrinks or grows, that is.

I am actually not sure what the best practice is. What do you think?

src/Data/Deque/Mutable/Linear.hs
```diff
@@ -0,0 +1,182 @@
...
+
+-- | Remove the last added element from the left queue
+popLeft :: HasCallStack => Deque a %1-> Deque a
+popLeft = pop L
+
+-- | Remove the last added element from the right queue
+popRight :: HasCallStack => Deque a %1-> Deque a
+popRight = pop R
```
utdemir, line 160: Instead of `Left`/`Right`, I would refer to them as `Front`/`Back` as it is more common.

src/Data/Deque/Mutable/Linear.hs
```diff
@@ -0,0 +1,182 @@
...
+    R -> Array.write arr (nextIx sz r) x &
+      \arr' -> Deque sz (Just (l,nextIx sz r)) arr'
+
+-- | Add to the left queue
+addLeft :: HasCallStack => Deque a %1-> a -> Deque a
+addLeft = add L
+
+-- | Add to the right queue
+addRight :: HasCallStack => Deque a %1-> a -> Deque a
+addRight = add R
```
utdemir, line 144: I would expect those to be named `push*` (because of the presence of `pop`, and I think `push` is more common within queue implementations).

src/Data/Deque/Mutable/Linear.hs
```diff
@@ -0,0 +1,182 @@
...
+
+-- | Remove the last added element from the right queue
+popRight :: HasCallStack => Deque a %1-> Deque a
+popRight = pop R
+
+-- Note: We can't use a Prelude.Functor nor a Data.Functor
+-- because the mapped function need not be linear but we must
+-- consume the Deque linearly. The types don't align.
+map :: (a -> b) -> Deque a %1-> Deque b
+map f (Deque sz p arr) = Deque sz p (Array.map f arr)
```
utdemir, line 166: This implementation is `O(capacity)` insted of `O(number of items)`, I think it would be better to make it so that it only iterates the "live" elements.

Not that this is super important, since mapping is not a common operation for queues.

src/Data/Deque/Mutable/Linear.hs
```diff
@@ -0,0 +1,182 @@
...
+  | l <= r = (Ur (1 + r - l), Deque sz p arr)
+  | otherwise = (Ur (sz + 1 - l + r), Deque sz p arr)
+
+-- | We are full if the length equals the size
+isFull :: Deque a %1-> (Ur Bool, Deque a)
+isFull d =
+  length d & \(Ur len, Deque sz p arr) -> (Ur (len == sz), Deque sz p arr)
+
+peek :: HasCallStack => Dir -> Deque a %1-> (Ur a, Deque a)
+peek _ (Deque _ Nothing arr) = error "Peeking in empty deque." $ arr
```
utdemir, line 110: I would expect this function to return a `Maybe` (`(Ur (Maybe a), Deque a)`), I think it's a bit different than out-of-bounds array access; where when you're working with queues you usually don't expect them to have certain number of stuff inside and you are supposed to handle them being empty.

utdemir (commented):
You were late removing me as a reviewer, now I am aware of this and had to comment :).

I like the idea of having a few more data structures. I have some comments about the implementation and the interface; also some tests would be good.

src/Data/Deque/Mutable/Linear.hs
Divesh-Otwani: It doesn't overwrite. It crashes if you try to fill it after it's full. I think that makes more sense.

src/Data/Deque/Mutable/Linear.hs
Divesh-Otwani: That makes sense, but unless someone's storing giant things on a deque, stack or queue, I don't think it's important. The relevant case seems rare to me.

src/Data/Deque/Mutable/Linear.hs
Divesh-Otwani: Good idea :+1:

src/Data/Deque/Mutable/Linear.hs
Divesh-Otwani: Yup :+1:

src/Data/Deque/Mutable/Linear.hs
Divesh-Otwani: Due to lazyness, I don't think that's correct. When we read an overwritten value we don't evaluate the overwritten value. If the overwritten value was to be mapped, it won't be mapped.

src/Data/Deque/Mutable/Linear.hs
Divesh-Otwani: I don't think so because it's analogous to `top` for stacks which should fail if the stack is empty.

src/Data/Deque/Mutable/Linear.hs
Divesh-Otwani: Sigh, yes, I'll use the simpler representation.

src/Data/Deque/Mutable/Linear.hs
utdemir: I do not think Array.map works that way. Even though we are not evaluating the expression, we are still writing a thunk to an array which will evaluate that expression when forced. So even though we only evaluate the mapping function once per item, it calls `Array.write` for each Array cell, alive or dead.

src/Data/Deque/Mutable/Linear.hs
utdemir: I am having a hard time thinking when would I require a queue where I certainly know that it will contain less than some number of elements. I can think of cases where it's okay to be forgetful, or don't accept the new element, or in concurrent cases (which is not the case in this code) it should block; but in all cases I can think of there is no guarantee that the queue is not empty or it is not full. 

Even if we go with the crashy version, we should still mention that it is a bounded queue.

Divesh-Otwani: I agree with adding comments saying it is bounded. But ...

Do you think it's just better for me to use array resizing internally and have an unbounded structure to begin with? I feel like your argument makes sense and I can't think of a case where we want a bounded queue or stack.

Divesh-Otwani:
Just going to comment that I don't know how I closed this a moment ago. I think I accidentally used a strange key-sequence or something that caused that.

src/Data/Deque/Mutable/Linear.hs
Divesh-Otwani: Yes, that's correct... okay, I'll make a code comment.

src/Data/Deque/Mutable/Linear.hs
Divesh-Otwani: On second, thought, this ^ makes sense.

src/Data/Deque/Mutable/Linear.hs
Divesh-Otwani: Did this ^, done.

utdemir:
Sorry for the radio silence. For now I think we should hold off merging this PR (alongside with other new collection PR's #293 and #295), because we already have some performance issues with our existing linear collections; and having even more collections built on top of them would not solve that.

Once we get their performance right, we should take another look at these PR's, so we can merge them more confidently of their usefulness.

Divesh-Otwani:
Dude I agree. Let's do things right and try to only keep solid stuff in the codebase.

PR #294, opened 2021-01-19
Developer comment for push arrays

Divesh-Otwani:
I made myself a note to add a comment since there was some added clarifying discussion on issue #80.


aspiwack (approved):
It's probably still too brief an explanation. But it's definitely better than nothing.

PR #299, opened 2021-02-09
Remove some incomplete pattern matches

utdemir:
They only appear when linear-base with ghc HEAD (`ab5fd982`, slightly
newer than GHC 9.0.1 release). Likely the recent pattern exhaustiveness
checking are relevant, but I don't know why the released version not throws
these errors.

Anyhow, unfortunately they're only exhaustive because of the implementation
rather than the types, so I added other cases as bottoms.

(Making `scanr` return `NonEmpty` would solve one of the problems, but I do
not think it's worth diverging the API from `Data.List` and also we don't have
`Data.List.NonEmpty.Linear`)

### DIFF
--- a/src/Data/List/Linear.hs
+++ b/src/Data/List/Linear.hs
@@ -295,17 +295,26 @@ scanl1 f (x:xs) = scanl f x xs
 scanr :: Dupable b => (a %1-> b %1-> b) -> b %1-> [a] %1-> [b]
 scanr _ b [] =  [b]
 scanr f b (a:as) =
-  scanr f b as & \(b':bs') ->
-    dup2 b' & \(b'', b''') ->
-      f a b'' : b''' : bs'
+  scanr f b as & \case
+    (b':bs') ->
+      dup2 b' & \(b'', b''') ->
+        f a b'' : b''' : bs'
+    [] ->
+      -- this branch is impossible since scanr never returns an empty list.
+      Prelude.error "impossible" a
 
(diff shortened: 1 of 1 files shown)


src/Data/List/Linear.hs
```diff
@@ -295,17 +295,26 @@ scanl1 f (x:xs) = scanl f x xs
...
-  scanr1 f as & \(a':as') ->
-    dup2 a' & \(a'', a''') ->
-      f a a'' : a''' : as'
+  scanr1 f as & \case
+    (a':as') ->
+      dup2 a' & \(a'', a''') ->
+        f a a'' : a''' : as'
+    [] ->
+      -- this branch is impossible since we know that 'as' has at least one
+      -- element inside.
```
aspiwack, line 316: But this is not `as`, this is the result of `scanr1`. Which, accordingly, is actually not empty.

utdemir: Right. I was thinking `scanr1` returns a non-empty list because `as` is non-empty; which is indirect. I'll fix the comment.

aspiwack:
It's probably too late to change this bit of API, indeed. Unless we really really want to. It could make some sense, but we'd have to carefully weigh the option. If I remember correctly, the Generic Scans paper gives scans a type like `t a -> (t b, b)`. where the returned `t b` has the same shape as the argument, and the singled out `b` being the sum of the entire structure. Following this example doesn't get us quite the `NonEmpty` list thing.

I guess that we'll probably keep returning lists.

aspiwack (approved):

PR #304, opened 2021-02-16
Add UrT, the unrestricted monad transformer

sjoerdvisscher:
I found this useful, so it could be nice to add this to `linear-base`.


src/Data/Unrestricted/Internal/UrT.hs
```diff
@@ -0,0 +1,43 @@
+{-# LANGUAGE LinearTypes #-}
+
+module Data.Unrestricted.Internal.UrT
```
facundominguez, line 3: The module haddock could be a good place to share some examples where the UrT monad could be useful.

It makes me wonder if we could have a style of interfaces where absolutely all linear resources go in the linear state, and they are only consumed with monadic operations.

sjoerdvisscher: Here's an example where I used it: https://gist.github.com/sjoerdvisscher/03299f8b0c8f208f6239bc75a42b004e#file-unrestrictedmonadtransformer-hs

utdemir (approved):
Thanks!

PR #308, opened 2021-02-17
Add FUN 'One Profunctor instance

sjoerdvisscher:
I'm trying to add support for generating linear traversals to my one-liner package, and noticed this instance is missing. I could use the `LinearArrow` instance instead, but having this instance makes things a bit cleaner.

### DIFF
--- a/src/Control/Optics/Linear/Internal.hs
+++ b/src/Control/Optics/Linear/Internal.hs
@@ -104,8 +104,8 @@ _Nothing = prism (\() -> Nothing) Left
 traversed :: Traversable t => Traversal (t a) (t b) a b
 traversed = Optical $ wander traverse
 
-over :: Optic_ LinearArrow s t a b -> (a %1-> b) -> s %1-> t
-over (Optical l) f = getLA (l (LA f))
+over :: Optic_ (FUN 'One) s t a b -> (a %1-> b) -> s %1-> t
+over (Optical l) f = l f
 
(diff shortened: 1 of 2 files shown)


aspiwack:
Mmm… That's a good point. We must have used the `LinearArrow` newtype when `FUN` was not a thing. I'm kind of wondering if we should remove the `LinearArrow` newtype altogether, since `FUN 'One` does the trick, and that would save us some `coerce`.

Does it make some code less readable? We will have to experiment.

aspiwack (approved):

b-mehta:
> Mmm… That's a good point. We must have used the `LinearArrow` newtype when `FUN` was not a thing. 

Can confirm this is what happened.

sjoerdvisscher:
In this case it is even possible to merge the `(->)` and `FUN 'One` instances into one `FUN m` instance, but that made it too magical imho.

aspiwack:
@b-mehta
thanks for the feedback. I'm thinking we can just kill `LinearArrow` then.

@sjoerdvisscher 
I think the `FUN m` instance may not be super robust to changes in the future. So maybe let's avoid it for now, indeed.

sjoerdvisscher:
I removed the wrapper.

Note that it is now possible to combine `over` and `overU`:
```Haskell
over :: Optic_ (FUN m) s t a b -> (a %m-> b) -> s %m-> t
```

But I guess we should wait for this until we can combine `traverseOf` and `traverseOfU` too.

aspiwack (approved):
Thanks. Feel free to merge.

PR #311, opened 2021-02-18
Add default instances for the Data.Unrestricted classes based on the Generic framework 

sjoerdvisscher:
Just trying to understand the combination of Generics and linearity. Would using `Unsafe.toLinear` on `from` and `to` be the way to go for now?


sjoerdvisscher:
Merged into #316

PR #316, opened 2021-02-25
Add deriving via Generically/Generically1

sjoerdvisscher:
I think the Traversable instances nicely show the difference between traversing with linear control Applicatives and linear data Applicatives. (I'm planning to write a blogpost about this part.)

The linear control Functor generics are a bit more complicated, since you need to prove that the mapping function is used exactly once, which means having `Par1` exactly once. My approach has some runtime overhead, maybe there's a better way?


aspiwack:
I'm not really very comfortable with this, until `Generic` actually exposes linear functions. Maybe we could have deriving via stuff, so that we can hide this capability behind a module named `OmgDangerWillRobinsonHereBeDragons.Generic`. Because it probably mostly works. But is not technically legal.

Also, one quick question: why does `Control.Functor` require a `GFunctor` class, but in `Data.Functor` you seem to reuse the `Functor` class on the representation?

sjoerdvisscher:
Using deriving via is actually a really good idea, I don't like the default signatures way.

Control.Functor needs a separate class because I need to prove that the mapping function is used only once. This doesn't work f.e. for Control.Functor: 

```haskell
instance (Functor f, Functor g) => Functor (f :*: g) where
  fmap f (l :*: r) = fmap f l :*: fmap f r
```

because `f` needs to be used linearly. `GFunctor` threads the `f` through, starting out as `One (a %1-> b)` and switching to `Zero (a %1-> b)` once it is used.

`Par1` is where this happens:
```haskell
instance GFunctor One Zero Par1 where
  gmap (One f) (Par1 a) = (Par1 (f a), Zero)
```

And for the product instance you see the threading happening:
```haskell
instance (GFunctor i m l, GFunctor m o r) => GFunctor i o (l :*: r) where
  gmap f (l :*: r) = gmap @i @m f l & \case
    (l1, f') -> gmap @m @o f' r & \case
      (r1, f'') -> (l1 :*: r1, f'')
```

aspiwack:
Makes sense. I was trying to remember why I've always needed a separate class for the generic thing. I haven't made a generic deriver for quite a while. And maybe it actually was for this very reason of needing extra parameters.

src/Control/Functor/Linear/Internal/Class.hs
```diff
@@ -147,3 +162,28 @@ foldM :: forall m a b. Monad m => (b %1-> a %1-> m b) -> b %1-> [a] %1-> m b
 foldM _ i [] = return i
 foldM f i (x:xs) = f i x >>= \i' -> foldM f i' xs
 
+
+data Zero a = Zero
+data One a = One a
+class GFunctor i o f where
+  gmap :: i (a %1-> b) %1-> f a %1-> (f b, o (a %1-> b))
```
aspiwack, line 169: I'm not super convinced by the state-monad-like threading of `f`, to be honest. It would be preferable to figure out, in a product, where the `f` should go based on some type class magic.

sjoerdvisscher: Back to the drawing board!

sjoerdvisscher:
You also need a separate class for classes of kind `Type -> Constraint`, reusing the same class only works for classes of kind `(Type -> Type) -> Constraint` because GHC.Generics gives everything this `p` argument.

aspiwack (commented):
I approve of the new implementation strategy for deriving control functors. But it needs a little documentation for when we come back there later :slightly_smiling_face: 

Aside from that (and the comments below), may I ask of you to add an example file in the `examples/` directory. Probably with a couple of tests (we have a dedicated test suite for the examples).

src/Control/Functor/Linear/Internal/Class.hs
```diff
@@ -147,3 +156,70 @@ foldM :: forall m a b. Monad m => (b %1-> a %1-> m b) -> b %1-> [a] %1-> m b
...
+  unused U1 = U1
+instance Unused (K1 i v) where
+  unused (K1 c) = K1 c
+instance (Unused l, Unused r) => Unused (l :*: r) where
+  unused (l :*: r) = unused l :*: unused r
+instance (Unused l, Unused r) => Unused (l :+: r) where
+  unused (L1 l) = L1 (unused l)
+  unused (R1 r) = R1 (unused r)
+instance Unused l => Unused (l :.: r) where
+  unused (Comp1 a) = Comp1 (unused a)
```
treeowl, line 192: I don't think this composition instance will actually work. Did you test it?

aspiwack:
Thanks @treeowl for raising this again

I remember that there was an issue with this Pull Request which prevented us from moving forward. But it seems that we failed to document the problem here. @sjoerdvisscher do you remember what happened?

sjoerdvisscher:
The problem is that the generated code by deriving `Generic1` for types having `:.:` in the representation uses regular `fmap`, but it needs to be linear for our use cases. I guess this is also what @treeowl found out.

treeowl:
> The problem is that the generated code by deriving `Generic1` for types having `:.:` in the representation uses regular `fmap`, but it needs to be linear for our use cases. I guess this is also what @treeowl found out.

I wasn't even thinking about that aspect. Where does `fmap` show up in *generated* code? I remember it shows up in a lot of `Generic1`-based instances. The fundamental error in `Generic1` deriving is the way it constructs compositions. It *should* associate them to the left (e.g., `(Rec1 (Either Int) :.: []) :.: Maybe`), but instead associates them to the right (`Either Int :.: ([] :.: Rec1 Maybe)`). Is it possible to patch this up for `linear-base`?

sjoerdvisscher:
> I remember it shows up in a lot of Generic1-based instances.
Yes that's what I meant, the generated instances.

I still don't know what the best way is to do deriving of linear generics. I was hoping to do it with some type level hackery and maybe a minimal amount of TH based on the generated instances of regular generics, but I couldn't really figure that one out. Doing everything in TH seems a daunting task, maybe there's a library that we can hook into? Or else baking it in into GHC might be the easiest approach?

treeowl:
`Generic1` isn't nearly as popular as `Generic`, so I think it's best to start with the latter. But I'll make an attempt at a hacked `Generic1`.

aspiwack:
> `Generic1` isn't nearly as popular as `Generic`, so I think it's best to start with the latter. But I'll make an attempt at a hacked `Generic1`.

True, but this is partly because there are stock derivers for `Functor`/`Foldable`/`Traversable`. Maybe we ought to split the PR to get the easy bits in, though.

treeowl:
Where are the linear `to` and `from` exported? I couldn't figure that out easily from the PR. Yes, I think splitting is a great idea.

treeowl:
Great, as long as we're on the same page.

aspiwack:
This branch is slated for the v2.0 release (next week-ish). @treeowl will take care of updating it to master (when he recovers from his cold). Thanks @treeowl .

treeowl:
@aspiwack I'd really like to merge `Data.Unrestricted.Linear.Internal.Dupable` with `Data.Unrestricted.Linear.Internal.Movable` to get rid of orphan instances. Is that okay with you? My manual "rebase" has now reached the tricky bit (the generic `Dupable` stuff all needs to be redone), and I'd love to finish it up today or this weekend at the latest.

Another question: do we want `Data.Functor` and `Traversable` instances for `Data.Vector.Vector` and `Data.Primitive.Array`? I was a bit surprised not to see those.

treeowl:
Progress continues. I may not need to merge the `Dupable` and `Movable` modules after all. `AsMovable` doesn't have any obvious benefits over generic deriving for basic instances.

One frustrating thing: generic deriving of `Dupable a => Dupable [a]` doesn't work out all that great. The Core is quite nice, with one glaring exception: the whole `dupR` definition becomes a loop breaker and GHC doesn't produce an unfolding. Writing an instance manually, I can use an explicit `go` function with scoped type variables to get what I want. General suggestions welcomed. See #394 for the work in progress.

aspiwack:
> I finally felt well enough today to start the rebasing process. No major roadblocks yet. What `Trustworthy` module should I re-export `GHC.Types.Multiplicity` from so I can use it in (low dependency order) `Safe` modules?

I've never payed much attention `Safe` Haskell, to be honest I don't really know how this works (plus linear-base has a fairly large kernel of trust).

I suggest that we don't care about it for the sake of this PR. And if you want to make linear-base safe-haskell compatible, we discuss how in a second PR.

> @aspiwack I'd really like to merge Data.Unrestricted.Linear.Internal.Dupable with Data.Unrestricted.Linear.Internal.Movable to get rid of orphan instances. Is that okay with you? 

I honestly don't care either way. If it's useful to your PR, feel free. If it's orthogonal, I prefer we do that in a separate step.

> My manual "rebase" has now reached the tricky bit (the generic Dupable stuff all needs to be redone),

Is it worth splitting this to another PR? (I feel my remarks, so far, have been having a theme :slightly_smiling_face:; at the end of the day, do whatever is more expedient).

> Another question: do we want Data.Functor and Traversable instances for Data.Vector.Vector and Data.Primitive.Array? I was a bit surprised not to see those.

I'd say yes. It's an oversight that we haven't (but, you guessed it: it should be a different PR)

> One frustrating thing: generic deriving of Dupable a => Dupable [a] doesn't work out all that great. The Core is quite nice, with one glaring exception: the whole dupR definition becomes a loop breaker and GHC doesn't produce an unfolding. Writing an instance manually, I can use an explicit go function with scoped type variables to get what I want. General suggestions welcomed.

I suggest that we keep the instance for `[a]` manual for the time being, and that we think of ways to improve code generation for this case when the Generic work is merged and we have mental room to focus on the problem.

src/Prelude/Linear/Generically.hs
```diff
@@ -0,0 +1,23 @@
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+module Prelude.Linear.Generically where
+
+-- | A datatype whose instances are defined generically, using the
+-- 'Generics.Linear.Generic' representation.
+-- Generic instances can be derived via @'Generically' A@ using
+-- @-XDerivingVia@.
+newtype Generically a = Generically a
+
```
Topsii, line 10: `Generically` and `Generically1` have been [added to `base-4.17`](https://github.com/ghc/ghc/blob/master/libraries/base/changelog.md). I don't know whether this is relevant for you, but I thought I'd mention it just in case.

aspiwack: It's definitely relevant. I'm not sure how we will handle this, but when 9.4 is out, we will strive to make these newtypes re-exports from base.

PR #317, opened 2021-03-02
Reenable doctests

sjoerdvisscher:
Cabal 3.4.0.0 is out now, so I reenabled the doctest.

But most the doctests are failing with an error like:

```
attempting to use module ‘main:Foreign.Marshal.Pure’ (...) which is not loaded
```

@utdemir Do you know what might cause this?

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -8,7 +8,7 @@ author: Tweag
 maintainer: arnaud.spiwack@tweag.io
 copyright: (c) Tweag Holding and affiliates
 category: Prelude
-build-type: Simple
+build-type: Custom
 synopsis: Standard library for linear types.
 description: Please see README.md.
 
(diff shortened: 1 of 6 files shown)


utdemir:
> @utdemir Do you know what might cause this?

@sjoerdvisscher Unfortunately, I don't remember seeing that error with `cabal-doctests` before, I just had a quick look and couldn't solve it easily.

One interesting thing to note is that, only _some_ of the doctests fail, eg. the test in `Data/Array/Mutable/Linear.hs` seems to be fine. My guess is that it's something about the dependency order, where importing something from above the dependency chain in a test fails. This might be the reason of tests in `Data.V.Linear` failing to import `Prelude.Linear`, but this is not the case `Data.Array.Mutable.Linear`.

So, when I built the `linear-base` and passed the package to `doctests`, it works:

```
$ cabal build
$ cabal exec -- doctest -package linear-base src/**/*.hs
```

However, I don't have any opinion on how we can fix it.

andreabedini:
I am looking at this and FWIW, I seem to be able to succesfully run the doctests in the configured nix shell using cabal-docspec, just as Oleg suggests in the issue linked above.

In that issue Oleg says

> But cabal-docspec doesn't work with stack nor Nix(pkgs) based setup, so it's probably not an option for you.

I am not sure what this means. Given the current setup, I belive the only missing bit is including cabal-docspec in the nix shell.

linear-base.cabal
```diff
@@ -175,38 +175,21 @@ benchmark mutable-data
...
--- library to be built, and that causes a strange type error ("expecting IO,
--- but got IO"), which I suspect because it conflicts with the existing boot
--- packages.
---
--- [1]: https://github.com/haskell/cabal/blob/d28c80acc69b9e7fa992a0b2b7fced937734b238/cabal-install/src/Distribution/Client/ProjectPlanning.hs#L1132-L1149
+custom-setup
+ setup-depends:
+   base >= 4 && <5,
+   Cabal,
+   cabal-doctest
```
facundominguez, line 182: It would be nice having cabal-doctest behind a flag, as packages using `linear-base` don't need doctest to build the dependency.

andreabedini: AFAIK using `cabal-docspec` doesn't require anything in the cabal file. So this particular aspect would be solved.

andreabedini:
Closing in favour of #361

PR #322, opened 2021-04-09
Add a Github action ci

aspiwack:
Now that we don't need to build GHC ourselves for the sake of
linear-base to work, we don't need to lock a super beefy runner for 1h
from time to time.

On the other hand, the buildkite runner didn't run ci for
contributions outside of Tweag. Having a Github-action based ci will
be better for contributors.

I'll remove the Buildkite ci in a separate PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,37 @@
+name: Continuous integration
+on: [push]
+env:
+  # Bump this number to invalidate the Github-actions cache
+  cache-invalidation-key: 0
+jobs:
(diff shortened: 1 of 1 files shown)


aspiwack:
Auto-reviewing because I can. Let me just merge.

aspiwack:
I chose to make a few adjustments before merging, making my previous comment a lie.

- Take the Cabal dependencies to be a separate step (in large part so that we can observe the time it takes)
- Add a test that we are building the Haddock
- Add caching of Cabal dependencies

PR #325, opened 2021-05-12
Update dependencies and remove overrides

utdemir:
Our dependencies has catched up to GHC 9, and we do need any of our forks now.

This PR updates stackage resolver and removes the overrides on cabal.project and stack.yaml.


Divesh-Otwani (approved):
Since the CI passes it looks good to me.

PR #327, opened 2021-05-13
Use safer Vector.fromArray

### LINKED ISSUE #214: Find a safe way to convert an `GHC.Exts.Array#` to a `Data.Vector.Vector`
When converting our mutable array to a vector (https://github.com/tweag/linear-base/pull/205), we end up having a `GHC.Exts.Array#`. Which happens to be te same type `Data.Vector.Vector` is made of, but we can not use it since its constructor is hidden.

Instead, we have to go through a roundabaut way; from `Array#` to `MutableArray#`  to `MVector` to `Vector`; all using some kind of unsafe calls.

https://github.com/tweag/linear-base/blob/c4fec0e13c26041ab614056e2a94fab1b9ad24df/src/Data/Array/Mutable/Linear.hs#L223-L235

Making `Vector`'s constructor visible would solve this easily, so we should consider contributing upstream.

Related upstream issue: https://github.com/haskell/vector/issues/171

utdemir:
Closes #214.

Turns out there is [a recent addition](https://hackage.haskell.org/package/vector-0.12.3.0/docs/Data-Vector.html#v:fromArray) to `vector` library since version `0.12.2.0`:

```
fromArray :: Data.Primitive.Array a -> Vector a
```

This is very close to what we were looking for, so this PR uses that to simplify our `Data.Array.Mutable.Linear.freeze` function.

I also added a lower bound to `vector` package to ensure that we don't get an old one.

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -104,7 +104,7 @@ library
     storable-tuple,
     text,
     transformers,
-    vector,
+    vector >= 0.12.2,
     primitive
   default-language: Haskell2010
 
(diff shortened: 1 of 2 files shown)


aspiwack (approved):
Indeed, this is much better.

PR #329, opened 2021-06-10
Simplify 'seq' for performance

utdemir:
This simple change both looks nicer, and also speeds up
one of my (yet unpublished) local benchmarks considerably.

I think calling `toLinear` is very costly, since the `coerce`
does not inline and the parameter to it ends up being an allocation.
Surely my terminology here is not quite right, and I would love to
know more about how the `coerce` is handled on the simplifier and
core prep, so please do correct me.


aspiwack (approved):
Always happy to remove an `Unsafe.toLinear`!

PR #330, opened 2021-06-11
Remove a 'unsafeCoerce' from 'Unsafe.coerce'

utdemir:
Previously, the linear `Unsafe.coerce` was implemented using two `unsafeCoerce`'s, one to get a function `a -> b`, and another one to cast that function to be a linear one.

After the [unsafe equality proofs change][1], we have a `unsafeEqualityProof` function to cast two types to be the same without looking at the value, so we can linearly return it, removing the need for the outer `unsafeCoerce`.

This removes one allocation from the generated STG code for `Data.Array.Unlifted.get`, improving the performance slightly.

[1]: https://gitlab.haskell.org/ghc/ghc/-/commit/74ad75e87317196c600dfabc61aee1b87d95c214

### DIFF
--- a/src/Unsafe/Linear.hs
+++ b/src/Unsafe/Linear.hs
@@ -1,5 +1,8 @@
 {-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeInType #-}
 {-# LANGUAGE LinearTypes #-}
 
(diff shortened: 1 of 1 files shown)


src/Unsafe/Linear.hs
```diff
@@ -29,8 +32,11 @@ import qualified Unsafe.Coerce as NonLinear
 import GHC.Exts (TYPE, RuntimeRep)
 
 -- | Linearly typed @unsafeCoerce@
-coerce :: a %1-> b
-coerce = NonLinear.unsafeCoerce NonLinear.unsafeCoerce
+coerce :: forall a b. a %1-> b
+coerce a =
+  case NonLinear.unsafeEqualityProof @a @b of
+    NonLinear.UnsafeRefl -> a
+{-# INLINE coerce #-}
```
utdemir, line 39: This `INLINE` actually does not have an effect on my test case, since GHC already was inlining it even without. But I thought it's better to be explicit.

aspiwack (approved):
Neat.

Does it help with the benchmarks?

utdemir:
> Does it help with the benchmarks? [edit: you've already answered this question, sorry, I'm evidently distracted]

I didn't give too much details :). It does help the benchmarks but only a little bit. I saw that there ends up being one less allocation on `Data.Array.Unlifted.Linear.get`, but on my microbenchmark it only ended up being about %5 faster. The only reason I can think of is that, because there still is a `unsafeCoerce` at the same place, we only save one allocation but it doesn't cascade into other optimisations.

Issue #264, opened 2020-11-11
Cabal doesn't play nice with hedgehog


Divesh-Otwani:
**Describe the bug**
Using `cabal test` when there are failed tests sometimes crashes because the log files hedgehog write cannot be read because of the locale in the nix-shell determining the encoding to read with and the locale of the system determining the log file hedgehog writes.

Basically we should get back to just using stack and having a cleaner setup.

Divesh-Otwani:
Referring to the issue we had with stack: #199

utdemir:
> cannot be read because of the locale in the nix-shell determining the encoding to read with and the locale of the system determining the log file hedgehog writes.

I am not entirely sure if this is the case. Because I tried fixing this issue with different locale combinations both by using things like `with-utf8` on Haskell side, and setting my system locale, and it did not help. It definitely seems like a locale issue, but I'm not sure if that is either `hedgehog` or `nix` is doing the wrong thing. It feels like a `cabal-install` issue to me.

> Basically we should get back to just using stack and having a cleaner setup.

I think the common denominator of Haskell projects is using `cabal`, so even if we use `stack` for developing, I'd argue for having a working `cabal-install` setup.

aspiwack:
> I think the common denominator of Haskell projects is using `cabal`, so even if we use `stack` for developing, I'd argue for having a working `cabal-install` setup.

I don't think that this is an accurate description of the situation. We don't do anything special for `cabal` in almost every one of our libraries, I don't see a particular reason why this one would be different.

Libraries are meant to be installed via Hackage, and this will work (stack uses cabal anyway). If you can't run the tests locally with cabal, I don't think we should care.

However, this whole story is very very puzzling. Whether #199 or this one. There doesn't seem to be much sense or reason behind this. And why is linear-based particularly affected? And why does it happen randomly? So many questions!

For the time being, though, I'd argue for whatever is the cheapest, most stable solution.

utdemir:
> I don't think that this is an accurate description of the situation. We don't do anything special for cabal in almost every one of our libraries, I don't see a particular reason why this one would be different.

Oh, I agree with this. Of course if `cabal-install` works if we don't do anything special, we shouldn't do anything special. But if it, for some reason, requires us to do something special in `linear-base`, we should :). Or fix `cabal-install` if it requires us to do something that we shouldn't.

> Libraries are meant to be installed via Hackage, and this will work (stack uses cabal anyway). If you can't run the tests locally with cabal, I don't think we should care.

[...]

PR #332, opened 2021-06-24
Upgrade Stackage nightly and stop using Nix with Stack

utdemir:
(opinionated PR, so please do discuss if you disagree)

The latest Stackage nightly (`2021-06-21`) is the first resolver with GHC 9.0.1, so now `stack` can install the compiler itself, without us having to provide it via Nix.

I gave it a go and saw that on a fresh Ubuntu machine `stack` was able to build `linear-base` from scratch. I think this would be a welcome change for people who wants to tinker on `linear-base` without having to install Nix. We still have the bug in https://github.com/tweag/linear-base/issues/264, so `stack` can not run our tests.

Nix users can still use cabal, or manually pass `--nix` flag to `stack` (or use the global configuration file).

I also created a small CI step to ensure that the project can be built with `stack`.

Since a GHC 9 nightly is out, we should also aim to get `linear-base` into Stackage. I will create a separate issue for that.


.github/workflows/ci.yaml
```diff
@@ -35,3 +35,17 @@ jobs:
...
+    - uses: actions/checkout@v2.3.4
+    - name: Cache dependencies
+      uses: actions/cache@v2.1.6
+      with:
+        path: ~/.stack
+        key: stack-${{ runner.os }}-v${{ env.cache-invalidation-key }}-${{ hashFiles('stack.yaml.lock', 'linear-base.cabal') }}
+    - name: Upgrade stack
+      run: stack upgrade
+    - name: Build
+      run: stack build linear-base:test
```
aspiwack, line 51: This'll be more useful, I think:

```suggestion
      run: stack build --pedantic --test --bench --no-run-tests --no-run-benchmark
```

aspiwack (approved):
Much appreciated.

Small change to the CI action below.

Issue #328, opened 2021-05-15
Optimizing linear arrays


utdemir:
I spent some time benchmarking our linear collections, and wanted to create an issue with my findings.

Reading from our linear arrays seems to almost an order of magnitude slower than reading from a `vector`.

I benchmarked these two functions:

```haskell
   linear :: Array.Linear.Array Int %1-> ()
   linear hm =
     hm
       Linear.& Array.Linear.size
       Linear.& \(Linear.Ur sz, arr) -> arr
       Linear.& go 0 sz
    where
     go :: Int -> Int -> Array.Linear.Array Int %1-> ()
     go start end arr
       | start < end =
           Array.Linear.unsafeGet start arr
             Linear.& \(Linear.Ur i, arr) -> i `Linear.seq` go (start + 1) end arr
       | otherwise = arr `Linear.lseq` ()

   dataVector :: Data.Vector.Vector Int -> ()
   dataVector v =
     let sz = Data.Vector.length v
      in go 0 sz
    where
     go :: Int -> Int -> ()
     go start end
       | start < end =
           (v Data.Vector.! start) `seq`  go (start + 1) end
       | otherwise = ()
```

And here's the result on my system (this also includes allocating the array/vector with an initial element, but I measured that that is not dominating the runtime):

```
benchmarked arrays/reads/Data.Array.Mutable.Linear
time                 27.26 ms   (26.39 ms .. 28.05 ms)
                     0.997 R²   (0.996 R² .. 0.999 R²)
mean                 29.13 ms   (28.44 ms .. 30.28 ms)
std dev              1.972 ms   (1.178 ms .. 2.606 ms)
variance introduced by outliers: 26% (moderately inflated)

benchmarked arrays/reads/Data.Vector
time                 3.841 ms   (3.796 ms .. 3.887 ms)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 3.971 ms   (3.943 ms .. 4.011 ms)
std dev              107.8 μs   (80.19 μs .. 173.3 μs)
```

When looking at the relevant `-ddump-simpl` output (with `-O`), it can be seen that the generated code of `Data.Vector` is a pretty efficient loop:

```haskell
$wbReads1_raMu
  :: GHC.Prim.Int# -> GHC.Prim.Int# -> GHC.Prim.Array# Int -> ()
[GblId, Arity=3, Str=<L,U><L,U><L,U>, Unf=OtherCon []]
$wbReads1_raMu
[...]

PR #334, opened 2021-06-27
Optimize Data.Array.Mutable.Unlifted.Linear.map

utdemir:
This PR implements `map` function in the PR in a simpler way, directly on top of `Array#`'s. It's as fast as the `map` function on `Data.Vector` now.

In order to be able to measure the results, I also split up a benchmark using `map` and `toList` to two benchmarks each testing a single function.

Without `{-# NOINLINE map #-}` our benchmarks improve a bit. However, I'm keeping it as per the discussion around https://github.com/tweag/linear-base/pull/187#discussion_r489204108. 

Related: https://github.com/tweag/linear-base/issues/328

### DIFF
--- a/src/Data/Array/Mutable/Unlifted/Linear.hs
+++ b/src/Data/Array/Mutable/Unlifted/Linear.hs
@@ -129,21 +129,27 @@ copyInto start@(GHC.I# start#) = Unsafe.toLinear2 go
 {-# NOINLINE copyInto #-}  -- prevents the runRW# effect from being reordered
 
 map :: (a -> b) -> Array# a %1-> Array# b
-map (f :: a -> b) arr =
-  size arr
-    `chain2` \(# Ur s, arr' #) -> go 0 s arr'
- where
-  -- When we're mapping an array, we first insert `b`'s
-  -- inside an `Array# a` by unsafeCoerce'ing, and then we
-  -- unsafeCoerce the result to an `Array# b`.
-  go :: Int -> Int -> Array# a %1-> Array# b
-  go i s arr'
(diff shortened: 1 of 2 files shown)


src/Data/Array/Mutable/Unlifted/Linear.hs
```diff
@@ -129,21 +129,27 @@ copyInto start@(GHC.I# start#) = Unsafe.toLinear2 go
...
+      -- For each index ([0..len]), we read the element on 'as', pass
+      -- it through 'f' and write to the same location on 'bs'.
+      go :: GHC.Int# -> GHC.State# GHC.RealWorld -> ()
+      go i st
+        | GHC.I# i Prelude.== GHC.I# len = ()
+        | Prelude.otherwise =
+          case GHC.readArray# as i st of
+            (# st', a #) ->
+              case GHC.writeArray# bs i (f a) st' of
+                st'' -> go (i GHC.+# 1#) st''
```
aspiwack, line 150: This worries me a little. It may work because of how `State#` is not a lifted type, but `case` are lazy in the surface language. Would the following work instead, to appease my worries?

```suggestion
              case GHC.writeArray# bs i (f a) st' of
                !st'' -> go (i GHC.+# 1#) st''
```

aspiwack (approved):

src/Data/Array/Mutable/Unlifted/Linear.hs
utdemir: This makes sense. I don't see any difference on STG output, but better to be safe than sorry.

PR #337, opened 2021-07-12
Various HashMap optimisations

utdemir:
Just some minor optimisations for our HashMap implementation. It contains
a few unrelated small changes, so reviewing each individual commit might
be easier.

When I compare this against `Data.HashMap.Strict`, the results change based
on the input (likely based on the hash distribution of the key). I need to investigate
this more. Just as reference, here're the functions I am comparing (`Key` is just a
newtype wrapper around `Int`):

```
   linear :: [Key] -> HashMap.Linear.HashMap Key () %1-> ()
   linear inp hm = go inp hm `Linear.lseq` ()
    where
     go :: [Key] -> HashMap.Linear.HashMap Key () %1-> HashMap.Linear.HashMap Key ()
     go [] h = h
     go (x:xs) h = go xs Linear.$! HashMap.Linear.insert x () h

   dataHashMap :: [Key] -> Data.HashMap.Strict.HashMap Key () -> ()
   dataHashMap inp hm = go inp hm `seq` ()
    where
     go :: [Key] -> Data.HashMap.Strict.HashMap Key () -> Data.HashMap.Strict.HashMap Key ()
     go [] h = h
     go (x:xs) h = go xs $! Data.HashMap.Strict.insert x () h
```

I did not include the benchmark on this PR, since it looks a bit different
than our existing HashMap benchmarks. I have to combine them together at
one point.

### DIFF
--- a/src/Data/HashMap/Mutable/Linear/Internal.hs
+++ b/src/Data/HashMap/Mutable/Linear/Internal.hs
@@ -87,7 +87,7 @@ type RobinArr k v = Array (Maybe (RobinVal k v))
 
 -- | Robin values are triples of the key, value and PSL
 -- (the probe sequence length).
-data RobinVal k v = RobinVal {-# UNPACK #-} !PSL k v
+data RobinVal k v = RobinVal !PSL !k v
   deriving (Show)
 
 incRobinValPSL :: RobinVal k v -> RobinVal k v
--- a/src/Data/HashMap/Mutable/Linear/Internal.hs
+++ b/src/Data/HashMap/Mutable/Linear/Internal.hs
@@ -164,7 +164,7 @@ alterF f key hm =
           Ur (Just v)->
             HashMap
              (count+1)
-             (Array.write arr ix (Just (RobinVal psl key v)))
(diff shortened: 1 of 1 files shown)


aspiwack (approved):
I guess that the next step is committing these benchmarks.

facundominguez:
Hello! Just a random thought, are `alter` and `alterF` supposed to be inlined? Otherwise, the allocation of the `Maybe` value returned by the callback can't be eliminated. `insert` in `unordered-containers` isn't implemented in terms of alter, at least.

utdemir:
>  are alter and alterF supposed to be inlined? Otherwise, the allocation of the Maybe value returned by the callback can't be eliminated. 

This was a good point, I think they should be inlined. I can observe from the core output that both the callback passed to `alterF` from `alter` and eg. callback pased to `alter` from `insert` disappear. Unfortunately they do not correspond to a measurable speed increase on the insertion benchmarks, I think this is because they are only called once per insertion, but our bottleneck is on `tryInsertAtIndex` and `probeFrom` functions which are called multiple times per insertion.

But they make sense and make the core output prettier, so I commited those.

Issue #312, opened 2021-02-18
Using a mutable backing store (such as Array) for V


ekmett:
**Is your feature request related to a problem? Please describe.**

`dupV` can produce `V n a` values efficiently for many types. 

This privileges this over other possible linear storage solutions for bulk copied data.

I was hoping to work with an API that takes a `V n a` and uses reflection to package it up as a `Range` of values you can access that shares a common backing store. Ranges themselves are mostly type level constructs, the "real" `Vector Any` gets held by reflection in common, thus ensuring never shall two values of type `Range s i j _` will exist at the same time, letting you work on privately owned data, and two adjacent ranges with the same `s` can be safely re-merged.

One advantage of this over just working with `V n a`s is that I can split work up into sections, worksteal and reassemble my answers safely if the ranges show up in the types, and the original backing store identity is captured in a region parameter.

```haskell
claim    
  :: forall n a b. V n a %1
  -> (forall s. (KnownV s, VLength s ~ n) => Range s 0 n a %1 -> Ur b) %1    
  -> Ur b

merge :: forall s i j k a. Range s i j a %1 -> Range s j k a %1 -> Range s i k a
split ::: ...
declaim :: forall s i j a. KnownV s => Range s i j a %1 -> V (j - i) a
```

which all works great, right up until realize you can't write `Linear.Functor`, or lenses to the individual members of the range, because unlike an Array, the contents of `V n` aren't actually mutable.

**Describe the solution you'd like**
I'd like to look under the hood of `V` and see something that is either a `Array#` or `MVector`, so that lenses to modify the ith member can be written, that don't require complete copies, and linear fmap can be implemented in place.

**Describe alternatives you've considered**
One thought was to use unsafe thaw operations but there isn't a strong guarantee that the vector you have is the only one using its backing array, and freezing it on one and having another segment of it thinking it was thawed on the other would be bad from a GC perspective as I understand things. moving things to a fresh mutable vector on `claim` above, also seems needless and painful.

aspiwack:
I do like the `Range` design :slightly_smiling_face: .

I think that the trade-off, here, is whether we want `V n a` to be a mutable data structure under the hood. Or if we want `V n a` to have a `Movable` instance (which I've noticed you haven't raised in #310 , so I assume you've chosen your side there :slightly_smiling_face: ).

I've stayed away from this design choice, because I don't really have strong arguments either way.
- Having `Movable` would make `V` more like tuples, which is sort of my intuition of them
[...]

PR #341, opened 2021-08-19
Introduce Data.VM.Linear and Data.LArray.Mutable.Unlifted.Linear

utdemir:
Relevant #312 .

This PR introduces a new mutable linear array type called `LArray` (`Data.LArray.Mutable.Unlifted.Linear`), which contains linear elements. And it uses that to implement a `VM` type, which is our `V` on a mutable backing store.

There're many design decisions, so I decided to create a draft PR before working more on this.

The new `Data.LArray.Mutable.Unlifted.Linear` is mostly a copy of `Data.Array.Mutable.Unlifted.Linear`; with changes to preserve the linearity of the elements:

* It is meant to be used with the `update` function; and `get` and `set` functions require `Dupable` and `Consumable` constraints.
* Functions like `lseq` and `dup2` over arrays consumes or duplicates the elements one by one.

And we have `VM` implemented as a length-indexed wrapper over the `LArray`. It uses the type-level utilities from the original `V` module and provides a similar `make` and `elim` functions; the main difference being `make` function taking a linear callback.

And here is the missing parts (hence this is yet a draft):

* Comments on whether this is useful (having `Data.Array.Mutable.Unlifted.Linear`, a separate `VM` type).
* There's a bunch of unsafe code, but no tests. We should have some.
* We have to decide whether we should use the new `VM` on the `Dupable` class. I feel like it would be a good default (based on #312), but we have to have an efficient Applicative instance.
* It'd be good to try out our lenses here.

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -39,6 +39,8 @@ library
     Data.Array.Mutable.Linear
     Data.Array.Mutable.Linear.Internal
     Data.Array.Mutable.Unlifted.Linear
+    Data.LArray.Mutable.Unlifted.Linear
+    Data.LArray.Mutable.Unlifted.Linear.Internal
     Data.Array.Polarized
     Data.Array.Polarized.Pull
     Data.Array.Polarized.Pull.Internal
@@ -78,6 +80,9 @@ library
     Data.V.Linear
     Data.V.Linear.Internal.V
     Data.V.Linear.Internal.Instances
+    Data.VM.Linear
+    Data.VM.Linear.Internal.VM
+    Data.VM.Linear.Internal.Instances
     Data.Vector.Mutable.Linear
     Data.Vector.Mutable.Linear.Internal
     Debug.Trace.Linear
--- a/src/Data/LArray/Mutable/Unlifted/Linear.hs
+++ b/src/Data/LArray/Mutable/Unlifted/Linear.hs
@@ -1,0 +1,29 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE MagicHash #-}
+
+-- |
+-- This module provides an unlifted mutable array with a pure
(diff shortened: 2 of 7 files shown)


aspiwack:
> We have to decide whether we should use the new `VM` on the `Dupable` class. I feel like it would be a good default (based on

What would be the type of `dupVM`?

utdemir:
> What would be the type of dupVM?

Very good point; it can not be `a #1-> VM n a` because then we can get an unrestricted `VM`. So it has to be something like `a #-> (VM n a #-> Ur a) #-> Ur a`, which is not pretty (and likely more restrictive than the curent `dupV`). 

For a similar reason, something like a `:: V n a #1-> VM n a`  is also not possible.

What do you think?

utdemir:
@tbagrel1 Do you think this can still be useful, especially after #360? Please feel free to close this otherwise :).

tbagrel1:
I'm not a user of `linear-base` myself at this point, so I don't really feel legitimate to say what we need and what we don't need 😅. But right now I don't see a strong point for adding `VM`. Compile-time-known-sized array are somewhat unpractical to deal with, especially when they are mutable. So I think that people won't really use them for large `n`, which is when having a mutable store starts to bring performance improvements (maybe for cryptographic applications?). `V n a` behaves rather like a n-ary tuple given its API at this point.

However, I think that `LArray` is a more basic type that is really missing in `linear-base`. We definitely want to have these functions:

```haskell
LArray.fromReplicator :: Int -> Replicator a %1 -> LArray a

LArray.fill :: Dupable a => Int -> a %1 -> -> LArray a
LArray.fill n = fromReplicator n . dupR
```
that we couldn't write for `Data.Array.Mutable.Linear`

_____
Anyway, I think that it would be interesting to add some Array-like memo/2-axis table to #376 #368

PR #344, opened 2021-09-13
Add linear arrows

oliverbunting:
Propose definitions for `Control.Category.Linear` and  `Control.Arrow.Linear`

Happy to discuss / modify.

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -22,6 +22,7 @@ library
   hs-source-dirs: src
   exposed-modules:
     Control.Monad.IO.Class.Linear
+    Control.Category.Linear
     Control.Functor.Linear
     Control.Functor.Linear.Internal.Class
     Control.Functor.Linear.Internal.Instances
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -1,5 +1,5 @@
 name: linear-base
-version: 0.1.1
+version: 0.1.2
 cabal-version: >=1.10
 homepage: https://github.com/tweag/linear-base#README
 license: MIT
@@ -22,6 +22,7 @@ library
   hs-source-dirs: src
   exposed-modules:
     Control.Monad.IO.Class.Linear
+    Control.Arrow.Linear
     Control.Category.Linear
     Control.Functor.Linear
     Control.Functor.Linear.Internal.Class
--- a/src/Control/Arrow/Linear.hs
+++ b/src/Control/Arrow/Linear.hs
@@ -1,0 +1,128 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Control.Arrow.Linear (
+  -- | Arrows
+  Arrow(..),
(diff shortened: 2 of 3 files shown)


oliverbunting:
@aspiwack thanks for a detailed reply. I'll confess most of that went by me - both undergrad and research is electronic engineering, and whilst I've used Haskell professionally for the best part of a decade, I've no theoretical background.

---

To respond to your aside, and possibly explain my motivations better:

Let me crudely outline my thoughts on circuit representations. It starts with the observation that in most systems the connections between circuits are bidirectional - circuits communicate using busses - in CMOS with dedicated routing master to slave in each direction (not shared routing resources). 

Therefore, if `Circuits` take and produce instances of `Bus` as arguments which are glorified functions, linear arrows can be used to define the connectivity between functions. Working with the more restrictive version of arrow is basically essential here, because you want to ensure a bus is consumed exactly once. Circuits with hanging busses are probably a design mistake, whilst reusing a bus implies either hardware duplication, or multiply driven busses, both of which don't want to happen by accident.

You can do _most_ of this with linear `Control.Monad`, but linear `MonadFix` is a none starter. However, linear `ArrowLoop` allows you to define recursive connectivity.

My thoughts are publicly visible, although incomplete and possibly flawed, [here](https://github.com/oliverbunting/clash-circuit/blob/bc18821c8904689c1c12ac8312e1a3295d4f7211/src/Clash/Circuit/Bus.hs#L65) if you're interested. I intend to finish that work, it still interests me, but free time is at a premium, so it's happening very slowly. I'd be happy to discuss, but there's got to be a better forum than this PR.

Whilst doing all this a number of years ago, we basically used a source plugin hack to workaround the lack of linear types. An ex-colleague of mine, and one of those (along with christiaanb, martijnbastiaan, JonFowler) who've spent countless hours discussing circuit representations with me, has published that [here](https://github.com/cchalmers/circuit-notation). I believe its still used in production.

aspiwack:
Hello again,

Sorry I'm taking so much time to answer. It's been busy times, I'm afraid.

> I'll confess most of that went by me - both undergrad and research is electronic engineering, and whilst I've used Haskell professionally for the best part of a decade, I've no theoretical background.

It's alright, I suppose this was mostly for my personal consumption :slightly_smiling_face: .

> Therefore, if `Circuits` take and produce instances of `Bus` as arguments which are glorified functions, linear arrows can be used to define the connectivity between functions. Working with the more restrictive version of arrow is basically essential here, because you want to ensure a bus is consumed exactly once.

Ok, so, do I understand right that, in this scenario, a `bus` is a value of type `Circuit A B`, and we have `Arrow Circuit`?

> My thoughts are publicly visible, although incomplete and possibly flawed, [here](https://github.com/oliverbunting/clash-circuit/blob/bc18821c8904689c1c12ac8312e1a3295d4f7211/src/Clash/Circuit/Bus.hs#L65) if you're interested.

It is a bit over my head, I'm afraid.

> I'd be happy to discuss, but there's got to be a better forum than this PR.

I'm available for a call if you wish. You can send me an email. First name dot last name @tweag.io .

PR #346, opened 2021-10-04
Add toLinearN; narrow the coercion scope

treeowl:
* Make `toLinear`, etc., explicitly coerce only *multiplicities*,
rather than whole arrow types. This is just a cleanup; it shouldn't
have any practical impact.

* Add `toLinearN` and `toLinearL` to generalize `toLinear`, `toLinear2`, and `toLinear3`.

### DIFF
--- a/src/Unsafe/Linear.hs
+++ b/src/Unsafe/Linear.hs
@@ -37,10 +37,9 @@ module Unsafe.Linear
   where
 
 import Data.Type.Equality (type (~~))
-import Data.Kind (Type, Constraint)
+import Data.Kind (Constraint)
 import Unsafe.Coerce (UnsafeEquality (..), unsafeEqualityProof)
 import GHC.Exts (TYPE, RuntimeRep (..))
-import GHC.Types (Multiplicity (..))
 import GHC.TypeNats
 
 -- | Linearly typed @unsafeCoerce@
@@ -53,26 +52,26 @@ coerce a =
 -- | Converts an unrestricted function into a linear function
 toLinear
   :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep)
-     (a :: TYPE r1) (b :: TYPE r2) p.
-     (a %p-> b) %1-> (a %1-> b)
-toLinear f = case unsafeEqualityProof @p @'One of
+     (a :: TYPE r1) (b :: TYPE r2) p x.
+     (a %p-> b) %1-> (a %x-> b)
+toLinear f = case unsafeEqualityProof @p @x of
   UnsafeRefl -> f
 
 -- | Like 'toLinear' but for two-argument functions
 toLinear2
   :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep) (r3 :: RuntimeRep)
-     (a :: TYPE r1) (b :: TYPE r2) (c :: TYPE r3) p q.
-     (a %p-> b %q-> c) %1-> (a %1-> b %1-> c)
-toLinear2 f = case unsafeEqualityProof @'(p,q) @'( 'One, 'One) of
+     (a :: TYPE r1) (b :: TYPE r2) (c :: TYPE r3) p q x y.
+     (a %p-> b %q-> c) %1-> (a %x-> b %y-> c)
+toLinear2 f = case unsafeEqualityProof @'(p,q) @'(x, y) of
   UnsafeRefl -> f
 
 -- | Like 'toLinear' but for three-argument functions
 toLinear3
   :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep)
      (r3 :: RuntimeRep) (r4 :: RuntimeRep)
-     (a :: TYPE r1) (b :: TYPE r2) (c :: TYPE r3) (d :: TYPE r4) p q r.
-     (a %p-> b %q-> c %r-> d) %1-> (a %1-> b %1-> c %1-> d)
-toLinear3 f = case unsafeEqualityProof @'(p,q,r) @'( 'One, 'One, 'One) of
+     (a :: TYPE r1) (b :: TYPE r2) (c :: TYPE r3) (d :: TYPE r4) p q r x y z.
+     (a %p-> b %q-> c %r-> d) %1-> (a %x-> b %y-> c %z-> d)
+toLinear3 f = case unsafeEqualityProof @'(p,q,r) @'(x,y,z) of
   UnsafeRefl -> f
 
 -- | @toLinearN@ subsumes the functionality of 'toLinear1', 'toLinear2', and
--- a/src/Unsafe/Linear.hs
+++ b/src/Unsafe/Linear.hs
@@ -1,10 +1,11 @@
-{-# LANGUAGE KindSignatures #-}
-{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE GADTs #-}
-{-# LANGUAGE TypeApplications #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeInType #-}
+{-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 
 -- | Unsafe coercions for linearly typed code.
 --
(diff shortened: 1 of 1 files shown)


treeowl:
Should we make at least some of these functions more polymorphic? I would think something like

```haskell
toLinear
  :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep)
     (a :: TYPE r1) (b :: TYPE r2) p m x.
     (a %p-> b) %m-> (a %x-> b)

toLinear2
  :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep) (r3 :: RuntimeRep)
     (a :: TYPE r1) (b :: TYPE r2) (c :: TYPE r3) p q m x y.
     (a %p-> b %q-> c) %m-> (a %x-> b %y-> c)

toLinear3
  :: forall (r1 :: RuntimeRep) (r2 :: RuntimeRep)
     (r3 :: RuntimeRep) (r4 :: RuntimeRep)
     (a :: TYPE r1) (b :: TYPE r2) (c :: TYPE r3) (d :: TYPE r4) p q r m x y z.
     (a %p-> b %q-> c %r-> d) %m-> (a %x-> b %y-> c %z-> d)
```

This doesn't do much for `toLinear`, but for `toLinear2` and `toLinear3` I imagine something like this would be a considerable improvement.

treeowl:
Since `toLinearN` is actually capable of changing multiplicities in any and all directions, should it have a different name?

treeowl:
Might we want a version that takes a full list of multiplicity changes? Something like this?

```haskell
toLinearList @'[ 'One :=> 'Many, _ :=> 'One, 'Many :=> n] ...
```

src/Unsafe/Linear.hs
```diff
@@ -25,37 +32,125 @@ module Unsafe.Linear
...
+-- multiplicities of the first @n@ arrows from any multiplicity to any
+-- other multiplicity.
+--
+-- == Examples
+--
+-- @
+-- toLinearN \@2 :: (a -> b -> Int) -> a %1-> b %1-> Int
+-- 'toLinear3' = toLinearN \@3
+-- @
+toLinearN :: forall n f g m. ToLinearN n f g => f %m-> g
```
treeowl, line 97: I don't know the conventions; is it better to use `%m` here or `%1`?

aspiwack: We tend to use `%m` for functions which are meant to be compatible with an existing library (typically base; things like `const`) and %1 otherwise. Especially for a function like this which we never intend to pass as an argument for a function.

src/Unsafe/Linear.hs
```diff
@@ -20,42 +27,188 @@
...
+-- number it's handed.
+type ToLinearN' :: forall {rep :: RuntimeRep}. INat -> TYPE rep -> TYPE rep -> Constraint
+class ToLinearN' arrs f g where
+  prf :: UnsafeEquality f g
+instance ToLinearN' k fb gb => ToLinearN' ('S k) ((a :: TYPE repa) %p-> (fb :: TYPE repb)) (a %q-> (gb :: TYPE repb)) where
+  prf = case prf @k @fb @gb of
+    UnsafeRefl -> case unsafeEqualityProof @p @q of
+      UnsafeRefl -> UnsafeRefl
+instance ToLinearN' 'Z (a :: TYPE rep) (a :: TYPE rep) where
+  prf = UnsafeRefl
```
aspiwack, line 162: Does this produce an O(n) irreducible proof? I'm not super up to date with the new `unsafeEqualityProof` mechanism.

And if so: is it bad?

src/Unsafe/Linear.hs
```diff
@@ -20,42 +27,188 @@
...
+    ( UnifyL n b (GetArrowRes g :: TYPE repb)
+    , g ~ (a %(GetArrowMult g)-> (GetArrowRes g :: TYPE repb)))
+
+  -- This is a bit unsatisfying. Can we use the fact that the count is positive
+  -- to *infer* that there's another arrow, squeezing out one last bit of
+  -- inference? The representation polymorphism seems to make this a bit
+  -- tricky, but maybe it's possible. Whether it's worth the trouble is another
+  -- question. There's probably not a huge practical benefit.
+  UnifyL ('S _) (_ :: TYPE rep) g =
+    TypeError ('Text "unsafeLinearityProofN: not enough arrows to satisfy request.")
```
aspiwack, line 185: I don't think that `UnifyL` is necessary. You can simply phrase your `S` instance of `ToLinearN'` to force unification.

Something like
```haskell
 instance (ToLinearN' k fb gb, x~((a :: TYPE repa) %p-> (fb :: TYPE repb)), y~(a %q-> (gb :: TYPE repb)) => ToLinearN' ('S k) x y where
```

src/Unsafe/Linear.hs
```diff
@@ -20,42 +27,188 @@
...
+-- as described by the list @l@.
+type ToLinearL :: forall {rep :: RuntimeRep}. [Change] -> TYPE rep -> TYPE rep -> Constraint
+class ToLinearL l f g where
+  -- | @unsafeLinearityProofL \@l \@f \@g @ is a fake proof that @f@ and @g@ are
+  -- the same with the possible exception of the multiplicities of the first
+  -- arrows given by the list @l@.
+  unsafeLinearityProofL :: UnsafeEquality f g
+instance ( ToLinearN' ni f g, UnifyL ni f g, UnifyL ni g f
+         , ni ~ Length l, UseChanges l f g) => ToLinearL l f g where
+  unsafeLinearityProofL = prf @ni
```
aspiwack, line 142: Is this really necessary? This can be achieved with a `ToLinearN` and a type annotation. I think that I'm essentially questioning the need for exporting an operator from this `Unsafe` module.

aspiwack (commented):
I have a bunch of questions below. But I'm favourable to the change as a whole. As well as making the `toLinearX` functions returning functions of an arbitrary multiplicity (though maybe their name isn't ideal anymore, but it's not too serious).

I've got to admit, I'm feeling a bit of a mix of amazement and dismay that the `Unsafe` module is seeing such love :slightly_smiling_face: .

aspiwack:
@treeowl there appears to be an issue with CI. But I think I've fixed it in #348 , so you just need to rebase this PR and CI should start.

src/Unsafe/Linear.hs
treeowl: I'll try it.

src/Unsafe/Linear.hs
treeowl: The proof can get a bit long, but I don't think that we can do anything about it. It'll be erased in Core Prep or so.

src/Unsafe/Linear.hs
treeowl: Okay, I guess I'll make them unconditionally linear.

treeowl:
@aspiwack, I implemented your suggestions. Your suggested approach to `ToLinearN'` (with a heterogeneous equality tweak) is much better than what I was doing before. I also generalized the `toLinear`, `toLinear2`, and `toLinear3` in a separate commit. I hope that doesn't wreak too much havoc on inference.

treeowl:
Commits are now squashed to logical units.

src/Unsafe/Linear.hs
```diff
@@ -20,42 +26,138 @@
...
+-- @y@ are lifted types, rather than needing that information to come from
+-- elsewhere.
+instance
+  ( ToLinearN' k fb gb
+  , x ~~ ((a :: TYPE repa) %p-> (fb :: TYPE repb))
+  , y ~~ (a %q-> (gb :: TYPE repb))
+  ) => ToLinearN' ('S k) (x :: TYPE rep) (y :: TYPE rep) where
+  prf = case prf @k @fb @gb of
+    UnsafeRefl -> case unsafeEqualityProof @p @q of
+      UnsafeRefl -> UnsafeRefl
```
aspiwack, line 163: This is just 
```suggestion
    UnsafeRefl -> unsafeEqualityProof @p @q
```

Isn't it?

src/Unsafe/Linear.hs
```diff
@@ -20,42 +26,138 @@
...
+  case unsafeEqualityProof @a @b of
+    UnsafeRefl -> a
 {-# INLINE coerce #-}
 
+-- TODO: In GHC 9.0, using toLinearN to implement toLinear2 and toLinear3
+-- produces more Core terms, though fewer types and coercions. Does this hamper
+-- inlining? I imagine it might. I suspect this is still the case for 9.2, but
+-- I haven't checked. If we want, we might be able to reduce
+-- unsafeLinearityProofN to one case expression by using some (quantified
+-- constraint?) tricks.
```
aspiwack, line 57: This would really be nice but I don't see how to do this at the moment. And I'm not sure this deserves to spend more time on.

aspiwack (approved):
Let's address the small comment below, then when you are happy, I'll merge.

src/Unsafe/Linear.hs
treeowl: No, it isn't. We're looking at types

```haskell
a %p-> r1
-- and
a %q-> r2
```

We use a recursive call to get `UnsafeEquality r1 r2`, and we use `unsafeEqualityProof` to get `UnsafeEquality p q`. We need `UnsafeEquality (a %p-> r1) (a %q-> r1)`, and we can't get that by substitution into either of the proofs we have; we need to match on both to build it.

src/Unsafe/Linear.hs
treeowl: Yeah, not planning to touch it any more for now. Sliding back toward the `linear-generics` thing.

treeowl:
Oh, I guess I should remove the word TODO...

src/Unsafe/Linear.hs
aspiwack: Indeed, thanks.

treeowl:
OK, now it's really done.

aspiwack (commented):
Fantastic. Thanks a lot!

CI is green, which does mean that generalising the types of the unsafe operations hasn't been a problem in practice.

Let me merge now.

@utdemir it would be interesting to see if this PR has any effect on your benchmarks (I don't expect so, but you never know).

PR #349, opened 2021-10-12
Add genericTraverse

treeowl:
`genericTraverse` is an efficient implementation of `traverse` for
any `Generic1` type.


treeowl:
Thanks, @utdemir!

tbagrel1:
I think that we can close it now that #366 is merged.

aspiwack:
Yes. Let's close thank you both!

PR #355, opened 2022-01-11
Improve shell.nix, add niv, ormolu, and improve CI

tbagrel1:


format.sh
```diff
@@ -0,0 +1,15 @@
...
+set -e
+
+export LANG="C.UTF-8"
+
+cabal format
+
+ormolu -m inplace $(find src -type f -name "*.hs-boot" -o -name "*.hs")
+ormolu -m inplace $(find test -type f -name "*.hs-boot" -o -name "*.hs")
+ormolu -m inplace $(find examples -type f -name "*.hs-boot" -o -name "*.hs")
+ormolu -m inplace $(find bench -type f -name "*.hs-boot" -o -name "*.hs")
```
aspiwack, line 15: Any reason why this isn't 
```suggestion
ormolu -m inplace $(find . -type f -name "*.hs-boot" -o -name "*.hs")
```
If so, please document.

aspiwack: In case there is a good reason to list all these directories, it's probably better to do
```suggestion
ormolu -m inplace $(find src test example bench -type f -name "*.hs-boot" -o -name "*.hs")
```

tbagrel1: There is no reason not to do what you suggested

aspiwack (approved):

tbagrel1:
> Why did you remove the restore-key?

I didn't see that the `key` and `restore-key` were different, my bad. By the way, what is `github.sha` at the end of the primary `key`?

.github/workflows/ci.yaml
```diff
@@ -3,49 +3,73 @@ on: [push, pull_request]
...
+    - name: Cache Stack dependencies
+      uses: actions/cache@v2
+      with:
+        path: ~/.stack
+        key: stack-deps-${{ runner.os }}-${{ hashFiles('nix/sources.json') }}-v${{ env.cache-invalidation-key }}-${{ hashFiles('stack.yaml.lock', 'linear-base.cabal') }}-${{ github.sha }}
+        restore-keys: stack-deps-${{ runner.os }}-${{ hashFiles('nix/sources.json') }}-v${{ env.cache-invalidation-key }}-${{ hashFiles('stack.yaml.lock', 'linear-base.cabal') }}-
+    - name: Build Nix dependencies
+      run: nix-shell --pure --run "echo '=== Nix dependencies installed ==='"
+    - name: check formatting
+      run: nix-shell --pure --run 'stack build ormolu && stack exec ormolu -- -m check $(find . -type f -name "*.hs-boot" -o -name "*.hs")'
```
aspiwack, line 56: Did you check that the Ormolu build was cached?

.github/workflows/ci.yaml
```diff
@@ -3,49 +3,73 @@ on: [push, pull_request]
...
+    - uses: actions/checkout@v2
+    - uses: cachix/install-nix-action@v15
+      with:
+        nix_path: "${{ env.nixpkgs-url }}"
+    - name: Cache Stack dependencies
+      uses: actions/cache@v2
+      with:
+        path: ~/.stack
+        key: stack-deps-${{ runner.os }}-${{ hashFiles('nix/sources.json') }}-v${{ env.cache-invalidation-key }}-${{ hashFiles('stack.yaml.lock', 'linear-base.cabal') }}-${{ github.sha }}
+        restore-keys: stack-deps-${{ runner.os }}-${{ hashFiles('nix/sources.json') }}-v${{ env.cache-invalidation-key }}-${{ hashFiles('stack.yaml.lock', 'linear-base.cabal') }}-
```
aspiwack, line 52: I think that if both the caches are called the same they will be shared, which probably won't have the intended effect. It's probably best to give this one a different name.

aspiwack (approved):

.github/workflows/ci.yaml
tbagrel1: I wanted to share the cache in that case. I didn't see reasons not to share it, but maybe there are?

.github/workflows/ci.yaml
tbagrel1: I can't really check that while the ormolu invocation fails

.github/workflows/ci.yaml
tbagrel1: Well, in fact, you are right, it's better to separate the keys (otherwise, because the jobs are executed in parallel, and because ormolu is the slowest at the moment, the "stack build" job write its cache first, and the ormolu job then cannot upload its own cache)

PR #356, opened 2022-01-11
Changing (>>) in monad to accept consumable first inputs

saolof:
Currently (>>) has type m () %1-> m b %1-> m b .

This can easily be generalized to (Consumable a) =>  m a %1-> m b %1-> m b  .

This change would make (>>) more similar to the nonlinear prelude and much more ergonomic to work with in effectful code. But it would technically break backwards compatibility for anyone who explicitly implemented (>>) on their types by making it more constraining, since they'd have to revert to using the default impl or sprinkle a consume somewhere.

As far as disadvantages go, I guess it might also make the library less explicit by adding the implicit consume? So in that sense there may be a tradeoff between ergonomics and being explicit.

### DIFF
--- a/src/Control/Functor/Linear/Internal/Class.hs
+++ b/src/Control/Functor/Linear/Internal/Class.hs
@@ -119,8 +119,8 @@ class Applicative m => Monad m where
   -- | @x >>= g@ applies a /linear/ function @g@ linearly (i.e., using it
   -- exactly once) on the value of type @a@ inside the value of type @m a@
   (>>=) :: m a %1-> (a %1-> m b) %1-> m b
-  (>>) :: m () %1-> m a %1-> m a
-  m >> k = m >>= (\() -> k)
+  (>>) :: (Consumable a) => m a %1-> m b %1-> m b
+  m >> k = m >>= ((\() -> k) . consume)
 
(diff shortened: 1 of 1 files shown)


utdemir:
> As far as disadvantages go, I guess it might also make the library less explicit by adding the implicit consume? So in that sense there may be a tradeoff between ergonomics and being explicit.

You're right. To document it here, I think an alternative is to add:

```
void :: (Control.Functor f, Consumable a) => f a -> f () 
```

So, one would call `void $` explicitly when they need to consume the value.

However, I have no preference either way, I can see how different use cases can choose convenience over explicitness or vice versa. As we have at least your case choosing for convenience, I see no reason going the other way :).

aspiwack:
Thanks for your interest @saolof !

I prefer having a `void` function and be explicit about it (in fact, it is a bug that we don't have one :slightly_smiling_face: ). I wrote `(>>)` with the unit type for this reason (I've long had a dislike for the type in `base`). I think it invites fewer mistakes.

So I'll insist pretty strongly on not changing the type of `(>>)`.

tbagrel1:
I think that we could close this as soon as https://github.com/tweag/linear-base/pull/387 is merged.

PR #357, opened 2022-01-12
Add newtype 'AsMovable' to derive Consumable and Dupable from Movable

tbagrel1:


src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -22,59 +23,55 @@
...
+
+instance Movable a => Consumable (AsMovable a) where
+  consume (AsMovable x) =
+    move x & \case
+      Ur x' -> ()
+
+instance Movable a => Dupable (AsMovable a) where
+  dupV (AsMovable x) =
+    move x & \case
+      Ur x' -> Data.pure (AsMovable x')
```
aspiwack, line 55: Just to outline an unimportant alternative: you could have given an instance `instance Movable a => Movable (AsMovable a)`, and used it to save a little space in the `Consumable` and `Dupable` instances.

It doesn't change the generated code, and the code is still very short as it stands; I just wanted to point out that the possibility exists.

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -22,59 +23,55 @@
...
 
-instance Consumable () where
-  consume () = ()
+deriving via (AsMovable ()) instance Consumable ()
 
-instance Dupable () where
-  dupV () = Data.pure ()
+deriving via (AsMovable ()) instance Dupable ()
```
aspiwack, line 59: Since I'm doing trivial comments: I'm voting removing the empty line between two `deriving instance` one-liners.

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -426,24 +313,36 @@ instance Prelude.Traversable Ur where
 
 -- Some stock instances
 deriving instance Consumable a => Consumable (Sum a)
+
```
aspiwack, line 316: Ah, but I see some such lines have been added. Is that an Ormolu thing?

aspiwack (approved):
I guess that I'm seeing formatting here because of the fact that this isn't based on #355 ? You will have silly conflict, probably.

Looking good, though :slightly_smiling_face: I like seeing -100 net lines.

src/Data/Unrestricted/Internal/Instances.hs
tbagrel1: Oh, I didn't know it was possible to do so, and it looks clearer to me with `instance Movable a => Movable (AsMovable a)`, so I'll add it

src/Data/Unrestricted/Internal/Instances.hs
tbagrel1: As you guessed it, it is an ormolu thing. Maybe I could rewrite it in the "packed" style (so as to make the diff cleaner), as PR https://github.com/tweag/linear-base/pull/355 will ormolize the whole codebase later?

aspiwack: It would have been preferable. But it's probably not worth the time investment now.

aspiwack (approved):

PR #358, opened 2022-01-14
Format the codebase with ormolu/cabal format

tbagrel1:


aspiwack (approved):

andreabedini:
I think this PR accidentally removed the [comments](https://github.com/tweag/linear-base/blob/9b27ec8d3b32e0d2ee598ed3107a93bb1e83c6c9/linear-base.cabal#L183-L214) in `linear-base.cabal` about doctests. Removing comment is an unfortunate and well-know issue with cabal-format.

I am not suggesting the comments should be reinstated.

For reference this is the removed comment block

```
-- TODO: Uncomment below block and set 'build-type' to 'Custom' to enable
-- doctests once cabal-install 3.4 is released.
--
-- Longer story:
--
-- cabal-install has a piece of code[1] which injects a Cabal upper bound to
-- packages with custom Setup.hs's. And this happens after the overrides,
-- so the usual mechanisms of overriding upper bounds does not work.
--
-- GHC 9 comes with Cabal 3.4, which is above that bound. So, when using
-- GHC 9 with cabal-install 3.2; `build-type: Custom` causes another Cabal
-- library to be built, and that causes a strange type error ("expecting IO,
-- but got IO"), which I suspect because it conflicts with the existing boot
-- packages.
--
-- [1]: https://github.com/haskell/cabal/blob/d28c80acc69b9e7fa992a0b2b7fced937734b238/cabal-install/src/Distribution/Client/ProjectPlanning.hs#L1132-L1149

-- custom-setup
--  setup-depends:
--    base >= 4 && <5,
--    Cabal,
--    cabal-doctest
--
-- test-suite doctests
--   type:                 exitcode-stdio-1.0
--   hs-source-dirs:       test/
--   main-is:              Doctest.hs
--   build-depends:        base
--                       , doctest
--                       , linear-base
--   ghc-options:          -Wall -threaded
--   default-language:     Haskell2010
```

tbagrel1:
Nice catch! If anyone thinks that these comments should be reinstated, please let me know.

PR #360, opened 2022-01-18
Dupable wars : implementation

tbagrel1:


src/Data/Unrestricted/Linear.hs
```diff
@@ -70,6 +70,8 @@ module Data.Unrestricted.Linear
     Consumable (..),
     Dupable (..),
     Movable (..),
+    RepStream (..),
```
aspiwack, line 73: I don't think that we should export `RepStream`'s constructor. It's better left as an abstract type.

Also, since we export it, we probably want to give it a better name still (I do recognise that `RepStream` is significantly better than `Streamish` :smile: ).

src/Data/V/Linear/Internal/V.hs
```diff
@@ -25,6 +24,9 @@ module Data.V.Linear.Internal.V
 
     -- * Type-level utilities
     caseNat,
+    expandFunN,
+    Dict (..),
+    predNat,
```
aspiwack, line 29: This looks pretty unrelated to this PR. (edit: found it, see below)

src/Data/Unrestricted/Internal/Replicator.hs
```diff
@@ -0,0 +1,19 @@
...
+import Data.V.Linear.Internal.V (FunN)
+import Prelude (Either(..))
+import GHC.TypeLits
+
+data RepStream a where
+  RepStream :: (s %1 -> a) -> (s %1 -> (s, s)) -> (s %1 -> ()) -> s %1 -> RepStream a
+
+data Replicator a where
+  Moved :: a -> Replicator a
+  CollectFrom :: RepStream a %1 -> Replicator a
```
aspiwack: You don't have access to the `Applicative` class at this point†, but you should have a `pure` and a `(<*>)` function nonetheless. For both types.

We will probably need other functions, like a function `Replicator %1 -> RepStream`.

† Sometimes I feel as though this library should all be one big `Internal` file where everything is mutually recursive, and then we use the file hierarchy purely for exports.

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -351,3 +361,74 @@ instance (Movable a, Prelude.Semigroup a) => Semigroup (MovableMonoid a) where
...
+    CollectFrom stream -> CollectFrom $ Data.fmap f stream
+instance Data.Applicative Replicator where
+  pure = Moved
+  (Moved f) <*> (Moved x) = Moved (f x)
+  sf <*> sx = CollectFrom (toRepStream sf Data.<*> toRepStream sx)
+
+toRepStream :: Replicator a %1 -> RepStream a
+toRepStream = \case
+  Moved x -> Data.pure x
+  CollectFrom stream -> stream
```
aspiwack: Ah, here it is. But it ought to be defined near the type definition.

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -351,3 +361,74 @@ instance (Movable a, Prelude.Semigroup a) => Semigroup (MovableMonoid a) where
...
+
+nextR :: Replicator a %1 -> (# a, Replicator a #)
+nextR (Moved x) = (# x, Moved x #)
+nextR (CollectFrom (RepStream givex dupsx consumesx sx)) =
```
aspiwack: This function should also be defined near the type.

I'm not quite sure why you need it yet.

aspiwack:
```suggestion
nextR (CollectFrom (RepStream give dups consumes x)) =
```

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -351,3 +361,74 @@ instance (Movable a, Prelude.Semigroup a) => Semigroup (MovableMonoid a) where
...
+  dupsx sx & \case
+    (sx1, sx2) -> (# givex sx1, CollectFrom (RepStream givex dupsx consumesx sx2) #)
+
+elim' :: forall n a b. KnownNat n => Replicator a %1 -> FunN n a b %1 -> b
```
aspiwack, line 426: Ah, hence the export before. This points to a need or reordering things a little.

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -55,9 +64,9 @@ instance Movable a => Consumable (AsMovable a) where
       Ur _ -> ()
 
 instance Movable a => Dupable (AsMovable a) where
-  dupV x =
+  dupR x =
     move x & \case
-      Ur x' -> Data.pure x'
+      Ur x' -> Moved x'
```
aspiwack: `Data.pure` works as well, doesn't it? This seems spurious to me.

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -239,18 +248,19 @@ instance (Movable a, Movable b, Movable c) => Movable (a, b, c) where
 instance (KnownNat n, Consumable a) => Consumable (V n a) where
   consume (V xs) = consume (Unsafe.toLinear Vector.toList xs)
 
+-- TODO: check?
```
aspiwack, line 251: Check what so?

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -351,3 +361,74 @@ instance (Movable a, Prelude.Semigroup a) => Semigroup (MovableMonoid a) where
...
+
+instance Consumable (Replicator a) where
+  consume = \case
+    Moved _ -> ()
+    CollectFrom stream -> consume stream
+
+instance Dupable (Replicator a) where
+  dupR = \case
+    Moved x -> Moved (Moved x)
+    CollectFrom stream -> CollectFrom Data.<$> dupR stream
```
aspiwack: Similarly to previously, should be defined near the types, and the type-class instance should just be a wrapper around the primitive functions.

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -351,3 +361,74 @@ instance (Movable a, Prelude.Semigroup a) => Semigroup (MovableMonoid a) where
...
+
+instance Data.Functor RepStream where
+  fmap f (RepStream givef dupsf consumesf sf) =
+    RepStream (f . givef) dupsf consumesf sf
+instance Data.Applicative RepStream where
```
aspiwack, line 387: These instances should be in the module where `Functor` and `Applicative` are defined, I think.

aspiwack (commented):
What I would like to see, as well, is a function in `Data.Array.Mutable.Linear` which defines an array from a `Replicator`. To test the `Replicator` API.

src/Data/Unrestricted/Internal/Replicator.hs
aspiwack: Random thought: should we use unboxed pairs (and unit) in the projections? Maybe it'll be better?

aspiwack:
I pushed a little something to restore the backward functional
dependency.

As I see it, there are two solutions:

1. Remove the `n` argument from the `Elim` type class arguments
   altogether. It is effectively redundant, after all. This removes
   all functional dependencies.
2. We need to tell the compiler how to compute `n` from `f` and
   `b`. One way to convince oneself that GHC cannot know that `n` is
   effectively determined by `f` is to consider what would happen if
   we added an `instance Elim 42 a b b`? We need to make this
   impossible!
   
While (1) is definitely the simpler solution, I went for (2) in my
commit, because I do like the idea of having the option of saying
`elim @7` when the function type is under-determined. Rather than
having to state the entire function type.

What this implied is making a type family `Arity b f` which computes
the arity of `f` (assuming its headed by `b`). And force `n ~ Arity b
f` as a super-class constraint. This way, no instance 42. This
suffices to convince GHC.

What I don't know if why I had to add `Arity b (a %1 -> b) = 1`, this
should follow from the definition. But when I remove it, Haskell is
very unhappy.

tbagrel1:
Should I redefine `Arity` when writing `V.Elim`, or should we share it someway?

Also, I was wondering yesterday, could we add a `2 <= n` extra constraint on the last `Elim n a b f` instance instead of creating overlapping ones? What would be the cost of such a measure?

src/Data/Unrestricted/Internal/Replicator.hs
tbagrel1: If we use unboxed tuple for the `dups` projector, then it becomes harder to write `dupR` in terms of `dup2` (or we should also change the `dup2` signature to use unboxed tuples).
I don't know how to write an unboxed unit, but if we return some in the `consumes` projector, then writing `consume` (from `Consumable`) for `ReplicationStream` is also less trivial.
I also tried to use unboxed tuples in the `Applicative` instance for `ReplicationStream` (for `<*>`), but AFAIK, it's not possible because the projections cannot receive an unboxed tuple as a function parameter.

Also, without 9.2, I don't see how we can actually write this:
```
dupsf sf' & \case
  (# sf1, sf2 #)
```
We need the `func_call & lambda_case` syntax here for linearity, but because the result of the function call is now passed as an argument to `&`, then unboxed tuples are not allowed in that position.

treeowl: Unboxed unit is `(##)`.

aspiwack:
> Should I redefine `Arity` when writing `V.Elim`, or should we share it someway?

I guess it's better to use the same.

Either way, it gives some pretty unappealing types, so we will have to give pretty solid documentation to `elim`.

> Also, I was wondering yesterday, could we add a `2 <= n` extra constraint on the last `Elim n a b f` instance instead of creating overlapping ones? What would be the cost of such a measure?

Unfortunately it wouldn't work. GHC only looks at the instance's head (the part to the right of the `=>`) to determine if instances are overlapping. You could add this constraint, but it would never apply.

tbagrel1:
Do you see where we could but the `Arity` thing so? In `Data.Unrestricted.Linear.Internal.Arity`?

> Unfortunately it wouldn't work. GHC only looks at the instance's head (the part to the right of the =>) to determine if instances are overlapping. You could add this constraint, but it would never apply.

I would have used something like `Elim (n + 1) a b f) => Elim (n + 2) a b (a %1 -> f)` then, if only we could.

src/Data/Unrestricted/Internal/Replicator.hs
aspiwack: Well, let's wait 9.2 before we decide to revisit this issue, then. It's not worth agonising over.

aspiwack:
> Do you see where we could but the `Arity` thing so? In `Data.Unrestricted.Linear.Internal.Arity`?

I guess. There is no good place where to put it anyway. The relation with `Unrestricted` is kind of dubious, though. Maybe `Data.Linear.Internal.Arity` instead?

facundominguez (commented):
I came to have some fun explaining the code. I shared some suggestions for documentation that I think could be helpful.

src/Data/Unrestricted/Internal/Replicator.hs
```diff
@@ -0,0 +1,92 @@
...
+  elim f r = f (extract r)
+  {-# INLINE elim #-}
+
+instance {-# OVERLAPPABLE #-} (n ~ Arity b (a %1 -> f), Elim (n - 1) a b f) => Elim n a b (a %1 -> f) where
+  elim g r =
+    next r & \case
+      (a, r') -> elim (g a) r'
+  {-# INLINE elim #-}
+
+type family Arity b f where
```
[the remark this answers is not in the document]
aspiwack: Oh, I didn't mean to leave this here indeed. I meant to remove the line altogether or make a type error (probably the latter), it was just a quickly put together placeholder. Thanks for spotting this.

src/Data/V/Linear.hs
```diff
@@ -36,17 +37,20 @@
 -- rather than linear types).
 module Data.V.Linear
   ( V,
+    next#,
+    next,
```
aspiwack, line 41: These are not good names for vector operations. I guess they could be called `uncons` or something. Though I'm not sure whether there is impetus to export them at all.

src/Data/V/Linear/Internal/Ambiguous.hs
```diff
@@ -0,0 +1,90 @@
...
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# OPTIONS_HADDOCK hide #-}
+
+module Data.V.Linear.Internal.Ambiguous (theLength, make, FunN) where
+
+import Data.Kind
+import Data.Proxy
+import Data.Type.Equality
+import {-# SOURCE #-} Data.V.Linear.Internal.V (V (..), cons)
```
aspiwack, line 20: Let's not. I really don't want recursive imports unless it's absolutely necessary. I'd rather you add extra files instead.

Though, I guess this file is temporary? Until we clean up `make`? If so, I guess it's ok.

src/Data/V/Linear/Internal/Ambiguous.hs
```diff
@@ -0,0 +1,90 @@
...
+    error,
+    fromIntegral,
+    (-),
+  )
+
+theLength :: forall n. KnownNat n => Int
+theLength = fromIntegral (natVal' @n (proxy# @_))
+
+-- TODO: improve
+-------------------------------------------------------------------------------
```
aspiwack, line 40: Improve what?

src/Data/V/Linear/Internal/V.hs
```diff
@@ -61,115 +62,74 @@ structure.
...
-theLength :: forall n. KnownNat n => Int
-theLength = fromIntegral (natVal' @n (proxy# @_))
+consume :: V 0 a %1 -> ()
+consume = Unsafe.toLinear (\_ -> ())
+
+-- Backward compatility
+-- consumeV :: V 0 a %1 -> b %1 -> b
+-- consumeV v b = consume v & \case
+--   () -> b
+
```
aspiwack, line 83: Remove the dead code.
```suggestion
```

src/Data/V/Linear.hs
```diff
@@ -36,17 +37,20 @@
...
+    -- split,
```
aspiwack, line 42: Remove the dead code
```suggestion
```

src/Data/V/Linear.hs
```diff
@@ -36,17 +37,20 @@
...
+    Elim (..),
+    cons,
+    -- fromReplicator,
+    -- dupV,
```
aspiwack, line 46: You should export these.

src/Data/V/Linear.hs
```diff
@@ -36,17 +37,20 @@
 -- rather than linear types).
 module Data.V.Linear
   ( V,
+    next#,
+    next,
+    -- split,
+    Elim (..),
```
aspiwack, line 43:
```suggestion
    Elim ,
    elim ,
```

src/Data/V/Linear/Internal/V.hs
```diff
@@ -61,115 +62,74 @@ structure.
...
-      prepend :: a %1 -> FunN (n - 1) a (V n a)
-      prepend t = case predNat @n of
-        Dict -> continue @(n - 1) @a @(V (n - 1) a) (cons t) (make @(n - 1) @a)
+    next' :: 1 <= n => V n a -> (a, V (n - 1) a)
+    next' (V xs) = (Vector.head xs, V (Vector.tail xs))
+
+-- Backward compatility
+-- split :: 1 <= n => V n a %1 -> (# a, V (n - 1) a #)
+-- split = next#
+
```
aspiwack, line 110:
```suggestion
```

src/Data/V/Linear/Internal/V.hs
```diff
@@ -61,115 +62,74 @@ structure.
...
-          Prelude.Right Refl ->
-            dup a & \(a', a'') ->
-              a' `cons` go @(m - 1) a''
-
-    -- An unsafe cast to prove the simple equality.
-    pr1 :: forall k. 0 :~: (k - 1) -> k :~: 1
-    pr1 Refl = Unsafe.coerce Refl
+-- TODO: fails because Elim and make are not dual with the current implementation
+-- fromReplicator :: forall n a. KnownNat n => Replicator a %1 -> V n a
+-- fromReplicator = Replicator.elim (make @n)
```
aspiwack, line 132: I see your problem, this should be solvable though. You just need to ask GHC to verify an additional fact (which happens to always be true).

```suggestion
fromReplicator :: forall n a. (KnownNat n, Replicator.Elim n a (V n a) (FunN n a (V n a))  => Replicator a %1 -> V n a
fromReplicator = Replicator.elim (make @n)
```

src/Data/V/Linear/Internal/V.hs
```diff
@@ -61,115 +62,74 @@ structure.
...
+
+-- dupV :: forall n a. (KnownNat n, Dupable a) => a %1 -> V n a
+-- dupV = fromReplicator . dupR
```
aspiwack, line 135: Then this should work as well
```suggestion
dupV :: forall n a. (KnownNat n, Dupable a, Replicator.Elim n a (V n a) (FunN n a (V n a)) => a %1 -> V n a
dupV = fromReplicator . dupR
```

src/Data/V/Linear/Internal/V.hs-boot
```diff
@@ -0,0 +1,53 @@
...
+    pure,
+    (<*>),
+    next#,
+    next,
+    -- split,
+    Elim (..),
+    cons,
+    -- fromReplicator,
+    -- dupV,
+  )
```
aspiwack, line 27: You don't have to write everything in the hs-boot file. Just whatever you will need in your loop. (I'm still unhappy about this, I very much expect that it doesn't survive until the end of the week)

src/Streaming/Internal/Process.hs
```diff
@@ -94,7 +94,7 @@ import Data.Functor.Sum
 import qualified Data.IntSet as IntSet
 import qualified Data.Sequence as Seq
 import qualified Data.Set as Set
-import Data.Unrestricted.Linear
+import Data.Unrestricted.Linear (Consumable, Ur (..), lseq)
```
aspiwack, line 97: This is a little suspicious. What happened?

aspiwack:
I should add: this is another PR that got bigger than we planned. We should be more careful for the next ones.

src/Data/Arity/Linear/Internal/Arity.hs
```diff
@@ -0,0 +1,14 @@
...
+{-# LANGUAGE UndecidableInstances #-}
+
+module Data.Arity.Linear.Internal.Arity (Arity) where
+
+import GHC.TypeLits
+
+type family Arity b f where
+  Arity b b = 0
+  Arity b (a %1 -> f) = Arity b f + 1
+  Arity _ _ = TypeError ('Text "The given object of type f is not a well-formed function")
```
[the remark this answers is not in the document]
tbagrel1: Agreed.

src/Data/Replicator/Linear.hs
```diff
@@ -0,0 +1 @@
+module Data.Replicator.Linear () where
```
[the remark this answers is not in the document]
tbagrel1: Well, when I started to write this file, I didn't know what to export yet, so I left it empty. It definitely needs to be changed.

src/Data/Replicator/Linear/Internal/ReplicationStream.hs
```diff
@@ -0,0 +1,46 @@
...
+
+data ReplicationStream a where
+  ReplicationStream :: (s %1 -> a) -> (s %1 -> (s, s)) -> (s %1 -> ()) -> s %1 -> ReplicationStream a
+
+consume :: ReplicationStream a %1 -> ()
+consume (ReplicationStream _ _ consumes s) = consumes s
+
+fmap :: (a %1 -> b) -> ReplicationStream a %1 -> ReplicationStream b
+fmap f (ReplicationStream give dups consumes s) =
+  ReplicationStream (f . give) dups consumes s
```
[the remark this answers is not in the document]
tbagrel1: It would be too simple if the "traditional name convention" didn't add more confusion :joy:. I will change that

src/Data/Replicator/Linear/Internal/ReplicationStream.hs
```diff
@@ -0,0 +1,46 @@
...
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# OPTIONS_HADDOCK hide #-}
+
+module Data.Replicator.Linear.Internal.ReplicationStream (ReplicationStream (..), consume, fmap, pure, (<*>)) where
+
+import Data.Unrestricted.Internal.Ur
+import Prelude.Linear.Internal
+
+data ReplicationStream a where
+  ReplicationStream :: (s %1 -> a) -> (s %1 -> (s, s)) -> (s %1 -> ()) -> s %1 -> ReplicationStream a
```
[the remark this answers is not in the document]
tbagrel1: Agreed.

src/Data/Replicator/Linear/Internal/Replicator.hs
```diff
@@ -0,0 +1,86 @@
...
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneKindSignatures #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UnboxedTuples #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# OPTIONS_HADDOCK hide #-}
+
+module Data.Replicator.Linear.Internal.Replicator (Replicator (..), consume, fmap, pure, (<*>), next, next#, extract, Elim (..)) where
```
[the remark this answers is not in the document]
tbagrel1: Ok for `elim`.

About the module name, should we also change `Data.V.Linear.Internal.V` to `Data.V.Linear.Internal` ?

src/Data/V/Linear.hs
tbagrel1: `uncons` looks good to me. They were part of the original API (named `split`). Do we have any major reason to preserve the old API?

src/Data/V/Linear/Internal/Ambiguous.hs
tbagrel1: Yeah, the goal is just to have most of the `V` functions in a file without the `AllowAmbiguousTypes` extension.

src/Data/V/Linear/Internal/Ambiguous.hs
tbagrel1: Improve the `make` implementation, and remove all the Nat-related functions

src/Data/V/Linear/Internal/V.hs
tbagrel1: So, we don't need to preserve the old API, right?

src/Data/V/Linear.hs
tbagrel1: As you found below, I can't export them because these functions do not typecheck yet (because of the `elim/make` mess).

src/Data/Replicator/Linear/Internal/Replicator.hs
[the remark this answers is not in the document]
aspiwack: Probably, yes.

src/Data/V/Linear.hs
aspiwack: Not really. Not this at the very least. I probably exported this function myself… so it's my bad.

src/Data/V/Linear/Internal/Ambiguous.hs
aspiwack: Mmm… but `make` will always have ambiguous type variables, most likely. So we will have this problem. I'd really like to avoid recursive imports.

src/Data/V/Linear/Internal/V.hs
aspiwack: Right.

src/Streaming/Internal/Process.hs
tbagrel1: When I started this PR, `Replicator`, and its `next/next#` methods were defined in `Data.Unrestricted.Linear.Internal.Replicator`, so they were publicly reexported by `Data.Unrestricted.Linear`.
`src/Streaming/Internal/Process.hs` also defines a `next` function, creating a name clash.

aspiwack: So you can revert this, now? Please do.

src/Data/Replicator/Linear/Internal/Replicator.hs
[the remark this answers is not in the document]
tbagrel1: The internals of the `Replicator` and `ReplicationStream` types are used in `Dupable` definition, and in `Dupable` instances.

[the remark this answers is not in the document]
aspiwack: Ah, yes, more silly circular dependencies (though I don't know why they would be used in the definition of `Dupable`). Fair enough. But then you want to restrict the export in the non-internal API.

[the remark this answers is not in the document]
tbagrel1: In instances:
```haskell
instance Dupable (ReplicationStream a) where
  dupR (ReplicationStream give dups consumes s) =
    Streamed $
      ReplicationStream
        (ReplicationStream give dups consumes)
        dups
        consumes
        s

instance Dupable (Replicator a) where
  dupR = \case
    Moved x -> Moved (Moved x)
    Streamed stream -> Replicator.fmap Streamed $ dupR stream
```

In Dupable:
```haskell
class Consumable a => Dupable a where
  {-# MINIMAL dupR | dup2 #-}

  dupR :: a %1 -> Replicator a
  dupR x = Streamed $ ReplicationStream id dup2 consume x
```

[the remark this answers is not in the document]
tbagrel1: But I agree, we should restrict them in the non-internal API. I think that we said only `Replicator` (type) would be exported, and `Replicator` constructors, `ReplicationStream` type, and its constructor should stay internal.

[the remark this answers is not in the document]
aspiwack: Now that I think of it, these functions (at least those on `Replicator`) should be defined in the API anyway. As `Replicator.duplicate` and `Replicator.unfold`.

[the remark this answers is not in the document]
tbagrel1: We need to export `Elim (..)` at internal level so that we can make `fromReplicator` typecheck.

src/Data/Unrestricted/Internal/Instances.hs
```diff
@@ -42,6 +45,8 @@ import Prelude.Linear.Internal
 import qualified Unsafe.Linear as Unsafe
 import qualified Prelude
 
+-- | Newtype that must be used with @DerivingVia@ to get efficient @Dupable@
+-- and @Consumable@ implementation for @Movable@ types.
```
aspiwack, line 49:
```suggestion
-- | Newtype that must be used with @DerivingVia@ to get efficient 'Dupable'
-- and 'Consumable' implementation for 'Movable' types.
```

src/Data/Replicator/Linear.hs
```diff
@@ -0,0 +1,15 @@
+{-# LANGUAGE MagicHash #-}
+{-# OPTIONS_GHC -Wno-dodgy-exports #-}
+
+module Data.Replicator.Linear
+  ( Replicator,
+    next,
+    next#,
+    extract,
+    elim,
+    module Data.Replicator.Linear.Internal.Instances,
```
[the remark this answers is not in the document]
aspiwack: Also export the other function `map`, `pure`, etc… they don't hurt. And that lets the programmer call them explicitly rather than through the

aspiwack:
> Why are we deriving Prelude (non-linear) versions of `Eq`, `Ord`, and especially `Functor` for `V`?

I don't remember for sure. But if I had to guess, it's because we wanted `V n a` to behave as much as possible as `(a, …, a)`. We'd need a multiplicity polymorphic `elim` for this to work completely, though. We may have to wait until we are fully committed to 9.2 for this to work fine though. It may be worth a try (in a separate PR).

```haskell
elim :: (a %p -> … %p -> b) %1 -> V n a %p -> b
```

src/Data/Replicator/Linear.hs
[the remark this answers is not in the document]
aspiwack: I don't. Maybe it has non-instances in it, despite the name?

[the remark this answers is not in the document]
tbagrel1: Nice catch. Indeed, there are non-instance things exported. Maybe we should write an explicit export list for this module?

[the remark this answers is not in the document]
aspiwack: As you feel like. Probably not in this PR, though.

You may have to write explicit export list for any non-internal module, anyway, just so that you can add Haddock sections in them.

src/Data/Replicator/Linear/Internal.hs
```diff
@@ -0,0 +1,101 @@
...
+import Data.Replicator.Linear.Internal.ReplicationStream (ReplicationStream (..))
+import qualified Data.Replicator.Linear.Internal.ReplicationStream as ReplicationStream
+import GHC.TypeLits
+import Prelude.Linear.Internal
+
+-- | @Replicator@ is used to linearly duplicate values in an efficient fashion.
+-- @Movable@ types are deep-copied only once (no matter how many copies are
+-- required), whereas non-@Movable@ values are duplicated using the internal
+-- construct @ReplicationStream@. @Replicator@ also have a linear
+-- @Data.Applicative@ instance.
```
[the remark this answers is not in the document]
tbagrel1: See https://stackoverflow.com/a/23858109
(`consume`, `dup2`) seems to be the definition of a comonoid.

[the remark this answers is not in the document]
tbagrel1: But I'm not sure that `Replicator` is a co-monoid though. But any `Dupable` type seems to be an instance of comonoid.

src/Data/Replicator/Linear.hs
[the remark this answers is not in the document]
tbagrel1: > Also export the other function map, pure, etc… they don't hurt. And that lets the programmer call them explicitly rather than through the

I don't really see the point. That could create conflicts if the programmer import the module unqualified, couldn't it?
I'm going to do it anyway (we could always revert the change). Should we also export `consume` (I don't think so, but again, I don't have enough hindsight)?

src/Data/Replicator/Linear/Internal.hs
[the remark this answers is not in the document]
aspiwack: To be completely exact, comonoids can be presented with `consume` and `dup2`. But it's not the only presentation. The `Dupable` class, is quite concretely the class of comonoids for the linear functions (and the usual product), it's a different presentation (if more redundant).

In particular `Replicator a` is a comonoid, since it implements `Dupable` (which means the same thing).

The presentation of comonoids in terms of `dupR` corresponds to a presentation of a coalgebraic structures as co-algebras of comonads (here, the comonad is `Replicator`). That being said, with a bit more thought into it, we've basically just defined `Replicator a` as `∃ x. Dupable x ∧ x %1 -> a`, which pretty much reads as “I'm the free `Dupable`”. So I guess that it's fairly unsurprising after all. And not very informative.

src/Data/Replicator/Linear/Internal/ReplicationStream.hs
```diff
@@ -0,0 +1,68 @@
...
+-- | @'ReplicatorStream' s g dup2 c@ is the infinite linear stream
+-- @repeat (g s)@ where @dup2@ is used to make as many copies of @s@ as
+-- necessary, and @c@ is used to consume @s@ when consuming the stream.
+-- 'ReplicatorStream' is used to implement 'Replicator', for the
+-- non-'Data.Unrestricted.Movable' types, and is not exposed to the end user.
+--
+-- Although it isn't enforced at type level, @dup2@ should abide by the same
+-- laws as 'Data.Unrestricted.Dupable.dup2':
+-- * @first c (dup2 a) ≃ a ≃ second c (dup2 a)@ (neutrality)
+-- * @first dup2 (dup2 a) ≃ (second dup2 (dup2 a))@ (associativity)
```
aspiwack, line 28:
```suggestion
-- * @first dup2 (dup2 a) ≃ (second dup2 (dup2 a))@ (associativity)
--
-- This type is solely used for the implementation of 'Data.Replicator.Linear'.
```

src/Data/Unrestricted/Internal/Dupable.hs
```diff
@@ -1,47 +1,72 @@
...
 -- * @first dup2 (dup2 a) ≃ (second dup2 (dup2 a))@ (associativity)
 --
 -- Where the @(≃)@ sign represents equality up to type isomorphism.
 --
--- When implementing 'Dupable' instances for composite types, using 'dupV'
--- should be more convenient since 'V' has a zipping 'Applicative' instance.
+-- Implementation of 'Dupable' for 'Data.Unrestricted.Movable' types should
+-- be done with @deriving via 'Data.Unrestricted.AsMovable' ...@, whereas
+-- implementation for non-'Data.Unrestricted.Movable' types should be done
+-- using either 'dupR' (for data types composed of 'Dupable' members) or 'dup2'.
```
aspiwack, line 36:
```suggestion
-- Implementation of 'Dupable' for 'Data.Unrestricted.Movable' types should
-- be done with @deriving via 'Data.Unrestricted.AsMovable' ...@'.
```

src/Data/Unrestricted/Internal/Dupable.hs
```diff
@@ -1,47 +1,72 @@
...
+import qualified Data.Replicator.Linear.Internal as Replicator
+import Data.Replicator.Linear.Internal.ReplicationStream (ReplicationStream (..))
 import Data.Unrestricted.Internal.Consumable
-import Data.V.Linear.Internal.V (V)
-import qualified Data.V.Linear.Internal.V as V
-import GHC.TypeLits
+import Prelude.Linear.Internal
 
--- | The laws of @Dupable@ are dual to those of 'Monoid':
+-- | The laws of 'Dupable' are dual to those of 'Monoid':
```
aspiwack, line 26: We are missing laws for `dupR`.

Let's see. We need

```haskell
dup2 = Replicator.elim (,) . dupR
```

Correspondingly for `consume`

```haskell
consume = Replicator.consume . dupR
```

We could state this one another way, like below. Maybe it's better, actually.

```haskell
consume = Replicator.elim () . dupR
```

And we need laws for the comonadic structure

```haskell
* Replicator.extract . dupR = id
* Replicator.duplicate . dupR = (Replicator.map dupR) . dupR
```

These laws make the `consume`/`dup2` laws redundant. But we should leave them in as they are probably easier to understand.
```

File: src/Data/Unrestricted/Internal/Dupable.hs [hunk 35]
```diff
@@ -1,47 +1,72 @@
...
+  {-# MINIMAL dupR | dup2 #-}
 
-  dupV :: forall n. KnownNat n => a %1 -> V n a
-  dupV a =
-    case V.caseNat @n of
-      Prelude.Left Refl -> a `lseq` V.make @0 @a
-      Prelude.Right Refl -> V.iterate dup2 a
+  -- | Creates a 'Replicator' for the given @a@. This 'Replicator' type has
+  -- a 'Data.Functor.Linear.Applicative' instance, which allows for an easy
+  -- composition.
```
aspiwack, line 42:
```suggestion
  -- | Creates a 'Replicator' for the given @a@.
  --
  -- You usually want to define this method using 'Replicator'\'s 'Data.Functor.Linear.Applicative'
  -- instance. For instance, here is an implementation of `Dupable [a]`:
  -- 
  -- > instance Dupable a => Dupable [a] where
  -- >   dupR [] = pure []
  -- >   dupR (a : as) = (:) <$> dupR a <*> dupR as
```

aspiwack (approved):
I've got some comments on the documentation. You can address them then merge.

src/Data/Replicator/Linear.hs
```diff
@@ -0,0 +1,30 @@
...
+-- | This module defines a stream-like type named 'Replicator', which is
+-- mainly used in the definition of the 'Dupable' class to provide efficient
+-- linear duplication.
+-- The API of 'Replicator' is close to the one of an infinite stream: it
+-- can either produce a new value linearly (with 'next' or 'next#'), or be
+-- linearly discarded (with 'consume' or 'extract').
+-- 'Data.Unrestricted.Movable' types are deep-copied only once (no matter
+-- how many copies are required), whereas non-'Data.Unrestricted.Movable'
+-- values are duplicated on demand using an internal stream type.
+-- 'Replicator' also have a linear 'Data.Functor.Linear.Applicative' instance.
```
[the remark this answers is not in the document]
tbagrel1: I like the use of the world "replica", and the whole paragraph looks good to me.

src/Data/Replicator/Linear.hs
```diff
@@ -0,0 +1,30 @@
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# OPTIONS_GHC -Wno-dodgy-exports #-}
```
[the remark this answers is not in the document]
tbagrel1: > -fwarn-dodgy-exports:
Causes a warning to be emitted when a datatype T is exported with all constructors, i.e. T(..), but is it just a type synonym.
Also causes a warning to be emitted when a module is re-exported, but that module exports nothing.

This is something I copied from `V/Linear.hs` when I had a warning about exporting the `Instances` module. But you taught me that it's unecessary to explicitely export a module containing only instances, so it is useless. I will remove it (and also remove it from `V/Linear.hs`).

src/Data/Replicator/Linear.hs
```diff
@@ -0,0 +1,30 @@
...
+  ( Replicator,
+    consume,
+    map,
+    pure,
+    (<*>),
+    next,
+    next#,
+    take,
+    extract,
+    elim,
```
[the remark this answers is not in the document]
tbagrel1: Why should we export `Elim`?

src/Data/Replicator/Linear/Internal.hs
```diff
@@ -0,0 +1,136 @@
...
+  consume r & \case
+    () -> []
+take 1 r = [extract r]
+take n r =
+  next r & \case
+    (a, r') -> a : take (n - 1) r'
+
+-- | Returns the next item from @'Replicator' a@ and efficiently consumes
+-- the replicator at the same time.
+extract :: Replicator a %1 -> a
```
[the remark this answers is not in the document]
tbagrel1: If I'm not mistaken, `(Replicator a %1 -> b) %1 -> Replicator a %1 -> Replicator b` is the signature of `(<*>)`

src/Data/Unrestricted/Internal/Dupable.hs
tbagrel1: > And we need laws for the comonadic structure
    * Replicator.extract . dupR = id
    * Replicator.duplicate . dupR = (Replicator.map dupR) . dupR

I don't think that we need to add these laws. `ReplicationStream` and `Replicator` constructors are not exposed, so the user cannot implement something where they would fail. At the moment, `dupR` can only be defined by the user via `dup2` (where `give == id`), or via `AsMovable` (where `give == unur`). Also, `Replicator.{extract,duplicate,map}` are black boxes for the user, so I don't see how he could check that the laws hold. Finally, `Replicator.duplicate` is `dupR`.

Maybe we could write these laws in the internal documentation of `Replicator` (or `ReplicationStream`) though.

src/Data/Replicator/Linear.hs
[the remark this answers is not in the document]
aspiwack: If we don't, the type of `elim` becomes unutterable. Then users can't make abstractions or aliases around it. It's very frustrating when it happens to you, believe me.

src/Data/Replicator/Linear/Internal.hs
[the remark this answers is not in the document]
aspiwack: The parentheses are not what you think

```haskell
((Replicator a) %1 -> b) %1 -> Replicator a %1 -> Replicator b
```

src/Data/Unrestricted/Internal/Dupable.hs
aspiwack: Here is a way to write a faulty `Dupable` instance:

```haskell
instance Dupable Bool where
  dupR b = b `lseq` pure True
```

tbagrel1: > Here is a way to write a faulty `Dupable` instance:
> 
> ```haskell
> instance Dupable Bool where
>   dupR b = b `lseq` pure True
> ```

But in this case, either `dup2` is deduced from `dupR` and then, the `dup2` neutrality rule is violated, or `dup2` is implemented in a law-abiding way, and then, the `dup2`-`dupR` coherence rule is violated. My point is, I don't think that the user can implement something that would respect the first four rules, but would violate one of the comonadic-related laws.

aspiwack: Ah, I see what you mean.

I can't think of a way to define `dupR` such that `dup2` is correct, and associative and such, yet these laws are violated. It might exist in principle though (whereas if the `dupR` laws are correct, then `dup2`, being uniquely determined by `dupR`, will clearly have the laws validated (well, I say “clearly” I don't really anticipate that the proof will be easy actually)).

That being said, it doesn't hurt writing these laws as a guideline for someone who'd only write a `dupR` instance, rather than having them work backward from `dup2`. The laws are definitely redundant, and we can say it if you feel more comfortable. But I still think that we should have them.

src/Data/Replicator/Linear/Internal.hs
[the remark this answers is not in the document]
tbagrel1: So, this function implementation might be `thing f = Replicator.map f . Replicator.duplicate` ?
What would be the use-case of such a function ?

EDIT: after check, this function is called `extend` in `Control.Comonad`, and is effectively `extend f = fmap f . duplicate`. I will add it.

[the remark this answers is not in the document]
aspiwack: Right. This thing :slightly_smiling_face: .

src/Data/Replicator/Linear/Internal.hs
```diff
@@ -0,0 +1,148 @@
...
+extract (Moved x) = x
+extract (Streamed (ReplicationStream s give _ _)) = give s
+
+-- | 'extend' function to comply with the linear Comonad interface
+-- of 'Replicator'.
+extend :: ((Replicator a) %1 -> b) -> Replicator a %1 -> Replicator b
```
aspiwack, line 108:
```suggestion
-- | Comonadic 'extend' function.
--
-- > extend f = map f . duplicate
extend :: (Replicator a %1 -> b) -> Replicator a %1 -> Replicator b
```

PR #361, opened 2022-01-19
Enable doctesting through cabal-docspec

andreabedini:
This PR enables doctests using cabal-docspec

cabal-docspec has some advantages:

1. It doesn't need to be mentioned in linear-base.cabal
2. The same binary works with multiple versions of GHC

This PR:

1. Adds cabal-docspec to the nix-shell. *Note* I have done this in a
   terrible way, downloading the precompiled binary from github. I would
   appreciate any help.
2. Adds a step to the CI workflow that runs the tests.
3. There's also a small tweag to the doctests, since (AFAU) cabal-docspec
   is a bit more picky about what it puts in scope.

This PR superseeds #317.


nix/cabal-docspec.nix
```diff
@@ -0,0 +1,18 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+pkgs.stdenv.mkDerivation {
+  name = "cabal-docspec";
+
+  src = pkgs.fetchurl {
```
aspiwack, line 6: Presumably you could use `fetchzip` here, instead. This would simplify your fetch a bit.

nix/cabal-docspec.nix
```diff
@@ -0,0 +1,18 @@
...
+    url = "https://github.com/phadej/cabal-extras/releases/download/cabal-docspec-0.0.0.20211114/cabal-docspec-0.0.0.20211114.xz";
```
aspiwack, line 7: Honestly, considering the constraints, this may be the best solution available to us, since it seems to be working well enough.

aspiwack (approved):

nix/cabal-docspec.nix
andreabedini: apprently it doesn't like the xz compression:
```
[~] nix-prefetch-url --unpack https://github.com/phadej/cabal-extras/releases/download/cabal-docspec-0.0.0.20211114/cabal-docspec-0.0.0.20211114.xz
error: failed to open archive: Unrecognized archive format
```

(I am assuming `nix-prefetch-url --unpack` is the same as `fetchzip`)

nix/cabal-docspec.nix
andreabedini: I would think so. In the end, we are not redistributing nor depending on `cabal-docspec`. It's an utility anyone can use to run the doctests. Perhaps I could include some instructions about getting and using `cabal-docspec`, from a developer point of view.

PR #363, opened 2022-01-28
Hide internal module from haddock

tbagrel1:


tbagrel1:
I checked all modules, especially the internal ones that are not hidden on the Hackage released version, and all of them have the `{-# OPTIONS_HADDOCK hide #-}` flag, except `UrT.hs`

aspiwack (approved):

PR #364, opened 2022-02-04
Improve `make`/`Make` implementation for `Data.V.Linear`

tbagrel1:
`V n a` represents vector having `n` elements of type `a`. Such a vector can be built using the `make` function, taking `n` linear arguments of type `a`, and returning a `V n a`.

At the moment, `make` uses a lot of type-level machinery around `Nat`s, and is implemented in a way such that it cannot be inlined; even if `n` is compile-time known.

With https://github.com/tweag/linear-base/pull/360 (now merged), the `Data.V.Linear` module has been overhauled, and the `elim` function has been rewritten to be inlinable.

The aim of this PR is to provide an efficient `make` implementation using typeclasses so as to make it inlinable when `n` is known at compile-time. It also adds more symmetry between `make` and `elim`, and allows us to get rid of a lot of internal helper functions.

### DIFF
--- a/src/Data/V/Linear/Internal.hs
+++ b/src/Data/V/Linear/Internal.hs
@@ -34,17 +34,15 @@ module Data.V.Linear.Internal
     fromReplicator,
     dupV,
     theLength,
+    Make,
     make,
-    FunN,
   )
 where
 
 import Data.Arity.Linear.Internal.Arity
 import Data.Kind
-import Data.Proxy
 import Data.Replicator.Linear.Internal (Replicator)
 import qualified Data.Replicator.Linear.Internal as Replicator
-import Data.Type.Equality
 import Data.Unrestricted.Internal.Dupable (Dupable (dupR))
 import Data.Vector (Vector)
 import qualified Data.Vector as Vector
@@ -52,7 +50,6 @@ import GHC.Exts (proxy#)
 import GHC.TypeLits
 import Prelude.Linear.Internal
 import qualified Unsafe.Linear as Unsafe
-import Prelude (Bool (..), Either (..), Maybe (..), error, (-))
 import qualified Prelude
 
 -- | @'V' n a@ represents an immutable sequence of @n@ elements of type @a@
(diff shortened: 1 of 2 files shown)


tbagrel1:
I set the base of this PR to [tbagrel1/dupable-new-impl](https://github.com/tweag/linear-base/tree/tbagrel1/dupable-new-impl), so as to reduce clutter in the diff pane, but the idea is to merge this one only after that [tbagrel1/dupable-new-impl](https://github.com/tweag/linear-base/tree/tbagrel1/dupable-new-impl) has been merged into `master`.

aspiwack:
I know what this PR is about. Today. But please don't make a PR without a description. I'm not the only one who is to read these. And even if this was just for you and me, we may want to revisit the PR in the future to remember why we did X and Y, without an apt description, it becomes impossible.

src/Data/V/Linear/Internal.hs
```diff
@@ -125,13 +121,44 @@ instance (1 <= n, n ~ Arity b (a %1 -> f), Elim (n - 1) a b f) => Elim n a b (a
...
+-- in the accumulator
+-- (@V[ ... ] \<\>@ represents the "extend" operation for 'V'):
+--
+-- >     make @n
+-- > --> make' @n @n (V[] <>)
+-- > --> λx. make' @(n - 1) @n (V[x] <>)
+-- > --> λx. λy. make' @(n - 2) @n (V[x, y] <>)
+-- > --> ...
+-- > --> λx. λy. ... λz. make' @0 @n (V[x, y, ... z] <>)    -- make' @0 @n f = f V[]
+-- > --> λx. λy. ... λz. V[x, y, ... z]
```
aspiwack, line 150: I think that this documentation should be on the `make` function, not the `Make` class.

Plus it speaks of `make'`, which isn't exported. Let's avoid this. If you need to document the implementation, do so outside of the docstring, in regular comments.

Remember to give instantiations of the type of `make` with concrete values. Examples are very helpful to disentangle very general types.

aspiwack (approved):
Fix the documentation then merge.

We will want to have some tests of the staging of both `elim`-s as well as `make` in the test suite before we make a release.

PR #366, opened 2022-02-04
Continue `genericTraverse` work

tbagrel1:
The idea is to bring the open PR https://github.com/tweag/linear-base/pull/349 on `genericTraverse` to a conclusion. I cannot edit it directly because it is a branch originating from a @treeowl's fork.

> genericTraverse is an efficient implementation of traverse for
any Generic1 type.


treeowl:
FYI, I'm emerging from the void, and hoping to get back into some of this stuff this week. But I don't oppose your assistance.

ConnorBaker:
I mis-clicked and marked this as ready to merge by accident -- trying to figure out how to undo that now.

treeowl:
That's because GitHub's UI for that is awful. The "mark ready" is at the bottom, but the "convert to draft" is on the top of the right sidebar.

aspiwack:
> FYI, I'm emerging from the void, and hoping to get back into some of this stuff this week. But I don't oppose your assistance.

Hi David! Glad to hear from you again.

My suggestion is to use this rebased branch for you and Thomas to coordinate. Since this branch is hosted on the main repository and you both have commit rights, it seems like the easier solution.

tbagrel1:
At this point, on this branch, compared to @treeowl 's one,
- [x] I rebased to master and resolved merge conflicts
- [x] I changed `D.xxx` to `Data.xxx` as suggested by @aspiwack 
- [x] I added the `{-# OPTIONS_HADDOCK hide #-}` pragmas

Here's the list of what might be required to merge this PR:
- [ ] Public export of the modules ?
- [ ] Write explicit export lists
- [ ] Add comments and documentation, especially at the public module level (a docspec example would be nice)
- [ ] See if we can use the latest release of `linear-generics` (0.2.0 if I'm not mistaken)

Feel free to amend this list, and select which items you want to tackle. @treeowl we can also set up a video call to sync up our work (email: firstname dot lastname at tweag dot io).

aspiwack:
> Write explicit export lists

You don't need explicit export lists on internal modules. But we ought to make sure that at least, all the classes which appear in the Haddock documentation is exported from somewhere non-internal. The `gtraverse` function ought to be exported from the same module which exports the `Traversable` type class.

tbagrel1:
I tried to address what I wrote in the TODO list yesterday.

The docspec that I added in `Data.Functor.Linear` is not working, because I don't know how to derive linear `Generic1` on a custom `data`. I might need your help for this @treeowl.

treeowl:
I expect to have time to look at it this evening.

tbagrel1:
Well, I still don't have a lot of luck with linear generics:
```
ghci> :{
ghci| data Pair a = MkPair a a
ghci| :}
ghci>  $(deriveGeneric1 ''Pair)

<interactive>:25:4: error:
    • Couldn't match type ‘[template-haskell-2.17.0.0:Language.Haskell.TH.Syntax.Dec]’
                     with ‘template-haskell-2.17.0.0:Language.Haskell.TH.Syntax.Exp’
      Expected: template-haskell-2.17.0.0:Language.Haskell.TH.Lib.Internal.ExpQ
        Actual: template-haskell-2.17.0.0:Language.Haskell.TH.Syntax.Q
                  [template-haskell-2.17.0.0:Language.Haskell.TH.Syntax.Dec]
    • In the expression: deriveGeneric1 ''Pair
      In the untyped splice: $(deriveGeneric1 ''Pair)
```
Did I do anything wrong? (using `linear-generics-0.2`)

treeowl:
I think GHCi may not support declaration splices at all. So you can do that in a module, but I think not in the interpreter. The `ViaGHCGenerics` stuff should work anywhere, but use the latest (major) version of `linear-generics` because it's different (much better) than what came before. Don't expect that deriving method to be too efficient, but it'll work for playing around just fine.

src/Data/Functor/Linear.hs
```diff
@@ -13,8 +13,27 @@
...
+-- >>> :set -XLinearTypes
+-- >>> :set -XTypeApplications
+-- >>> :set -XNoImplicitPrelude
+-- >>> :set -XDeriveAnyClass
+-- >>> import Prelude.Linear
+-- >>> import Data.Functor.Linear
+-- >>> import Generics.Linear
+-- >>> :{
+--  data Pair a = MkPair a a
+--    deriving (Show, Generic1)
```
aspiwack, line 31: I'm putting a note here to not forget to update this with the proper code.

I suggest writing an example in `examples/` directory.

src/Data/Functor/Linear.hs
```diff
@@ -13,8 +13,27 @@
...
+-- >>> :}
+-- >>> :{
+--  isNothing :: Maybe Int
+--  isNothing = genericTraverse (\x -> if x < 0 then Nothing else Just x) (MkPair 1 (-1))
```
aspiwack, line 35: It's better to make an example with the intended use of `genericTraverse`:

```suggestion
--  instance Traversable Pair where
--    traverse = genericTraverse
```

src/Data/Functor/Linear.hs
```diff
@@ -26,6 +45,9 @@ module Data.Functor.Linear
 
     -- * Linear traversable hierarchy
     Traversable (..),
+    genericTraverse,
+    gtraverse,
+    GTraversable,
```
aspiwack, line 50: I don't think that we want to export `gtraverse`
```suggestion
    genericTraverse,
    GTraversable,
```

aspiwack (commented):
I think that all we need is an example use in the `examples/` directory. Let's call it `Simple/Deriving.hs`.

Then we can merge.

examples/Generic/Traverse.hs
```diff
@@ -0,0 +1,64 @@
...
+data Pair a = MkPair a a
+  deriving (Show, Prelude.Eq)
+
+$(deriveGeneric1 ''Pair)
+
+data WithLog a = WithLog (Maybe String) a
+  deriving (Show, Prelude.Eq)
+
+instance Control.Functor WithLog where
+  fmap f (WithLog lx x) = WithLog (logApp Nothing lx) (f x)
```
aspiwack, line 33: For what it's worth, this doesn't verify the `Functor` laws, because, for instance, `fmap id` is not the identity (as it get something logged). You can get away with it by asserting that the log string is not actually meaningful.

Honestly, though, I'd just use an applicative from linear-base.

tbagrel1: But we don't have many `Control.Applicative` instances in linear-base. I only see `Identity`, `IO`, `ReaderT`, `StateT`, `StateR`, and `(,) a`. Which one would make a good example in your opinion?

treeowl: `(,) a` is a pretty obvious candidate!

aspiwack: Or `StateT Id`. But `(,) a` is likely to be the better choice.

examples/Generic/Traverse.hs
```diff
@@ -0,0 +1,64 @@
...
+logApp :: Maybe String %1 -> Maybe String %1 -> Maybe String
+logApp lf lx =
+  (fromMaybe "?fun?" lf, fromMaybe "?val?" lx) & \case
+    (lf', lx') -> Just $ "(" ++ lf' ++ " " ++ lx' ++ ")" where
+
+pairTest :: TestTree
+pairTest =
+  testProperty "genericTraverse with WithLog and Pair" $
+    property $
+      ( genericTraverse
```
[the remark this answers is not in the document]
aspiwack: Right. And then 
```suggestion
      ( traverse
```

examples/Generic/Traverse.hs
treeowl: Sorry, I got confused for a moment. I thought you were looking for a `Traversable` instance to derive. For an appropriate `Applicative` for testing, is there such a thing as a free control applicative? Free structures tend to make good test cases.

treeowl:
BTW, the reason I haven't hopped back into this as I promised is that I woke up good and sick Monday. Ugh.

aspiwack:
> BTW, the reason I haven't hopped back into this as I promised is that I woke up good and sick Monday. Ugh.

Oh :slightly_frowning_face: . I hope you are feeling better.

examples/Generic/Traverse.hs
aspiwack: We don't have one defined. We probably could, but on the other hand, it could not be printed or compared for equality. I don't think it would actually make for a good test case.

The goal of this file is primarily to serve as a code example for documentation purposes. I think that using writer is fine. It's a perfectly serviceable applicative structure.

aspiwack (approved):

PR #367, opened 2022-02-07
Add inspection tests to `Replicator.elim` and `V.{make,elim}`

tbagrel1:
`Replicator.elim`, `V.make`, and `V.elim` are implemented with classes so as to enable inlining. With inspection testing, we need to ensure that their final Core form are the same as a manually unrolled recursion.

### DIFF
--- a/test/Test/Data/Replicator.hs
+++ b/test/Test/Data/Replicator.hs
@@ -1,8 +1,6 @@
-{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# OPTIONS_GHC -O -dno-suppress-type-signatures -fplugin=Test.Tasty.Inspection.Plugin #-}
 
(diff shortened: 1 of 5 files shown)


tbagrel1:
For consistency, I added `INLINABLE` pragmas on every function that should either be inlined in both `{elim,make}` and `manual{Elim,Make}`, or not inlined in both contexts.

test/Main.hs
```diff
@@ -23,5 +25,7 @@ allTests =
       mutHMTests,
       mutSetTests,
       destArrayTests,
-      polarizedArrayTests
+      polarizedArrayTests,
+      vInspectionTests,
+      replicatorInspectionTests
```
aspiwack, line 30: Divide this list in two `testGroup`s one for data structure tests, and one for inspection tests.

test/Test/Data/Replicator.hs
```diff
@@ -0,0 +1,31 @@
...
+import Test.Tasty.Inspection
+
+replicatorInspectionTests :: TestTree
+replicatorInspectionTests =
+  testGroup
+    "Inspection testing of elim for Replicator"
+    [$(inspectTest $ 'elim3 === 'manualElim3)]
+
+elim3 :: (a %1 -> a %1 -> a %1 -> [a]) %1 -> Replicator a %1 -> [a]
+elim3 f r = Replicator.elim f r
```
aspiwack, line 22: Next PR, let's make sure that we can remove the `f` and the `r` and still have the inspection test pass.

aspiwack (approved):

PR #378, opened 2022-02-14
Add warnings to the cabal file

aspiwack:
This helps, in particular, when developping with ghcid, which
otherwise doesn't raise warnings.

I've had to deactivate `cabal` format in the formatting script due to
the fact that it inlines `common` stanzas, which I'm using to ensure
consistent warning settings in every target. See
[haskell/cabal#5734](https://github.com/haskell/cabal/issues/5734).

---

I've written this while I was trying to do something smart with the
`Elim` and `Make` type classes. It didn't work. So there won't be a
followup PR, but this one is worth it on its own, I think.

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -6,5 +6,9 @@ set -e
 export LANG="C.UTF-8"
 
 stack build ormolu
-cabal format
+## We can't format cabal at the moment because `cabal format` inlines
+## common stanzas, which is very much something that we don't want. See
+## https://github.com/haskell/cabal/issues/5734
+#
+# cabal format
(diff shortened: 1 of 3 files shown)


tbagrel1:
Well, I don't like the idea of removing `cabal format` (because it's nice to have alphabetically-sorted module names :sweat_smile: ); but adding common warnings seems to be more important. So let's do it (it's a pity that we don't have another solution at hand).

aspiwack:
> it's a pity that we don't have another solution at hand

Yet…

Let's be on the lookout.

Issue #190, opened 2020-09-18
Traversable with data applicatives


utdemir:
This discussion started at #126, where we discussed a proposed linear `Foldable` implementation and the fact that it is not implied by linear `Traversable`; and ended up concluding that what we need is a variant of the existing `Traversable` which uses data `Applicative`s instead of control `Applicative`s.

See the thread at #126 for a longer discussion, but I'm quoting the critical comment [there](https://github.com/tweag/linear-base/pull/126#issuecomment-668042694), from @aspiwack:
> We've established that being “traversable” by Control.Applicative-s is a common occurrence, but doesn't imply Foldable.
> 
> However, being “traversable” by Data.Applicative-s, a less common occurrence, does imply Foldable (indeed Const m is a Data.Applicative (for m a monoid)).
> 
> So, here is the thought. Why don't we give a name to the equivalent of traverse, but with Data.Applicative and add it to the Foldable type class (I'd call it fold but it's already taken, let's find another name)?

This issue to discuss whether we indeed want another `Traversable` using data `Applicative`s; and if we do, bikeshed the details (mainly the name, I guess).

aspiwack:
This wouldn't introduce backwards incompatible change. So I'll downgrade it to after-first-milestone status.

sjoerdvisscher:
Reagarding naming, since we have:

```Haskell
consumeFoldable :: (Consumable a, Foldable f) => f a %1 -> ()
consumeFoldable = foldMap consume

moveTraversable :: (Movable a, Traversable t) => t a %1 -> Ur (t a)
moveTraversable = traverse move
```

i.e. `Foldable` implies `Consumable` and `Traversable` using `Data.Applicative` implies `Movable`. So I think it makes sense to put `Foldable` and `Traversable` under `Data.Unrestricted`, which allows us to keep the names the same.

utdemir:
@treeowl That idea is seems to be similar to the one on the issue https://github.com/tweag/linear-base/issues/27. 

I personally would prefer an interface like that than the explosion of classes as we have right now. The only downside I can (without any evidence) imagine is this possibly being confusing to the type checker, I think it is worth experimenting with.

PR #381, opened 2022-02-15
`Semigroup` and `Monoid` overhaul

### LINKED ISSUE #372: `Semigroup` and `Monoid`: clean-up and improvements
- Add more instances of `Semigroup` and `Monoid` to mimic instances defined in `Data.Monoid` and `Data.Semigroup` in `base`
- Do not re-export the whole `Data.Semigroup` module in `Data.Monoid.Linear`, but rather, export only `newtype`s, and add linear `Semigroup` and `Monoid` instances to those types

### LINKED ISSUE #379: Dependency between some linear classes and their non-linear counterpart (e.g. `Semigroup`)

At the moment, `Data.Monoid.Linear.Semigroup` is defined as

```haskell
class Prelude.Semigroup a => Semigroup a where
  (<>) :: a %1 -> a %1 -> a
```

with a `Prelude.Semigroup` superclass constraint. Theorically, this is correct. Indeed, any linear semigroup can be made into a non-linear semigroup just by upcasting `(<>) :: a %1 -> a %1 -> a` to `(<>) :: a -> a -> a`. A `NonLinear` newtype also exists in `Data.Monoid.Linear.Internal.Semigroup` to easily derive `Prelude.Semigroup` from (linear) `Semigroup`.

However, things get harder when we try to write a (linear) `Semigroup` instance for `Ap f a`. In `base`, we have `instance (Prelude.Applicative f, Prelude.Semigroup a) => Prelude.Semigroup (Ap f a)`. In `linear-base`, we would like to have `instance (Data.Functor.Linear.Applicative f, Data.Monoid.Linear.Semigroup a) => Data.Monoid.Linear.Semigroup (Ap f a)`.

But such an instance head requires `Prelude.Semigroup (Ap f a)` (because of the `Prelude.Semigroup => Data.Monoid.Linear.Semigroup` relationship), which is not deducible because `f` is a `Data.Functor.Linear.Applicative`, not a `Prelude` (non-linear) one.

Unfortunately, we can't add `instance (Data.Functor.Linear.Applicative f, Prelude.Semigroup a) => Prelude.Semigroup (Ap f a)`, because this has the same head as `instance (Prelude.Applicative f, Prelude.Semigroup a) => Prelude.Semigroup (Ap f a)` already present in `base`.

Also, `Semigroup` and `Monoid` are the only linear classes which have their non-linear counterpart as superclasses, although it would also make sense for `Ord` or `Eq` for example.

I see three potential solutions:

+ Breaking the `Prelude.Semigroup => Data.Monoid.Linear.Semigroup` and `Prelude.Monoid => Data.Monoid.Linear.Monoid` relationships, allowing us to add the `instance (Data.Functor.Linear.Applicative f, Data.Monoid.Linear.Semigroup a) => Data.Monoid.Linear.Semigroup (Ap f a)`
+ *OR* keeping them, but adding the same kind of reliationship for `Eq`, `Ord`, and any eligible linear class, and then
  * Not exporting the `Ap` newtype
  * *OR* adding a `LinAp` newtype

tbagrel1:
+ Break relationship between `Data.Monoid.Linear.{Semigroup,Monoid}` and their
non-linear counterparts (see https://github.com/tweag/linear-base/issues/379)
+ Add as many instances as possible for these classes (to replicate what is done
  in `base`

Closes #379. Closes #372.

### DIFF
--- a/src/Data/Array/Polarized.hs
+++ b/src/Data/Array/Polarized.hs
@@ -129,7 +129,16 @@ import Prelude.Linear
 -- NOTE: this does NOT require allocation and can be in-lined.
 transfer :: Pull.Array a %1 -> Push.Array a
 transfer (Pull.Array f n) =
-  Push.Array (\k -> NonLinear.foldMap' (\x -> k (f x)) [0 .. (n - 1)])
+  -- 'transfer' was
+  -- > transfer (Pull.Array f n) =
+  -- >   Push.Array (\k -> NonLinear.foldMap' (\x -> k (f x)) [0 .. (n - 1)])
+  -- but 'Linear.Monoid' no longer implies 'NonLinear.Monoid'. So we can have
+  -- @mempty :: a@ and @(<>) :: a -> a -> a@ (by degrading 'Linear.<>'), but we
+  -- no longer have the 'NonLinear.Monoid' instance required to use
+  -- 'NonLinear.foldMap\''. As a result, we just expand 'foldMap\'' to its
+  -- definition in terms of 'foldl\'', which doesn't require 'NonLinear.Monoid':
+  -- > foldMap' f' = foldl' (\acc a -> acc <> f' a) mempty
+  Push.Array (\k -> NonLinear.foldl' (\acc a -> acc <> k (f a)) mempty [0 .. (n - 1)])
 
 -- | This is a shortcut convenience function
 -- for @transfer . Pull.fromVector@.
--- a/src/Data/Functor/Linear/Internal/Applicative.hs
+++ b/src/Data/Functor/Linear/Internal/Applicative.hs
@@ -1,5 +1,9 @@
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 {-# OPTIONS_HADDOCK hide #-}
 
 module Data.Functor.Linear.Internal.Applicative
@@ -12,6 +16,7 @@ import Data.Functor.Compose
 import Data.Functor.Const
 import Data.Functor.Identity
 import Data.Functor.Linear.Internal.Functor
+import Data.Monoid (Ap (..))
 import Data.Monoid.Linear hiding (Sum)
 import Prelude.Linear.Internal
 
@@ -57,14 +62,6 @@ class Functor f => Applicative f where
 -- # Instances
 -------------------------------------------------------------------------------
 
-instance Monoid x => Applicative (Const x) where
-  pure _ = Const mempty
-  Const x <*> Const y = Const (x <> y)
-
-instance Monoid a => Applicative ((,) a) where
-  pure x = (mempty, x)
-  (u, f) <*> (v, x) = (u <> v, f x)
-
 instance Applicative Identity where
   pure = Identity
   Identity f <*> Identity x = Identity (f x)
@@ -77,3 +74,17 @@ instance (Applicative f, Applicative g) => Applicative (Compose f g) where
 instance Applicative m => Applicative (NonLinear.ReaderT r m) where
   pure x = NonLinear.ReaderT (\_ -> pure x)
   NonLinear.ReaderT f <*> NonLinear.ReaderT x = NonLinear.ReaderT (\r -> f r <*> x r)
+
+instance (Applicative f, Semigroup a) => Semigroup (Ap f a) where
+  (Ap x) <> (Ap y) = Ap $ liftA2 (<>) x y
+
+instance (Applicative f, Monoid a) => Monoid (Ap f a) where
+  mempty = Ap $ pure mempty
+
+instance Monoid x => Applicative (Const x) where
+  pure _ = Const mempty
+  Const x <*> Const y = Const (x <> y)
+
+instance Monoid a => Applicative ((,) a) where
+  pure x = (mempty, x)
+  (u, f) <*> (v, x) = (u <> v, f x)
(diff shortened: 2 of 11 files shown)


src/Data/Array/Polarized/Push.hs
```diff
@@ -39,7 +39,8 @@ import qualified Prelude
 -- | Push arrays are un-allocated finished arrays. These are finished
 -- computations passed along or enlarged until we are ready to allocate.
 data Array a where
-  Array :: (forall m. Monoid m => (a -> m) -> m) %1 -> Array a
+  -- TODO: do we really need both 'Prelude.Monoid' and 'Monoid' as constraints?
```
tbagrel1, line 42: This seems to be the only drawback from breaking the relationship between `Data.Monoid.Monoid` and `Data.Monoid.Linear.Monoid`

aspiwack: You don't need to do that, I think. In fact, I think that we can use `Prelude.Monoid` only, here, as the elements are presumed to be unrestricted.

src/Data/Functor/Linear/Internal/Applicative.hs
```diff
@@ -9,9 +13,10 @@ where
 
 import qualified Control.Monad.Trans.Reader as NonLinear
 import Data.Functor.Compose
-import Data.Functor.Const
+import Data.Functor.Const (Const (..))
```
aspiwack, line 16: When there is a conflict, I'd rather us use `hiding (conflicting thing)`. It's more robust.

src/Data/Monoid/Linear/Internal/Semigroup.hs
```diff
@@ -15,29 +15,56 @@ module Data.Monoid.Linear.Internal.Semigroup
...
+import Prelude (Either (..), Maybe (..))
 
 -- | A linear semigroup @a@ is a type with an associative binary operation @<>@
 -- that linearly consumes two @a@s.
-class Prelude.Semigroup a => Semigroup a where
+--
+-- Laws (same as 'Data.Semigroup.Semigroup'):
+--   * ∀ x ∈ G, y ∈ G, z ∈ G, x <> (y <> z) = (x <> y) <> z
+class Semigroup a where
+  {-# MINIMAL (<>) #-}
```
aspiwack, line 63: You don't need a `MINIMAL` pragma when there is a single method.
```suggestion
```

src/Data/Num/Linear.hs
```diff
@@ -150,35 +151,47 @@ liftU2 f x y = lifted f (move x) (move y)
     lifted :: (a -> b -> c) %1 -> (Ur a %1 -> Ur b %1 -> c)
     lifted g (Ur a) (Ur b) = g a b
 
+-- TODO: deprecate
```
aspiwack, line 154: Do before merge :slightly_smiling_face:

src/System/IO/Resource/Internal.hs
```diff
@@ -96,6 +94,12 @@ instance Control.Monad RIO where
     unRIO x releaseMap
     unRIO y releaseMap
 
+instance Semigroup a => Semigroup (RIO a) where
+  (<>) = Control.liftA2 (<>)
+
+instance Monoid a => Monoid (RIO a) where
+  mempty = Control.pure mempty
```
aspiwack, line 101: Use deriving via `Ap`?

Something like
```
  deriving (Semigroup, Monoid) via (Ap RIO)
```
on the definition of `RIO`.

src/Data/Array/Polarized/Push.hs
tbagrel1: In fact, we need a linear `Monoid` constraint here, or else we break linear guarantees in `cons` and `snoc`. But I found a workaround.

src/Data/Functor/Linear/Internal/Applicative.hs
tbagrel1: There seems to be no conflict after all

tbagrel1:
Note for #377  #393 : `Control.Optics.Linear.lengthOf` signature changed (moving from `Adding` to `Sum`)

src/Data/Array/Polarized.hs
```diff
@@ -129,7 +129,16 @@ import Prelude.Linear
...
-  Push.Array (\k -> NonLinear.foldMap' (\x -> k (f x)) [0 .. (n - 1)])
+  -- 'transfer' was
+  -- > transfer (Pull.Array f n) =
+  -- >   Push.Array (\k -> NonLinear.foldMap' (\x -> k (f x)) [0 .. (n - 1)])
+  -- but 'Linear.Monoid' no longer implies 'NonLinear.Monoid'. So we can have
+  -- @mempty :: a@ and @(<>) :: a -> a -> a@ (by degrading 'Linear.<>'), but we
+  -- no longer have the 'NonLinear.Monoid' instance required to use
+  -- 'NonLinear.foldMap\''. As a result, we just expand 'foldMap\'' to its
+  -- definition in terms of 'foldl\'', which doesn't require 'NonLinear.Monoid':
+  -- > foldMap' f' = foldl' (\acc a -> acc <> f' a) mempty
```
aspiwack, line 140: As a side note we still have that linear monoids have a corresponding classical monoid.

So we could have
```haskell
newtype Linear a = Linear a

instance Semigroup a => Linear.Semigroup (Linear a) where
  x <> y = x <>  y
instance Monoid a => Linear.Monoid (Linear a) where
  mempty = mempty
```

Then you could in principle do

```haskell
case NonLinear.Foldmap' (\x -> Linear (k (f x)) [0 .. (n - 1)] of
  Linear r -> r
```

It's not better than a simple `foldl'`, but it's worth being aware of the possibility (in an ideal world you would be able to use `coerce`, but `coerce` isn't linear).

---

There is also the option of using a linear `foldMap'`, but we don't have it available until we solve the Foldable issue (_e.g._ #190 ).

src/Data/Array/Polarized/Push.hs
aspiwack: Indeed, my bad.

aspiwack (approved):

PR #382, opened 2022-02-15
Force eta-expansion of `{Replicator,V}.elim`

### LINKED ISSUE #369: Eta-expansion of `elim`
See https://github.com/tweag/linear-base/pull/367#discussion_r802767909

tbagrel1:
`Replicator.elim` and `V.elim` are currently not inlined if they are not fully applied (as we found out when writing inspection tests for this method). For example, in the following case, `elim` doesn't get inlined:

```haskell
elim3 :: (a %1 -> a %1 -> a %1 -> [a]) %1 -> V 3 a %1 -> [a]
elim3 = V.elim
```

This PR renames `elim` to `elim'` and add a `elim` wrapper ensuring that `elim'` can always get inlined.

(I also move `make` into the `Make` class in this PR, for consistency with `Elim`)

Closes #369 .

### DIFF
--- a/src/Data/V/Linear/Internal.hs
+++ b/src/Data/V/Linear/Internal.hs
@@ -112,31 +112,29 @@ class (n ~ Arity b f) => Elim n a b f | n a b -> f, f b -> n where
   -- > elim @2 :: (a %1 -> a %1 -> b) %1 -> V 2 a %1 -> b
   -- > elim @3 :: (a %1 -> a %1 -> a %1 -> b) %1 -> V 3 a %1 -> b
   elim :: f %1 -> V n a %1 -> b
+  elim f v = elim' f v
+  {-# INLINEABLE elim #-}
+
+  -- 'elim' is used to force eta-expansion of 'elim\''.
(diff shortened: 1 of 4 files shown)


src/Data/V/Linear/Internal.hs
```diff
@@ -112,32 +112,30 @@ class (n ~ Arity b f) => Elim n a b f | n a b -> f, f b -> n where
   -- > elim @2 :: (a %1 -> a %1 -> b) %1 -> V 2 a %1 -> b
   -- > elim @3 :: (a %1 -> a %1 -> a %1 -> b) %1 -> V 3 a %1 -> b
   elim :: f %1 -> V n a %1 -> b
+  elim f v = elim' f v
+  {-# INLINEABLE elim #-}
```
tbagrel1, line 116: For obscure reasons, the `{-# INLINABLE #-}` pragma is necessary for `V.elim` (but not `Replicator.elim`). Also, neither `INLINE` nor `NOINLINE` work in this context. It seems that the contexts where something `INLINABLE` gets inlined are not completely contained in the contexts where something `INLINE` gets inlined.

PR #383, opened 2022-02-15
Improve consistency of module naming

### LINKED ISSUE #373: Rename `Data/Arity/Linear/Arity.hs` to `Data/Arity/Linear.hs` for consistency
Everything is in the title.

### LINKED ISSUE #380: Move internal modules of `Data.Unrestricted` under `Data.Unrestricted.Linear`

At the moment, internal modules of `Data.Unrestricted.Linear` are store in `src/Data/Unrestricted/Internal` (instead of `src/Data/Unrestricted/Linear/Internal`).

Moving them to `src/Data/Unrestricted/Linear/Internal`. At the moment, it is confusing when someone wants to import the internal module instead of the public one in `linear-base` code, because the public one is not a prefix of the internal one.

tbagrel1:
Before 1.0 release, I think that we should ensure that our naming conventions are respected for both internal and public modules.
In this PR, I renamed:

+ System.IO.Resource -> System.IO.Resource.Linear
+ Streaming.Internal -> System.Linear.Internal
+ Data.Unrestricted.Internal -> Data.Unrestricted.Linear.Internal
+ Data.Arity.Linear.Internal.Arity -> Data.Arity.Linear.Internal

The `Control.Optics` directory structure seemed a bit odd to me (but I haven't changed it yet), because it contains modules not ending with either `.Linear` or `.Linear.Internal.xxx` as this is the case in every other namespace. Should we rename something for this one?

Closes #373 . Closes #380 .

### DIFF
--- a/docs/DESIGN.md
+++ b/docs/DESIGN.md
@@ -14,7 +14,7 @@ Linear base consists of the following:
 * tools ported from [`base`] and from other critical haskell
   libraries, like `lens`,
 * new APIs for using system resources, e.g., file I/O in
-  [`System.IO.Resource`],
+  [`System.IO.Resource.Linear`],
 * new abstractions made possible by linear types, like monad-free
   mutable arrays in ([`Data.Array.Mutable.Linear`]).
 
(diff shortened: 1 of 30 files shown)


docs/DESIGN.md
```diff
@@ -14,7 +14,7 @@ Linear base consists of the following:
 * tools ported from [`base`] and from other critical haskell
   libraries, like `lens`,
 * new APIs for using system resources, e.g., file I/O in
-  [`System.IO.Resource`],
+  [`System.IO.Resource.Linear`],
```
aspiwack, line 17: This one is a breaking change. We will have to communicate on it.

tbagrel1: We need to write it down in the changelog, I suppose? Do we need to do something else?

(I am referencing the changelog issue there to not forget: #377 )

aspiwack: Yes, in the change log. All the breaking change will have to be clearly identified.

PR #384, opened 2022-02-15
Some documentation on genericTraverse and friends

aspiwack:
Just small docstring fixes.

@sjoerdvisscher, you pioneered the use of `Yoneda` and `Curried` for
fusion. Could you write the missing bit of documentation?

### DIFF
--- a/src/Control/Functor/Linear/Internal/Kan.hs
+++ b/src/Control/Functor/Linear/Internal/Kan.hs
@@ -6,7 +6,7 @@
 
 -- | A few things lifted from kan-extensions and lens for generic deriving of
 -- 'Data.Functor.Linear.Traversable' instances (see
--- 'Data.Functor.Linear.Internal.Generic').
+-- "Data.Functor.Linear.Internal.Traversable.Generic").
 module Control.Functor.Linear.Internal.Kan where
 
 import Control.Functor.Linear
--- a/src/Data/Functor/Linear/Internal/Traversable/Generic.hs
+++ b/src/Data/Functor/Linear/Internal/Traversable/Generic.hs
@@ -17,8 +17,14 @@ import GHC.Types (Multiplicity (..))
 import Generics.Linear
 import Prelude.Linear.Internal
 
--- | Linear counterpart of [@GTraversable@](https://hackage.haskell.org/package/traverse-with-class-1.0.1.1/docs/Data-Generics-Traversable.html#t:GTraversable).
+-- | This type class derives the definition of 'genericTraverse' by induction on
+-- the generic representation of a type.
 class GTraversable t where
+  -- gtraverse :: Applicative f => (a %1 -> f b) -> t a %1 -> forall r. (forall k. ((a %1 -> r) %1 -> k) %1 -> f k) %1 -> forall k. (t b %1 -> k) %1 -> f k
+  --
+  -- TODO: developer documentation on why we use this type rather than the more
+  -- straightforward type of `traverse`. Used, for instance, in the
+  -- generic-deriving package.
   gtraverse :: Applicative f => (a %1 -> f b) -> t a %1 -> Curried (Yoneda f) (Yoneda f) (t b)
 
 instance GTraversable t => GTraversable (M1 i c t) where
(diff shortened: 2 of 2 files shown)


src/Data/Functor/Linear/Internal/Traversable/Generic.hs
```diff
@@ -84,6 +90,19 @@ instance GTraversable UWord where
 
 -- | Implementation of 'Data.Functor.Linear.traverse' for types which derive
 -- (linear) 'Generics.Linear.Generic1'.
+--
+-- Use as
+--
+-- > data T
+-- > $(deriveGeneric1 ''Pair)
+-- >
+-- > instance Traversable T
```
sjoerdvisscher, line 99:
```suggestion
-- > instance Traversable T where
```

aspiwack:
@treeowl 
> I'm the one who wrote that

Indeed, I now realise that #316 was coauthored :slightly_smiling_face: (I'd like to bring it up to master, now, but it's a non-trivial amount of work to merge master in it. I don't really have time to do so, would you?).

Regarding this PR: I don't necessarily ask for a detailed explanation (though it would be welcome), but enough to get an intuition of why these types specifically?

treeowl:
I will work on bringing that branch (or its main ideas) up to `master` as soon as I can. I'm still getting wiped out by this cold that won't let me sleep. I will try to write an explanation of that, however brief, today.

treeowl:
I wrote a little documentation for `Curried` in #385. It's not great; hopefully someone will be able to flesh it out later.

src/Data/Functor/Linear/Internal/Traversable/Generic.hs
```diff
@@ -84,6 +90,19 @@ instance GTraversable UWord where
 
 -- | Implementation of 'Data.Functor.Linear.traverse' for types which derive
 -- (linear) 'Generics.Linear.Generic1'.
+--
+-- Use as
```
treeowl, line 94:
```suggestion
-- ### Example
```

src/Data/Functor/Linear/Internal/Traversable/Generic.hs
```diff
@@ -84,6 +90,19 @@ instance GTraversable UWord where
...
+--
+-- > data T
+-- > $(deriveGeneric1 ''Pair)
+-- >
+-- > instance Traversable T where
+-- >   traverse = genericTraverse
+--
+-- Note that, contrary to many other classes in linear-base, we can't define
+-- `Traversable T` using deriving via, because the
+-- [role](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/roles.html)
```
treeowl, line 104: The link for `role` may be hard to hit on mobile browsers and other touchscreens, since it's so short. How much does that matter?

aspiwack: I think it matters a little, but not a ton (you can always zoom on mobile). So if you have a good solution, I'll take it. But let's not spend a lot of time on it.

src/Control/Functor/Linear/Internal/Kan.hs
```diff
@@ -6,7 +6,7 @@
 
 -- | A few things lifted from kan-extensions and lens for generic deriving of
 -- 'Data.Functor.Linear.Traversable' instances (see
--- 'Data.Functor.Linear.Internal.Generic').
+-- "Data.Functor.Linear.Internal.Generic.Traversable").
```
tbagrel1, line 9:
```suggestion
-- "Data.Functor.Linear.Internal.Traversable.Generic").
```

src/Data/Functor/Linear/Internal/Traversable/Generic.hs
```diff
@@ -84,6 +90,19 @@ instance GTraversable UWord where
 
 -- | Implementation of 'Data.Functor.Linear.traverse' for types which derive
 -- (linear) 'Generics.Linear.Generic1'.
+--
+-- ### Example
+--
+-- > data T
+-- > $(deriveGeneric1 ''Pair)
```
tbagrel1, line 97:
```suggestion
-- > $(deriveGeneric1 ''T)
```

tbagrel1 (approved):
Let's just fix these two typos.

PR #385, opened 2022-02-16
A bit of Curried documentation

treeowl:
Not the best explanation I've ever written, but a lot of this
has fallen out of my cache and it is quite confusing to begin
with.


aspiwack (approved):
The ending is a bit abrupt. But otherwise, it's very informative, thanks.

PR #387, opened 2022-02-16
Add `void` function to consume `Control.Functor.Linear.Functor` inner value

tbagrel1:
Apply decision of PR #356 : we should have a `void :: (Functor f, Consumable a) =>  f a %1 -> f ()` function not only for `Data.Functor.Linear.Functor` but for `Control.Functor.Linear.Functor` as well.

Closes #356.

### DIFF
--- a/src/Control/Functor/Linear.hs
+++ b/src/Control/Functor/Linear.hs
@@ -14,6 +14,7 @@ module Control.Functor.Linear
     (<$>),
     (<&>),
     (<$),
+    void,
     dataFmapDefault,
     Applicative (..),
     dataPureDefault,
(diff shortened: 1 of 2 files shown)


aspiwack:
> PR #356 might be closed when this PR is merged.

I wonder if we can use the “closes X” action on PRs.

aspiwack:
> I wonder if we can use the “closes X” action on PRs.

Apparently we can. Fancy that!

tbagrel1:
I tried yesterday to link the PR using the Github UI; but failed. However, it seems that we can do it with the `Closes` keyword.

PR #390, opened 2022-02-17
A more robust implementation of Replicator.elim

### LINKED ISSUE #388: Data.Replicator.Linear.Internal doesn't compile under GHC 9.2
**Describe the bug**

```
> cabal configure --with-ghc=ghc-9.2
...
> cabal build
...
src/Data/Replicator/Linear/Internal.hs:169:18: error:
    • Could not deduce (Elim (Arity b (a %1 -> f)) a b f1)
        arising from a use of ‘elim'’
      from the context: (n ~ Arity b (a %1 -> f1), Elim (n - 1) a b f1)
        bound by the instance declaration
        at src/Data/Replicator/Linear/Internal.hs:166:31-101
      The type variable ‘f’ is ambiguous
      Relevant bindings include
        r' :: Replicator a
          (bound at src/Data/Replicator/Linear/Internal.hs:169:11)
        a :: a (bound at src/Data/Replicator/Linear/Internal.hs:169:8)
        r :: Replicator a
          (bound at src/Data/Replicator/Linear/Internal.hs:167:11)
        g :: a %1 -> f1
          (bound at src/Data/Replicator/Linear/Internal.hs:167:9)
        elim' :: (a %1 -> f1) %1 -> Replicator a %1 -> b
          (bound at src/Data/Replicator/Linear/Internal.hs:167:3)
    • In the expression: elim' (g a) r'
      In a case alternative: (a, r') -> elim' (g a) r'
      In the second argument of ‘(&)’, namely
        ‘\case (a, r') -> elim' (g a) r'’
    |
169 |       (a, r') -> elim' (g a) r'
    |                  ^^^^^
...
```

**Environment**
 - OS name + version:
 - Version of the code: master branch

**Additional context**

aspiwack:
I rely on more type families (I resurrected the `FunN` type family,
for instance) and additional type classes in order to give robust
hints to the type checker to infer some of the arguments based on the
others in multiple possible ways.

The previous version was abusing functional dependencies to that
effect. The problem is that it almost certainly depended on bugs. Plus
some behaviour changed in 9.2 making the recursive instance not
typecheck anymore.

The actual (staging) recursion now uses (type-level) Peano integers,
this way we can get rid of the `OVERLAPPABLE` instances altogether,
and avoid the sort of fragile behaviour that caused the failure in
9.2.

The 'elim' function itself has more type arguments, so I took it out
of the type class. A collection of type families and type classes fill
the role that functional dependencies used to play: figuring out `f`
from `n` or `n` from `f`.

Fixes #388.

---

Interestingly I've only had to fix `Replicator.elim` for the sake of
fixing the compilation in GHC 9.2. The rest works fine. But it'd be
better to make similar changes to `V.make` and `V.elim`. I won't have
time to, though, @tbagrel1 can I untrust you with this?

I've made some additional small changes for the sake of GHC 9.2
compilation:

### Some 𝜂-expansions to compile with GHC 9.2

This is really suspicious, there may be a bug in GHC 9.2 where
sections do not preserve linearity. Grmpf.

### DIFF
--- a/src/Data/Arity/Linear/Internal.hs
+++ b/src/Data/Arity/Linear/Internal.hs
@@ -1,13 +1,56 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LinearTypes #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE StandaloneKindSignatures #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 
-module Data.Arity.Linear.Internal (Arity) where
+module Data.Arity.Linear.Internal where
 
+import Data.Kind
 import GHC.TypeLits
+import GHC.Types
 
+data Peano = Z | S Peano
+
+type NatToPeano :: Nat -> Peano
+type family NatToPeano n where
+  NatToPeano 0 = 'Z
+  NatToPeano n = 'S (NatToPeano (n - 1))
+
+type PeanoToNat :: Peano -> Nat
+type family PeanoToNat n where
+  PeanoToNat 'Z = 0
+  PeanoToNat ('S n) = 1 + PeanoToNat n
+
+type FunN :: Peano -> Type -> Type -> Type
+type family FunN n a b where
+  FunN 'Z _ b = b
+  FunN ('S n) a b = a %1 -> FunN n a b
+
+-- | The 'Arity' type family exists to help the type checker fill in
+-- blanks. Chances are that you can safely ignore 'Arity' completely if it's in
+-- the type of a function you care. But read on if you are curious.
+--
+-- The idea is that in a function like 'Data.Replicator.Linear.elim' some of the
+-- type arguments are redundant. The function has an ambiguous type, so you will
+-- always have to help the compiler either with a type annotation or a type
+-- application. But there are several complete ways to do so. In
+-- 'Data.Replicator.Linear.elim', if you give the values of `n`, `a`, and `b`,
+-- then you can deduce the value of `f` (indeed, it's @'FunN' n a b@). With
+-- 'Arity' we can go in the other direction: if `b` and `f` are both known, then
+-- we know that `n` is @'Arity' b f@
+--
+-- 'Arity' returns a 'Nat' rather than a 'Peano' because the result is never
+-- consumed. It exists to infer arguments to functions such as
+-- 'Data.Replicator.Linear.elim' from the other arguments if they are known.
+--
+-- 'Arity' could be an associated type family to the 'IsFunN' type class. But
+-- it's better to make it a closed type family (which can't be associated to a
+-- type class) because it lets us give a well-defined error case.
+type Arity :: Type -> Type -> Nat
 type family Arity b f where
   Arity b b = 0
   Arity b (a %1 -> f) = Arity b f + 1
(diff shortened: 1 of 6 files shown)


tbagrel1:
I will take a look at it tomorrow morning!

src/Data/Unrestricted/Linear/Internal/Dupable.hs
```diff
@@ -57,7 +58,7 @@ class Consumable a => Dupable a where
 
   -- | Creates two @a@s from a @'Dupable' a@, in a linear fashion.
   dup2 :: a %1 -> (a, a)
-  dup2 x = Replicator.elim (,) (dupR x)
+  dup2 x = Replicator.elim @2 (,) (dupR x)
```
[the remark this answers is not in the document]
aspiwack: Oh, yes, sorry. This is debugging code that I forgot to undo…

src/Data/Arity/Linear/Internal.hs
```diff
@@ -19,3 +62,26 @@ type family Arity b f where
           ':<>: 'ShowType b
```
tbagrel1, line 62: But now, `Arity` seems to work on non-linear functions (because of `%p ->`), so maybe we should change the error message?

src/Data/Arity/Linear/Internal.hs
```diff
@@ -19,3 +62,26 @@ type family Arity b f where
...
+-- The reason why 'Arity' (read its documentation first) is not sufficient for
+-- our purpose, is that it can find `n` /if/ `f` is a linear function of the
+-- appropriate shape. But what if `f` is partially undetermined? Then it is
+-- likely that 'Arity' will be stuck. But we know, for instance, that if `f = a1
+-- %1 -> a2 %1 -> c` then we must have `a1 ~ a2`. The trick is that instance
+-- resolution of 'IsFun' will add unification constraints that the type checker
+-- has to solve. Look in particular at the instance @'IsFunN' a b (a' %p ->
+-- f))@: it matches liberally, so triggers on quite underdetermined `f`, but has
+-- equality constraints in its context which will help the type checker.
+class IsFunN a b f
```
tbagrel1, line 83: Couldn't we use something like
```haskell
class IsFunN a b f where
  type Arity

instance IsFunN a b b where
  type Arity = 'Z

instance (IsFunN a b f, a' ~ a, p ~ 'One) => IsFunN a b (a' %p -> f) where
  type Arity = 'S (Arity @a @b @f)
```
to get rid of the explicit `Arity` type family ?

tbagrel1 (commented):
I just have questions at the moment (see review comments)

src/Data/Arity/Linear/Internal.hs
treeowl: Why?

tbagrel1: I like the idea of having one class `FunN` to infer `f` from `n a b`, and one class `IsFunN` to infer `n` from `a b f`. Cheking `IsFunN a b f` should allow to compute the arity along the way (at least theorically).

src/Data/Arity/Linear/Internal.hs
aspiwack: This, as well, is debugging code which was left over…

src/Data/Arity/Linear/Internal.hs
aspiwack: Did you manage to answer your own question? if so, I'm just waiting for the green tick so that I can merge.

tbagrel1: Yes, sorry

tbagrel1 (approved):
After some time to process it, it looks good to me.

src/Data/Arity/Linear/Internal.hs
aspiwack: > I thought that you wanted me to continue the work on this branch.
I'll pretty never be asking you to continue work on a branch which passes tests :slightly_smiling_face: . If something is mergeable as it is, then we merge it. Merging small chunks early makes everybody's life easier.

PR #391, opened 2022-02-21
Finish robustness improvements on `make`/`elim`

tbagrel1:
+ Rework `V.{make,elim}` to use the same type-level machinery as
  the new `Replicator.elim` implementation introduced by @aspiwack in https://github.com/tweag/linear-base/pull/390
+ Improve documentation of `V.{make,elim}`, `Replicator.elim` and `Arity` module
+ Expose the type-level helpers from the internal `Arity` module as `Data.Arity.Linear`

### DIFF
--- a/src/Data/Arity/Linear/Internal.hs
+++ b/src/Data/Arity/Linear/Internal.hs
@@ -4,8 +4,10 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE StandaloneKindSignatures #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_HADDOCK hide #-}
 
 module Data.Arity.Linear.Internal where
 
@@ -15,16 +17,20 @@ import GHC.Types
 
 data Peano = Z | S Peano
 
+-- | Converts a GHC type-level 'Nat' to a structural type-level natural ('Peano').
 type NatToPeano :: Nat -> Peano
 type family NatToPeano n where
   NatToPeano 0 = 'Z
   NatToPeano n = 'S (NatToPeano (n - 1))
 
+-- | Converts a structural type-level natural ('Peano') to a GHC type-level 'Nat'.
 type PeanoToNat :: Peano -> Nat
 type family PeanoToNat n where
   PeanoToNat 'Z = 0
   PeanoToNat ('S n) = 1 + PeanoToNat n
 
+-- | @'FunN' n a b@ represents a function taking @n@ linear arguments of
+-- type @a@ and returning a result of type @b@.
 type FunN :: Peano -> Type -> Type -> Type
 type family FunN n a b where
   FunN 'Z _ b = b
@@ -47,9 +53,11 @@ type family FunN n a b where
 -- consumed. It exists to infer arguments to functions such as
 -- 'Data.Replicator.Linear.elim' from the other arguments if they are known.
 --
--- 'Arity' could be an associated type family to the 'IsFunN' type class. But
--- it's better to make it a closed type family (which can't be associated to a
--- type class) because it lets us give a well-defined error case.
+-- 'Arity' could /theorically/ be an associated type family to the 'IsFunN' type
+-- class. But it's better to make it a closed type family (which can't be
+-- associated to a type class) because it lets us give a well-defined error
+-- case. In addition, GHC cannot see that @0 /= 1 + (? :: Nat)@ and as a result we get
+-- some overlap which is only allowed in (ordered) closed type families.
 type Arity :: Type -> Type -> Nat
 type family Arity b f where
   Arity b b = 0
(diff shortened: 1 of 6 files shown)


src/Data/Replicator/Linear/Internal.hs
```diff
@@ -133,13 +133,26 @@ extend f = map f . duplicate
 --
 -- > elim (,) :: Replicator a %1 -> (a, a)
 -- > elim (,,) :: Replicator a %1 -> (a, a, a)
-elim :: forall (n :: Nat) a b f. (Elim (NatToPeano n) a b, IsFunN a b f, f ~ FunN (NatToPeano n) a b, n ~ Arity b f) => f %1 -> Replicator a %1 -> b
+elim ::
+  forall (n :: Nat) a b f.
+  ( -- Always fulfilled. Provides the actual implementation of 'elim' in
```
tbagrel1, line 138: Unfortunately, it seems that we can't add haddock comments on constraints :sad:

src/Data/Arity/Linear/Internal.hs
```diff
@@ -47,18 +53,20 @@ type family FunN n a b where
...
+-- 'Arity' could /theorically/ be an associated type family to the 'IsFunN' type
+-- class. But it's better to make it a closed type family (which can't be
+-- associated to a type class) because it lets us give a well-defined error
+-- case. Also, GHC cannot see that @0 /= 1 + (? :: Nat)@ and as a result we get
+-- some overlap which is only allowed in (ordered) closed type families.
 type Arity :: Type -> Type -> Nat
 type family Arity b f where
   Arity b b = 0
-  Arity b (a %1 -> f) = Arity b f + 1
+  Arity b (a %p -> f) = Arity b f + 1
```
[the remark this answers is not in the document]
aspiwack: You branch before I pushed a fix to said commit. You can simply remove the offending commit and force-push.

src/Data/Arity/Linear/Internal.hs
```diff
@@ -47,18 +53,20 @@ type family FunN n a b where
 -- consumed. It exists to infer arguments to functions such as
 -- 'Data.Replicator.Linear.elim' from the other arguments if they are known.
 --
--- 'Arity' could be an associated type family to the 'IsFunN' type class. But
--- it's better to make it a closed type family (which can't be associated to a
--- type class) because it lets us give a well-defined error case.
+-- 'Arity' could /theorically/ be an associated type family to the 'IsFunN' type
+-- class. But it's better to make it a closed type family (which can't be
+-- associated to a type class) because it lets us give a well-defined error
+-- case. Also, GHC cannot see that @0 /= 1 + (? :: Nat)@ and as a result we get
```
aspiwack, line 59:
```suggestion
-- case. In addition, GHC cannot see that @0 /= 1 + (? :: Nat)@ and as a result we get
```

src/Data/Replicator/Linear/Internal.hs
aspiwack: What do you mean by “always fulfilled”?

src/Data/Replicator/Linear/Internal.hs
```diff
@@ -133,13 +133,26 @@ extend f = map f . duplicate
...
+    -- terms of @elim\'@
+    Elim (NatToPeano n) a b,
+    -- Asserts that @f@ is a function taking @n@ linear arguments of type
+    -- @a@ and returning a value of type @b@
+    IsFunN a b f,
```
aspiwack, line 143: I'm not sure it's worth repeating the documentation of `IsFunN` here.

src/Data/Replicator/Linear/Internal.hs
```diff
@@ -133,13 +133,26 @@ extend f = map f . duplicate
...
+  b
 elim f r = elim' @(NatToPeano n) f r
 
--- | @'Elim' n a b f@ asserts that @f@ is a function taking @n@ linear arguments
--- of type @a@ and then returning a value of type @b@.
+-- | @'Elim' n a b@ is used to avoid recursion in the implementation of 'elim'
+-- so that 'elim' can be inlined.
 --
--- It is solely used to define the type of the 'elim' function.
+-- 'Elim' is solely used in the signature of that function.
```
aspiwack, line 155:
```suggestion
-- 'Elim' is solely used in the signature of 'elim'.
```

src/Data/Replicator/Linear/Internal.hs
```diff
@@ -133,13 +133,26 @@ extend f = map f . duplicate
...
+  ) =>
+  f %1 ->
+  Replicator a %1 ->
+  b
 elim f r = elim' @(NatToPeano n) f r
 
--- | @'Elim' n a b f@ asserts that @f@ is a function taking @n@ linear arguments
--- of type @a@ and then returning a value of type @b@.
+-- | @'Elim' n a b@ is used to avoid recursion in the implementation of 'elim'
+-- so that 'elim' can be inlined.
```
aspiwack, line 153:
```suggestion
-- | @'Elim' n a b@ is used to implement 'elim' without recursion
-- so that we can guarantee that 'elim' will be inlined and unrolled.
```

src/Data/V/Linear.hs
```diff
@@ -50,6 +50,7 @@ module Data.V.Linear
     theLength,
     Make,
     make,
+    ArityV,
```
aspiwack, line 53:
```suggestion
    * Type-level helpers for staging
    ArityV,
```

src/Data/V/Linear/Internal.hs
```diff
@@ -98,58 +98,111 @@ uncons = Unsafe.toLinear uncons'
...
+  forall (n :: Nat) a b f.
+  ( -- Always fulfilled. GHC cannot prove it for any @n@, but can prove it
+    -- at call site when @n@ is known
+    n ~ PeanoToNat (NatToPeano n),
+    -- Always fulfilled. Provides the actual implementation of 'elim' in
+    -- terms of @elim\'@
+    Elim (NatToPeano n) a b,
+    -- Asserts that @f@ is a function taking @n@ linear arguments of type
+    -- @a@ and returning a value of type @b@
+    IsFunN a b f,
```
aspiwack, line 123: Same comments as above.

src/Data/Replicator/Linear/Internal.hs
tbagrel1: I mean: the end-user doesn't need to check something / implement something on their type to provide an instance of `Elim peanoN a b` (now that `f` is now longer part of the class parameters). There should always be an instance of `Elim peanoN a b` available for any `(peanoN, a, b)`.

My idea was to distinguish actual requirements (that is, the ones that are not fulfilled for any types, e.g. `Hash a` or `Ord a`) and constraints added just to help GHC with typechecking or to avoid recursion.

src/Data/Replicator/Linear/Internal.hs
tbagrel1: The comment was targeting the three last constraints `IsFunN a b f, f ~ FunN (NatToPeano n) a b, n ~ Arity b f`, but Ormolu made them multiline :(

At the beginning, I added these comments for the end-user, but because haddock can't accept comments on constraints, maybe it's worth to remove them all.

aspiwack: You can always write text in the docstring. Not everything needs to be structured :slightly_smiling_face: .

src/Data/Replicator/Linear/Internal.hs
aspiwack: I see. It makes sense, though it may not be the best way to say so. Basically, what you are trying to say is: “dear reader, please don't worry about these instances, they won't get in your way”. Be explicit about it.

src/Data/Replicator/Linear/Internal.hs
```diff
@@ -133,13 +133,27 @@ extend f = map f . duplicate
 --
 -- > elim (,) :: Replicator a %1 -> (a, a)
 -- > elim (,,) :: Replicator a %1 -> (a, a, a)
-elim :: forall (n :: Nat) a b f. (Elim (NatToPeano n) a b, IsFunN a b f, f ~ FunN (NatToPeano n) a b, n ~ Arity b f) => f %1 -> Replicator a %1 -> b
+--
+-- About the constraints of this function (they won't get in your way):
+--
+-- * @'Elim' ('NatToPeano' n) a b@ provides the actual implementation of 'elim'; there is an instance of this class for any @(n, a, b)@
+-- * @'IsFunN' a b f, f ~ 'FunN' ('NatToPeano' n) a b, n ~ 'Arity' b f@ indicate the shape of @f@ to the typechecker (see documentation of 'IsFunN').
```
aspiwack, line 140: :+1:

aspiwack (approved):

PR #394, opened 2022-02-23
Rebasing generic work

treeowl:
Rebase in progress of #316.


treeowl:
I don't know how to set up ormolu for formatting, and I don't really have time to futz with that right now. Can someone help me out?

treeowl:
@aspiwack aside from the formatting thing, I think this is pretty much ready. It doesn't remove *all* the orphans, but that can be done in another PR or two.

tbagrel1:
> I don't know how to set up ormolu for formatting, and I don't really have time to futz with that right now. Can someone help me out?

You should be able to format with `./format.sh` at the root of the repo. It uses `stack` to fetch, build and exec `ormolu` on the codebase.

aspiwack:
If you have Nix installed, you can `$ nix-shell --run "./format,sh"`. Alternatively, if you have Stack installed, you can `./format.sh`.

I pushed a formatting commit.

@tbagrel1 This is a pretty big patch, please give it a review, I'll do the same myself.

src/Control/Functor/Linear/Internal/Class.hs
```diff
@@ -166,3 +208,97 @@ ap f x = f >>= (\f' -> fmap f' x)
...
+deriving via Generically1 ((,,,) a b c)
+  instance Functor ((,,,) a b c)
+deriving via Generically1 ((,,,,) a b c d)
+  instance Functor ((,,,,) a b c d)
+instance Monoid a => Monad ((,) a) where
+  (a, x) >>= f = go a (f x)
+    where go :: a %1-> (a,b) %1-> (a,b)
+          go b1 (b2, y) = (b1 <> b2, y)
+
+deriving via Generically1 (Sum.Sum f g)
```
tbagrel1: I don't really like using `Sum.Sum` (because the prefix doesn't give much information about where the type comes from in this case). Maybe you can change `import Data.Monoid.Linear` to `import Data.Monoid.Linear hiding (Sum)`, and then import `Data.Functor.Sum` unqualified? Also, it would be more consistent with your import of `Data.Functor.Compose`

linear-base.cabal
```diff
@@ -122,7 +124,7 @@ library
         containers,
         ghc-prim,
         hashable,
-        linear-generics,
+        linear-generics >= 0.2 && < 0.3,
```
aspiwack, line 127: Is the upper bound necessary? Do you fear that something will go wrong with the next version of linear-generics?

src/Control/Functor/Linear/Internal/Class.hs
```diff
@@ -166,3 +208,133 @@ ap f x = f >>= (\f' -> fmap f' x)
...
+instance (Generic1 f, Functor (Rep1 f)) => Functor (Generically1 f) where
+  fmap f = Generically1 . to1 . fmap f . from1 . unGenerically1
+
+-- True if the generic type does not contain 'Par1', i.e. it does not use its parameter.
+type family NoPar1 (f :: Type -> Type) :: Bool where
+  NoPar1 U1 = 'True
+  NoPar1 (K1 i v) = 'True
+  NoPar1 (l :*: r) = NoPar1 l && NoPar1 r
+  NoPar1 (l :+: r) = NoPar1 l && NoPar1 r
+  NoPar1 (l :.: r) = NoPar1 l || NoPar1 r
```
aspiwack, line 265: Shouldn't that be
```suggestion
  NoPar1 (l :.: r) = NoPar1 l && NoPar1 r
```
?

src/Control/Functor/Linear/Internal/Class.hs
```diff
@@ -123,6 +143,28 @@ class (Data.Applicative f, Functor f) => Applicative f where
 dataPureDefault :: Applicative f => a -> f a
 dataPureDefault x = pure x
 
+instance Monoid a => Applicative ((,) a) where
```
tbagrel1, line 146: Why can't we derive it through `Generically1` here as you do in `Data.Functor.Linear.Internal.Applicative`?

src/Data/Functor/Linear/Internal/Functor.hs
```diff
@@ -17,9 +26,12 @@ import qualified Control.Monad.Trans.Reader as NonLinear
 import qualified Control.Monad.Trans.State.Strict as Strict
 import Data.Functor.Compose
 import Data.Functor.Const
-import Data.Functor.Identity
+import Data.Functor.Identity (Identity (..))
```
tbagrel1, line 29: Is there any naming conflict if we import the whole `Data.Functor.Identity` module unqualified?

src/Data/Unrestricted/Linear/Internal/Dupable.hs
```diff
@@ -1,28 +1,50 @@
+{- options_ghc -ddump-simpl -dsuppress-type-applications -dsuppress-coercions #-}
```
tbagrel1, line 1: Should we keep this there?

src/Data/Unrestricted/Linear/Internal/Dupable.hs
```diff
@@ -1,28 +1,50 @@
...
+{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# OPTIONS_HADDOCK hide #-}
 
 module Data.Unrestricted.Linear.Internal.Dupable
   ( Dupable (..),
+    genericdupR,
```
tbagrel1, line 22: I think that we should change it to `genericDupR` for consistency

src/Data/Unrestricted/Linear/Internal/Instances.hs
```diff
@@ -23,13 +23,8 @@ module Data.Unrestricted.Linear.Internal.Instances where
```
tbagrel1, line 23: Maybe it would make sense to rename this module to `OtherInstances` or `ExtraInstances` given that most of the instances were moved back into `Consumable` and `Dupable`?

src/Data/Unrestricted/Linear/Internal/Dupable.hs
```diff
@@ -1,28 +1,50 @@
...
+import GHC.Tuple (Solo (..))
+import GHC.Types (Multiplicity (..))
+import Generics.Linear
+import Prelude.Linear.Generically
 import Prelude.Linear.Internal
+import qualified Unsafe.Linear as Unsafe
+import qualified Prelude
 
 -- | The laws of 'Dupable' are dual to those of 'Monoid':
 --
```
tbagrel1, line 50: >
```
-- Implementation of 'Dupable' for 'Data.Unrestricted.Movable' types should
-- be done with @deriving via 'Data.Unrestricted.AsMovable'@.
```

We should add a mention about generic deriving in the docstring of `Dupable` IMO.

src/Data/Functor/Linear/Internal/Applicative.hs
```diff
@@ -81,10 +124,106 @@ instance (Applicative f, Semigroup a) => Semigroup (Ap f a) where
...
+  {-# INLINE gliftA2 #-}
 
-instance Monoid a => Applicative ((,) a) where
-  pure x = (mempty, x)
-  (u, f) <*> (v, x) = (u <> v, f x)
+instance
+  Unsatisfiable
+    ( 'Text "Cannot derive a data Applicative instance for" ':$$: s
+        ':$$: 'Text "because it contains one or more primitive unboxed fields."
+        ':$$: 'Text "Such unboxed types lack canonical monoid operations."
```
aspiwack, line 224: I guess that there could be a notion of monoid parametric in the type representation. But this is definitely not a battle for today.

src/Data/Functor/Linear/Internal/Functor.hs
aspiwack, line 29: I'd rather we avoid explicit imports. If there are conflicts, I prefer using `hiding`.

src/Data/Functor/Linear/Internal/Functor.hs
```diff
@@ -17,9 +26,12 @@ import qualified Control.Monad.Trans.Reader as NonLinear
...
 import Data.Functor.Sum
 import Data.Unrestricted.Linear.Internal.Consumable
+import Data.Unrestricted.Linear.Internal.Ur (Ur (..))
```
aspiwack, line 32: Same comment.

src/Data/Unrestricted/Linear/Internal/Movable.hs
```diff
@@ -27,3 +57,141 @@ import Data.Unrestricted.Linear.Internal.Ur
...
+
+instance Movable Prelude.Ordering where
+  move LT = Ur LT
+  move GT = Ur GT
+  move EQ = Ur EQ
+
+instance (Movable a, Movable b) => Movable (a, b) where
+  move (a, b) = (,) Data.<$> move a Data.<*> move b
+
+instance (Movable a, Movable b, Movable c) => Movable (a, b, c) where
```
tbagrel1, line 104: Why are we able to derive some tuple instances with `deriving via Generically` (n = 4), but not for `n = 2 or 3`?

Also, I think we should go up to n = 5 (consistency with `base`).

src/Data/Functor/Linear/Internal/Traversable.hs
```diff
@@ -24,10 +29,13 @@ where
 
 import qualified Control.Functor.Linear.Internal.Class as Control
 import qualified Control.Functor.Linear.Internal.Instances as Control
+import Control.Functor.Linear.Internal.Kan
 import qualified Control.Functor.Linear.Internal.State as Control
 import Data.Functor.Const
 import qualified Data.Functor.Linear.Internal.Applicative as Data
 import qualified Data.Functor.Linear.Internal.Functor as Data
+import GHC.Types (Multiplicity (..))
```
aspiwack, line 37: Ok, here it makes sense :slightly_smiling_face:

tbagrel1 (commented):
Just a bit of nitpicking. Thank you really much for this impressive work!

aspiwack (commented):
(I've reviewed only about half way through, but I won't have time to resume for a while, so posting what I have)

tbagrel1:
@aspiwack Do you remember the motivations behind the difference of organisation between `Data.Functor.Linear.xxx` modules and `Control.Functor.Linear.xxx` modules?

treeowl:
@aspiwack , you seem to be fighting a long term trend to prefer explicit and qualified imports to `hiding` in most cases. Why is that? The former are more robust in the face of additions to the imported modules.

aspiwack:
> @aspiwack , you seem to be fighting a long term trend to prefer explicit and qualified imports to `hiding` in most cases. Why is that? The former are more robust in the face of additions to the imported modules.

They are more robust, but in practice they cost more of my time. My inclination is to use qualified import as much as is reasonable (which will be easier when the [Local Module proposal](https://github.com/ghc-proposals/ghc-proposals/pull/283) comes to fruition). But I prefer figuring out the occasional error, than preemptively making import list management something I have to care about on a daily basis.

aspiwack:
> @aspiwack Do you remember the motivations behind the difference of organisation between `Data.Functor.Linear.xxx` modules and `Control.Functor.Linear.xxx` modules?

I think @Divesh-Otwani found it easier to document modules when they were split into smaller chunks for me. I don't care much either way as long as we have a reasonably small list of modules to import in the user-facing land.

src/Control/Functor/Linear/Internal/Class.hs
treeowl: I'm sure we can.

src/Data/Functor/Linear/Internal/Functor.hs
treeowl: No; the module doesn't (currently) contain anything else. Even if something is added, though, we're quite unlikely to need it, so I don't see the point in changing this. I'll do it if you want.

src/Data/Unrestricted/Linear/Internal/Dupable.hs
treeowl: Probably not.

src/Data/Unrestricted/Linear/Internal/Dupable.hs
treeowl: In other packages, I think the lower case convention tends to be used, but I guess we have `genericTraverse` now? I'm not sure what the right path is.

src/Data/Unrestricted/Linear/Internal/Instances.hs
treeowl: I would like to wait till another PR to deal with the rest of that module. My general preference would be to get rid of it altogether if that can be done without too much pain.

src/Data/Unrestricted/Linear/Internal/Dupable.hs
treeowl: Good idea.

src/Data/Unrestricted/Linear/Internal/Movable.hs
treeowl: Oh, we can do it all generically.

linear-base.cabal
treeowl: Just being conservative. Depending on future development of `LinearTypes` it's conceivable we could want to change the multiplicity polymorphic arrows to linear ones.

src/Control/Functor/Linear/Internal/Class.hs
treeowl: `Data.Functor.Compose` only exports one type, and its name doesn't clash with anything else around, so I don't think it's analogous. I could import `Data.Functor.Sum` without renaming though....

src/Control/Functor/Linear/Internal/Class.hs
treeowl: Oh, no we can't. These are *control* applicatives. I don't think there's a useful generic way to define them.

src/Control/Functor/Linear/Internal/Class.hs
treeowl: Nope. I don't actually remember if I wrote this or if @sjoerdvisscher did, but it's right. The composition only uses its parameter if *both* components do so.

src/Data/Functor/Linear/Internal/Applicative.hs
treeowl: That wouldn't help without some kind of overhaul of the (currently very limited) mechanism for generic representations of unlifted types.

src/Data/Functor/Linear/Internal/Functor.hs
treeowl: Fine, whatever.

treeowl:
Note: when this is merged, it would be good to give @sjoerdvisscher credit for the substantial work he did on it before I showed up. As I recall, his most interesting contribution was generic derivation of control functors. I'm not sure how best to do that in the context of git with such a wild rebase.

treeowl:
@aspiwack ormolu chokes doing something or other with dependencies (`constraints` and `hedgehog`) that have CPP in their module export lists when I use the `stack` version. The nix thing is still compiling the world.

src/Data/Functor/Linear/Internal/Functor.hs
tbagrel1: Well, I don't mind personally, but I think that this is the import style that @aspiwack likes and it's the one that is used (in principle) across the codebase.

src/Control/Functor/Linear/Internal/Class.hs
tbagrel1: Well, I don't mind personally, but I think that this is the import style that @aspiwack likes and it's the one that is used (in principle) across the codebase.

src/Data/Unrestricted/Linear/Internal/Dupable.hs
tbagrel1: We have `genericTraverse`, `genericPure`, `genericLiftA2`... so I think we should use `genericDupR`. The question is, what should we do with `gXXXX` functions (but I think this is less important because they're not publicly exposed)?

src/Data/Unrestricted/Linear/Internal/Instances.hs
tbagrel1: I agree. I also plan on doing some reorganisation of the `Control.Functor` internals in another PR

src/Data/Unrestricted/Linear/Internal/Dupable.hs
treeowl: Doesn't matter. There's no point in exposing them that I know of.

linear-base.cabal
tbagrel1: Theorically, minor versions bumps before v1.0 can contain breaking changes, so I think that the upper bound make sense. However @treeowl at the moment we specify the version and hash of `linear-generics` in `stack.yaml`, so how does that work/interfere with `linear-base.cabal` version bounds?

treeowl:
The nix formatting approach fails for me too.

linear-base.cabal
treeowl: I don't use Stack in any meaningful way, so I don't have the slightest idea.

tbagrel1:
> @aspiwack ormolu chokes doing something or other with dependencies (`constraints` and `hedgehog`) that have CPP in their module export lists when I use the `stack` version. The nix thing is still compiling the world.

I formatted it. I got no issues with `ormolu` (but I already had it installed through `format.sh` a long time ago, so maybe there is something which prevents ormolu installation via stack now?).

src/Control/Functor/Linear/Internal/Class.hs
tbagrel1: I just checked, and `Data.Functor.Sum` [only exports one type](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-Functor-Sum.html). So I think that the situation is pretty analoguous to the `Data.Functor.Compose` or `Data.Functor.Product` case

tbagrel1:
Closes #316 .

aspiwack:
> Note: when this is merged, it would be good to give @sjoerdvisscher credit for the substantial work he did on it before I showed up. As I recall, his most interesting contribution was generic derivation of control functors. I'm not sure how best to do that in the context of git with such a wild rebase.

I believe that if you add the line

```
Co-authored-by: Sjoerd Visscher <firstname.name@tweag.io>
```

To the relevant commits' messages, then Sjoerd will appear – at least in Github – as a coauthor of the commits.

src/Prelude/Linear/Unsatisfiable.hs
```diff
@@ -0,0 +1,48 @@
...
+--     => Applicative V1 where
+--   pure = unsatisfiable
+--   (<*>) = unsatisfiable
+-- @
+class (Bottom, TypeError e) => Unsatisfiable (e :: ErrorMessage)
+
+-- | Produce a value of any type (and runtime representation) under
+-- an 'Unsatisfiable' or 'Bottom' constraint.
+unsatisfiable :: forall {rep} (a :: TYPE rep). Bottom => a
+unsatisfiable = unsatisfiable' (##)
```
aspiwack, line 48: Can you explain this choice of implementation? Namely to apply `unsatisfiable'` to `(##)`. This looks a bit mysterious.

aspiwack (commented):
There are only a few points to address to get this merged. @treeowl can you do that this week? If so we can include it in the next release.

I really like this. Thanks to both of you and @sjoerdvisscher for the great work that came into this.

aspiwack:
Thanks @tbagrel1 for taking care of the small issues. @treeowl do you have time to address the couple of request for extra comments? In either case, I plan to merge this tomorrow, and then we make a release :slightly_smiling_face: .

src/Prelude/Linear/Internal.hs
```diff
@@ -70,3 +71,8 @@ forget f a = f a
 -- XXX: Temporary, until newtype record projections are linear.
 runIdentity' :: Identity a %p -> a
 runIdentity' (Identity x) = x
+
+-- | A linear version of 'Data.Coerce.coerce' for types of kind 'Data.Kind.Type'.
+lcoerce :: forall a b. Coercible a b => a %1 -> b
+lcoerce = coerce ((\x -> x) :: a %1 -> a)
```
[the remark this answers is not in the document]
treeowl: Glad you like it ;-)

src/Data/Unrestricted/Linear/Internal/Dupable.hs
```diff
@@ -41,6 +62,12 @@ import Prelude.Linear.Internal
...
 -- Implementation of 'Dupable' for 'Data.Unrestricted.Movable' types should
 -- be done with @deriving via 'Data.Unrestricted.AsMovable'@.
+--
+-- Implementation of 'Dupable' for other types can be done with
+-- @deriving via 'Generically'@. Note that at present this mechanism
+-- can have performance problems for recursive parameterized types.
+-- Specifically, the methods will not specialize to underlying 'Dupable'
+-- instances.
```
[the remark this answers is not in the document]
treeowl: @aspiwack, the underlying `Dupable` dictionary ends up being a (static) argument to a recursive function that doesn't get an unfolding. So the underlying duplication function will *never* be inlined, no matter how small it is. Probably not the biggest problem for `Dupable`; probably a bigger problem for `Traversable`, where `traverse` won't specialize to a particular `Applicative`, let alone a particular function to traverse with. See https://gitlab.haskell.org/ghc/ghc/-/issues/21131

src/Prelude/Linear/Unsatisfiable.hs
treeowl: That's the way I thought of to do it. I just realized it's much easier to understand if we instead make `unsatisfiable' :: Void` and define `unsatisfiable = case unsatisfiable' of`, so let's do that.

src/Data/Unrestricted/Linear/Internal/Dupable.hs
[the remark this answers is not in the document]
treeowl: I've linked to the GHC issue from the docs.

treeowl:
I think everything's sorted now.

treeowl:
(Unless I've broken some ormolu junk again, in which case I imagine you can fix it.)

aspiwack:
Thanks!

I'm guessing the dance around `unsatisfiable` is necessary to make it representation-polymorphic. I don't have much of a grasp of why (or why the previous implementation worked). I'd be glad if you gave a few words about it in the thread.

treeowl:
> Thanks!
> 
> I'm guessing the dance around `unsatisfiable` is necessary to make it representation-polymorphic. I don't have much of a grasp of why (or why the previous implementation worked). I'd be glad if you gave a few words about it in the thread.

Yes, *partly* to make it representation polymorphic to match what the GHC built-in version will do as closely as we can. The previous implementation had

```haskell
unsatisfiable' :: Bottom => a
```

This was used at type

```haskell
unsatisfiable' :: Bottom  => (##) -> q
```

where `q` is the actual desired result type.

----

What's the rest of the dance for? The "obvious" thing is to give `unsatisfiable` an `Unsatisfiable e` constraint. But this would not be ergonomical at all, and would not match the "magical" behavior of the future built-in version. Why? Because each invocation of `unsatisfiable` would then have to specify the type error, completely uselessly. By using `Bottom`, we avoid caring, at the call site, *why* we couldn't possibly have gotten there.

aspiwack (approved):

treeowl:
Yes, I'll try to do it tonight. Lots of things going on!

PR #393, opened 2022-02-22
Prepare Changelog for v0.2.0

tbagrel1:
PRs submitted before 2022/02/25 16:00:00 have been taken into account at this point.

The file `CHANGELOG-ENTRIES--delete-before-merge.md` is the output I got from the [github-changelog-generator](https://github.com/github-changelog-generator/github-changelog-generator) tool. I then looked at each PR, improved its title/description, and added it to the right group in `CHANGELOG.md`. 

Closes #377 .

PR #397, opened 2022-02-25
Remove `Prelude.Linear.asTypeOf`

### LINKED ISSUE #354: Prelude asTypeOf isn't very useful
The linear prelude exports

```haskell
asTypeOf :: a %q-> a -> a
asTypeOf = const
```

That seems practically useless in a linear context. The traditional alternative is found in `Data.Proxy`, and can be given the following type:

```haskell
asProxyTypeOf :: a %q-> proxy a -> a
asProxyTypeOf a _ = a
```

Of course, if one is using `TypeApplications`, `id` is also an option. But I'd consider just removing `asTypeOf` here.

tbagrel1:
Closes #354 . The linked issue gives the motivations behind this removal.

### DIFF
--- a/src/Prelude/Linear.hs
+++ b/src/Prelude/Linear.hs
@@ -82,7 +82,6 @@ module Prelude.Linear
     ($),
     (&),
     Prelude.until,
-    asTypeOf,
     Prelude.error,
     Prelude.errorWithoutStackTrace,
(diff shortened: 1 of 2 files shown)


tbagrel1:
Reminder : #393 #377

PR #398, opened 2022-03-14
Move CI tests from cabal job to stack job

tbagrel1:
Cabal do not pick the same Hedgehog version as stack, so we get deprecation warnings (treated as errors) in the cabal CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,11 +28,9 @@ jobs:
     - name: Update Cabal's database
       run: nix-shell --pure --run "cabal update"
     - name: Build Cabal's dependencies
-      run: nix-shell --pure --run "cabal build --enable-tests --enable-benchmarks --dependencies-only"
+      run: nix-shell --pure --run "cabal build --disable-tests --disable-benchmarks --dependencies-only"
     - name: Build
-      run: nix-shell --pure --run "cabal build"
-    - name: Test
-      run: nix-shell --pure --run "cabal test"
+      run: nix-shell --pure --run "cabal build --disable-tests --disable-benchmarks"
     - name: Haddock
       run: nix-shell --pure --run "cabal haddock"
     - name: cabal-docspec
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -74,4 +72,4 @@ jobs:
     - name: Build Nix dependencies
       run: nix-shell --pure --run "echo '=== Nix dependencies installed ==='"
     - name: Build
-      run: nix-shell --pure --run "stack build --pedantic --test --bench --no-run-tests --no-run-benchmarks"
+      run: nix-shell --pure --run "stack build --pedantic --test --bench"
```
aspiwack, line 75: We should probably not run benchmarks here. I'm not sure what purpose they would serve.
```suggestion
      run: nix-shell --pure --run "stack build --pedantic --test --bench --no-run-benchmarks"
```

.github/workflows/ci.yaml
```diff
@@ -28,11 +28,9 @@ jobs:
     - name: Update Cabal's database
       run: nix-shell --pure --run "cabal update"
     - name: Build Cabal's dependencies
-      run: nix-shell --pure --run "cabal build --enable-tests --enable-benchmarks --dependencies-only"
+      run: nix-shell --pure --run "cabal build --disable-tests --disable-benchmarks --dependencies-only"
     - name: Build
-      run: nix-shell --pure --run "cabal build"
-    - name: Test
-      run: nix-shell --pure --run "cabal test"
+      run: nix-shell --pure --run "cabal build --disable-tests --disable-benchmarks"
```
aspiwack, line 33: You don't have to disable tests/benchmarks again (in fact, I believe it's the default anyway, that they are disabled).
```suggestion
      run: nix-shell --pure --run "cabal build"
```

tbagrel1: Well, I think that they are enabled by default: see this CI job that failed on friday when I didn't explicitely disabled them: https://github.com/tweag/linear-base/actions/runs/1969067243 and https://github.com/tweag/linear-base/commit/e64d220f764f7074e9e0f20cbafdeb8ff0d1a0a8

aspiwack: That's because they've been enabled above. the `--enable-tests` mutates a state. It doesn't tell `cabal` to run tests. Try with the suggestions applied, and we will have a definite answer to that question though.

tbagrel1: Cabal CLI is stateful? :scream:

aspiwack:
Well, I stand corrected.

PR #399, opened 2022-03-17
Bump `linear-base` version in `linear-base.cabal`

tbagrel1:
From 0.1.1 to 0.2.0


treeowl:
Ugh, I thought I already did this. Definitely necessary.

Issue #404, opened 2022-04-10
Add `evalState` (to `Control.Functor.Linear`)


andreasabel:
`evalState` is a common use-case of the state monad.
```haskell
import Control.Functor.Linear
import Data.Tuple.Linear

evalState :: Consumable s => State s a %1 -> s %1 -> a
evalState m s = fst (runState m s)
```
Probably there should also be `evalStateT`.

aspiwack:
I'm not sure that `evalState` is quite as useful as in the case of the unrestricted-state monad. That being said, I'll welcome a PR, but see my thoughts in https://github.com/tweag/linear-base/issues/405#issuecomment-1094625847.

andreasabel:
Maybe I can do without it, but I didn't see how because of the `Ur` in the type of `HashMap.empty`.  I'd be happy for advice, see my code at: https://gist.github.com/andreasabel/f5c877491d32b653efba44cb283ba490#file-interpretermain-hs

andreasabel:
> So if this is an admissible generalization, ...

I ain't...

andreasabel:
@aspiwack : I am back to `evalState`, see
- #411 

If you have other ideas how to write https://gist.github.com/andreasabel/f5c877491d32b653efba44cb283ba490#file-interpretermain-hs , please let me know.

PR #410, opened 2022-04-12
More polymorphic type for `Array.alloc` and `HashMap.empty`

andreasabel:
Replacing response type `Ur b` by simply `b` also type checks and enables uses like
```haskell
let (i, _) = HashMap.empty 100 (\ env -> runState (eval e) env)
```
(see issue #404). Without this PR, the respective code would be
```haskell
let Ur i = HashMap.empty 100 (\ env -> move (fst (runState (eval e) env)))
```
with `fst :: Consumable b => (a, b) -> a`.  
The old typing of `empty` forces the insertion of `move` which can only happen after consuming the `env` via `fst` (because `HashMap` does not implement `Consumable`).  
The new code seem to be better, allowing `env` to be garbage-collected unforced once we want to dispose of it.

Caveat: I am relying on the linear type checker (which might be in alpha state, dunno), so there might be reasons unknown to me why this generalization is invalid.  Test suite passes, though.

Also in this PR: Restrict `tasty-hedgehog < 1.2` to make `cabal build` succeed.

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -183,7 +183,7 @@ test-suite examples
         base,
         linear-base,
         tasty,
-        tasty-hedgehog,
+        tasty-hedgehog < 1.2,
         hedgehog,
         storable-tuple,
         vector,
--- a/src/Data/Array/Mutable/Linear/Internal.hs
+++ b/src/Data/Array/Mutable/Linear/Internal.hs
@@ -65,8 +65,8 @@ alloc ::
   HasCallStack =>
   Int ->
   a ->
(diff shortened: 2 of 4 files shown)


andreasabel:
Ok, so the type checker isn't sound yet.  
Was unlikely anyway that the simplification I suggested would have been overlooked...

aspiwack:
The typechecker is sound. But the implementation of `Array` uses `unsafePerformIO`, which isn't. The careful choice of API is what makes the use of `unsafePerformIO` ok. Just like any abstract type really.

Issue #389, opened 2022-02-17
Move to GHC 9.2


aspiwack:
The upcoming version 0.2 will presumably be the last version where we will support GHC 9.0.

Moving to GHC 9.2 will let us benefit from the improved multiplicity inference and Core generation that we shipped there.

**This issue is blocked on** haskell/haskell-language-server#2179 (partially so: I think that hls will be working well enough with GHC 9.2 before this issue is closed)
**This issue is blocked on** Stackage publishing a snapshot for GHC 9.2

- [ ] Change the default compiler in `shell.nix` to GHC 9.2
- [ ] Change the stack snapshot to a GHC 9.2 one
- [ ] Make sure the CI runs on GHC 9.2
- [ ] Remove the `scrutinee & \case` trick, both from the documentation and from the code

PR #414, opened 2022-05-03
Upgrade GHC to 9.2

aspiwack:
Part of #389.

---

This PR originally fixed the tasty-hedgehog deprecation warning. But because it turned out to that GHC 9.2 is having problem with our PR, I extracted this bit in #415.

I'll rebase on master as soon as #415 is merged to clear up the tasty-hedgehog bits.

---

Edit: I wanted to only make a draft PR, but I misclicked. I don't think that
I can convert it back. I don't know how anyway. I don't want to merge quite
yet, as there is a bug related to `stack ghci` which causes an anomaly in GHC:

```
$ stack ghci
Stack has not been tested with GHC versions above 9.0, and using 9.2.2, this may fail
Stack has not been tested with Cabal versions above 3.4, but version 3.6.3.0 was found, this may fail
Configuring GHCi with the following packages: linear-base
GHCi, version 9.2.2: https://www.haskell.org/ghc/  :? for help
[ 1 of 86] Compiling Data.Arity.Linear.Internal ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Arity/Linear/Internal.hs, interpreted )
[ 2 of 86] Compiling Data.Arity.Linear ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Arity/Linear.hs, interpreted )
[ 3 of 86] Compiling Data.Bool.Linear ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Bool/Linear.hs, interpreted )
[ 4 of 86] Compiling Prelude.Linear.GenericUtil ( /home/aspiwack/projects/tweag/linear-base/master/src/Prelude/Linear/GenericUtil.hs, interpreted )
[ 5 of 86] Compiling Data.Unrestricted.Linear.Internal.Ur ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Unrestricted/Linear/Internal/Ur.hs, interpreted )
[ 6 of 86] Compiling Prelude.Linear.Generically ( /home/aspiwack/projects/tweag/linear-base/master/src/Prelude/Linear/Generically.hs, interpreted )
[ 7 of 86] Compiling Prelude.Linear.Internal ( /home/aspiwack/projects/tweag/linear-base/master/src/Prelude/Linear/Internal.hs, interpreted )
[ 8 of 86] Compiling Data.Replicator.Linear.Internal.ReplicationStream ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Replicator/Linear/Internal/ReplicationStream.hs, interpreted )
[ 9 of 86] Compiling Data.Replicator.Linear.Internal ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Replicator/Linear/Internal.hs, interpreted )
[10 of 86] Compiling Prelude.Linear.Unsatisfiable ( /home/aspiwack/projects/tweag/linear-base/master/src/Prelude/Linear/Unsatisfiable.hs, interpreted )
[11 of 86] Compiling Unsafe.Linear    ( /home/aspiwack/projects/tweag/linear-base/master/src/Unsafe/Linear.hs, interpreted )
[12 of 86] Compiling Data.Unrestricted.Linear.Internal.Consumable ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Unrestricted/Linear/Internal/Consumable.hs, interpreted )
[13 of 86] Compiling Data.Unrestricted.Linear.Internal.Dupable ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Unrestricted/Linear/Internal/Dupable.hs, interpreted )
[14 of 86] Compiling Data.V.Linear.Internal ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/V/Linear/Internal.hs, interpreted )
[15 of 86] Compiling Data.Tuple.Linear ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Tuple/Linear.hs, interpreted )
[16 of 86] Compiling Data.Monoid.Linear.Internal.Semigroup ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Monoid/Linear/Internal/Semigroup.hs, interpreted )
[17 of 86] Compiling Data.Monoid.Linear.Internal.Monoid ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Monoid/Linear/Internal/Monoid.hs, interpreted )
[18 of 86] Compiling Data.Monoid.Linear ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Monoid/Linear.hs, interpreted )
[19 of 86] Compiling Data.Functor.Linear.Internal.Functor ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Functor/Linear/Internal/Functor.hs, interpreted )
[20 of 86] Compiling Data.Functor.Linear.Internal.Applicative ( /home/aspiwack/projects/tweag/linear-base/master/src/Data/Functor/Linear/Internal/Applicative.hs, interpreted )

/home/aspiwack/projects/tweag/linear-base/master/src/Data/Functor/Linear/Internal/Applicative.hs:101:18: error:ghc: panic! (the 'impossible' happened)
  (GHC version 9.2.2:
	No skolem info:
  [k_ad3j]
  Call stack:
      CallStack (from HasCallStack):
        callStackDoc, called at compiler/GHC/Utils/Panic.hs:181:37 in ghc:GHC.Utils.Panic
        pprPanic, called at compiler/GHC/Tc/Errors.hs:2912:17 in ghc:GHC.Tc.Errors

Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug


<no location info>: error:
    Could not load module ‘Control.Functor.Linear’
    It is a member of the hidden package ‘linear-base-0.2.0’.
    You can run ‘:set -package linear-base’ to expose it.
    (Note: this unloads all the modules in the current scope.)
Loaded GHCi configuration from /tmp/haskell-stack-ghci/eea5bf06/ghci-script
```

I haven't investigated this problem yet, and I want to know a bit more before we merge

---

Another problem that I noticed is that the format script cannot install Ormolu anymore (due to an upper bound somewher)

- [ ] Investigate the anomaly
- [x] Fix Ormolu installation in the format script

### DIFF
--- a/bench/Data/Mutable/HashMap.hs
+++ b/bench/Data/Mutable/HashMap.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE LinearTypes #-}
@@ -39,7 +40,7 @@ deriving instance Ord Key
 
 deriving instance Generic Key
 
-deriving instance NFData Key
+deriving anyclass instance NFData Key
 
 instance Hashable Key where
   hash (Key x) =
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -17,10 +17,10 @@
         "homepage": "https://github.com/NixOS/nixpkgs",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "b2737d4980a17cc2b7d600d7d0b32fd7333aca88",
-        "sha256": "012db5d6k0lajp4q37byhgamz3ry04av1dcpgf3ahm9kzjwsjcch",
+        "rev": "7beebb590d541ffa534aa34b0b163f81c3c72c2c",
+        "sha256": "0lb134206mdrsxa3id61bn8p1ylwagv1k19jx1a0x9ywkfp0mvp1",
         "type": "tarball",
-        "url": "https://github.com/NixOS/nixpkgs/archive/b2737d4980a17cc2b7d600d7d0b32fd7333aca88.tar.gz",
+        "url": "https://github.com/NixOS/nixpkgs/archive/7beebb590d541ffa534aa34b0b163f81c3c72c2c.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     }
 }
--- a/shell.nix
+++ b/shell.nix
@@ -1,4 +1,4 @@
-{ system ? builtins.currentSystem, sources ? import ./nix/sources.nix, ghcVersion ? "902" }:
+{ system ? builtins.currentSystem, sources ? import ./nix/sources.nix, ghcVersion ? "922" }:
 
 let
   selectHls = self: super: {
@@ -19,9 +19,6 @@ let
         "
     '';
   };
-  brokenIn92 = if ghcVersion == "921" then [] else [
-    pkgs.haskell-language-server
-  ];
 in with pkgs;
 
 mkShell {
@@ -35,5 +32,6 @@ mkShell {
     stack-wrapped
     nix
     cabal-docspec
-  ] ++ brokenIn92;
+    pkgs.haskell-language-server
+  ];
 }
--- a/src/Control/Functor/Linear/Internal/Class.hs
+++ b/src/Control/Functor/Linear/Internal/Class.hs
@@ -241,12 +241,12 @@ instance Monoid a => Monad ((,) a) where
       go b1 (b2, y) = (b1 <> b2, y)
 
 deriving via
-  Generically1 (Sum f g)
+  Generically1 (Sum (f :: Type -> Type) g)
   instance
     (Functor f, Functor g) => Functor (Sum f g)
 
 deriving via
-  Generically1 (Compose f g)
+  Generically1 (Compose (f :: Type -> Type) (g :: Type -> Type))
   instance
     (Functor f, Functor g) => Functor (Compose f g)
 
--- a/src/Data/Functor/Linear/Internal/Applicative.hs
+++ b/src/Data/Functor/Linear/Internal/Applicative.hs
@@ -32,6 +32,7 @@ import Data.Monoid (Ap (..))
 import Data.Monoid.Linear hiding (Product)
 import Data.Unrestricted.Linear.Internal.Ur (Ur (..))
 import GHC.TypeLits
+import GHC.Types
 import Generics.Linear
 import Prelude.Linear.Generically
 import Prelude.Linear.Internal
@@ -96,12 +97,12 @@ deriving via
     Monoid a => Applicative ((,) a)
 
 deriving via
-  Generically1 (Product f g)
+  Generically1 (Product (f :: Type -> Type) g)
   instance
     (Applicative f, Applicative g) => Applicative (Product f g)
 
 deriving via
-  Generically1 (f :*: g)
+  Generically1 ((f :: Type -> Type) :*: g)
   instance
     (Applicative f, Applicative g) => Applicative (f :*: g)
 
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,9 +1,3 @@
-resolver: nightly-2022-01-13
+resolver: nightly-2022-04-28
 packages:
 - '.'
-extra-deps:
-- tasty-hedgehog-1.2.0.0
-- ormolu-0.4.0.0
-- Cabal-3.6.2.0@sha256:ae204c95edd633538c3e502eee8838dc41d8b13c8f333955a39ff9df7b6de9e8,12852
-- ghc-lib-parser-9.2.1.20220109@sha256:5da80d84452a31b7d0dab46d849a9c3552897405048d466b766136d852d20cfe,12705
-- linear-generics-0.2@sha256:c1db1fcb96333be867978abfbed71e99dfbdcafa07d7d9642a89405e6bc971b1,5818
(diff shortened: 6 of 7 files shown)


aspiwack:
Solving the Ormolu problem took a negative number of lines. Convenient.

aspiwack:
The panic happens on the library. It's probably not caused by the clever setup that `stack ghci` does to load all your modules across units.

It doesn't happen with `stack build` though. This is quite puzzling for something that appear to be a type-checking error.

aspiwack:
The panic happens in the `Data.Functor.Linear.Internal.Applicative` module. On standalone instances derived via `Generically`. Not all of them:

This one breaks (the error appears to happen on the last line) ❌
```haskell
deriving via
  Generically1 (Product f g)
  instance
    (Applicative f, Applicative g) => Applicative (Product f g)
```

This one works :heavy_check_mark: 
```haskell
deriving via
  Generically1 ((,) a)
  instance
    Monoid a => Applicative ((,) a)
```

Why there is a different between the two does elude me some. And why only in ghci? This is a weird puzzle.

treeowl:
Could it be a bug in GHC's implementation of Template Haskell? Have you reported a GHC bug?

aspiwack:
I haven't reported a GHC bug yet (there doesn't seem to be any Template Haskell involved anyway). As I don't know what to report :slightly_smiling_face: .

I'll try to minimize later this week, now I've got to go. I've had a look at GHC's source to see where could the source of this message be, and it's not obvious.

aspiwack:
@monoidal points me at the fact that @andreasabel has already mentioned this very bug in the GHC bug tracker https://gitlab.haskell.org/ghc/ghc/-/merge_requests/7105#note_421618 .

examples/Test/Foreign.hs
```diff
@@ -36,16 +37,16 @@ listExampleTests :: TestTree
 listExampleTests =
   testGroup
     "list tests"
-    [ testProperty "List.toList . List.fromList = id" invertNonGCList,
-      testProperty "map id = id" mapIdNonGCList,
-      testProperty "memory freed post-exception" testExecptionOnMem
+    [ testPropertyNamed "List.toList . List.fromList = id" "invertNonGCList" invertNonGCList,
+      testPropertyNamed "map id = id" "mapIdNonGCList" mapIdNonGCList,
+      testPropertyNamed "memory freed post-exception" "testExecptionOnMem" testExecptionOnMem
```
andreasabel, line 42: Maybe you want to fix the typo `testExecptionOnMem` to `testExceptionOnMem`while you are at it.

andreasabel (approved):

examples/Test/Foreign.hs
aspiwack: You know, it was all search and replace :slightly_smiling_face: .

Thanks for noticing!

aspiwack:
This PR has now been changed to only care about GHC 9.2. The tasty-hedgehog bits are now handled by #415 .

aspiwack:
Rebased. I still don't want to merge though. Let me (or someone) figure out what is happening with ghci.

aspiwack:
Here is a minimised version of the bug (on GHC 9.2.2).

Uncomment any of the two commented signatures and the module seems to compile (I have yet to figure out why… it should fail to derive the instance I'm a little puzzled). At any rate, it makes the anomaly disappear.

```haskell
{-# LANGUAGE DerivingVia #-}
{-# LANGUAGE DeriveAnyClass #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE StandaloneDeriving #-}

module Data.Functor.Linear.Internal.Applicative
  ( Applicative (..),
  )
where

import GHC.Types

-- type Product :: (Type -> Type) -> (Type -> Type) -> Type -> Type
data Product f g a = Pair (f a) (g a)

-- type Generically1 :: (Type -> Type) -> Type -> Type
newtype Generically1 f a = Generically1 (f a)

type Functor :: (Type -> Type) -> Constraint
class Functor f where
class Functor f => Applicative f where

deriving instance (Functor f, Functor g) => Functor (Product f g)

deriving via
  Generically1 (Product f g)
  instance
    (Applicative f, Applicative g) => Applicative (Product f g)
```

aspiwack:
The bug appears to be fixed on (a recent) master:

```
$ ghci src/Data/Functor/Linear/Internal/Applicative.hs 
GHCi, version 9.3.20220406: https://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Data.Functor.Linear.Internal.Applicative ( src/Data/Functor/Linear/Internal/Applicative.hs, interpreted )

src/Data/Functor/Linear/Internal/Applicative.hs:29:18: error:
    • Couldn't match kind ‘k’ with ‘*’
      Expected kind ‘* -> *’, but ‘f’ has kind ‘k -> *’
      ‘k’ is a rigid type variable bound by
        the deriving clause for ‘Generically1 (Product f g)’
        at src/Data/Functor/Linear/Internal/Applicative.hs:27:3-28
    • In the first argument of ‘Applicative’, namely ‘f’
      In the stand-alone deriving instance for
        ‘(Applicative f, Applicative g) => Applicative (Product f g)’
   |
29 |     (Applicative f, Applicative g) => Applicative (Product f g)
   |                  ^
```

Maybe we can use this error message to build a workaround.

aspiwack:
I am a little bit confused as to why,  locally, stack tells me that `GeneralizedNewtypeDeriving` is available, and on the CI it fails. But not with Cabal. No matter, I have an easy fix.

aspiwack:
I wrote an issue upstream [ghc!21514](https://gitlab.haskell.org/ghc/ghc/-/issues/21514). I'll give myself a little time to think of a better workaround. Then we will be able to merge.

aspiwack:
I haven't found a better workaround. So we'll go with this, even though it's rather inelegant and fragile (it's probably a good idea to use GHC2021 as the default language in Cabal to avoid the fragility, but I think I had good reasons not to do that at some point, so I won't do this immediately).

PR #415, opened 2022-05-05
Don't use deprecated `testProperty` from tasty-hedgehog

aspiwack:
The `testProperty` function has been deprecated in the current 1.2
version. See https://github.com/qfpl/tasty-hedgehog/pull/42 for the
reason.

The `testProperty` function is replaced by `testPropertyNamed` which
requires one extra argument (the name of the test function as a
string).

There were a test for which the property didn't have a name, so I had
to name it. We also need `-XOverloadedString` in every module (some
already have it).

---

This is extracted from #414 , which was supposed to be a 5min fix and turned
out to spiral out of control. I realised it was possible to turn it back into a 5min
fix: add the more recent version of tasty-hedgehog to the `extra-deps`.

### DIFF
--- a/examples/Generic/Traverse.hs
+++ b/examples/Generic/Traverse.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE LinearTypes #-}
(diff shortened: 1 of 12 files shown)


andreasabel (approved):

tbagrel1 (approved):
Looks good to me!

PR #416, opened 2022-05-20
Add some instances.

sellout:
This does affect the API, as `V.<*>` now has parametric multiplicity.


aspiwack (approved):
This is all good.

It does make a choice: that `V n a` behaves like `(a, …, a)`, and can't be implemented with a mutable data structure. I had shied away from making this choice definite, so far. But honestly, I'm fine with it. It was already probably impossible to make `V n a` backed by a mutable data structure anyway, see #312.

PR #419, opened 2022-05-21
bitwise-and BitwiseAnd for Int and Word

yaitskov:


aspiwack:
Hello,

Thanks for your PR. Can you elaborate on your design? Why create a `BitwiseAnd` class, in particular?

yaitskov:
> Thanks for your PR. Can you elaborate on your design? Why create a `BitwiseAnd` class, in particular?

I was trying to mimic linear-base approach on tackling Num class i.e. a class per method.
Data.Bits class is pretty populated with method too. 
I haven't defined Bits in PR just because not sure about getting review responses so saved on initial efforts ;).

I don't have strong opinion about particular class name and open to alternatives.

PR #420, opened 2022-06-07
Fix haddock hyperlink reference (minor typo)

undergroundquizscene:
There’s a missing quote here, which is preventing the reference from being hyperlinked (e.g. on Hackage).


aspiwack (approved):

aspiwack:
Thanks for catching this.

PR #423, opened 2022-07-04
Fix haddock links in Data.Replicator.Linear

undergroundquizscene:
The module that `Movable` and `Dupable` are in (at least now) is
`Data.Unrestricted.Linear`, not `Data.Unrestricted`.


aspiwack:
Thanks!

aspiwack (approved):

PR #425, opened 2022-08-13
Add some Handle operations to RIO

### LINKED ISSUE #424: How to bring other `Handle`-using functions into the RIO world?
**TL;DR:** `linear-base` seems to not provide me with a way to turn functions of type `Handle -> IO a` into `Handle %1 -> RIO (Ur? a, Handle)`, and this seems to stop me from doing anything with `ByteString`.

I am trying to build a library to parse (and later, write) a data archive format from an old game, and do it with linear types to stretch myself. The file begins with a header listing the contents of all files in the archive, and their byte offsets into the archive itself. I want to write a function to return a `ByteString` stream from a `Handle`, when given the `Record` describing the name/offset/etc.

If I was using standard Haskell, I would write something like this, and trust the user not to abuse the `Handle` while streaming (we must `hSeek` to the offset and then stream the required number of bytes:

```haskell
hGetRecord :: MonadIO m => Record -> Handle -> ByteStream m ()
```

But in `linear-base`, we seem to hit the following:
1. We can't use the standard `streaming` types, we have to use the one in `Streaming.Linear`.
2. This means we can't use the `streaming-bytestring` `ByteStream` type, and must use `Stream (Of ByteString)` from `linear-base`. Annoying, but OK.

The function I think I want to write with linear types:

```haskell
hGetRecord :: Record -> Handle %1 -> Stream (Of ByteString) RIO Handle
```

This should let me safely seek the handle, without anyone interfering with it until the streaming is finished. But I cannot lift the `ByteString` I/O functions from `System.IO.IO` to `RIO` because the constructor for `System.IO.Linear.Resource.Handle` is not exported.

I think that variants of `unsafeFromSystemIOResource` etc that allow me to bring `Handle`-using `IO` functions into `RIO` and have them play nicely with the existing `Handle` type in `linear-base` would give me everything I need, but that's probably an unsafe level of power. The functions I need are `RIO` versions of:

* `openBinaryFile`
* `hSeek`
* `Data.ByteString.Lazy.hGet`

endgame:
Add `System.IO.Resource.Linear.{openBinaryFile,hSeek,SeekMode(..),hTell}`.

Haddocks are in anticipation of a near 0.2.1 release; happy to change/remove the `@since` annotations if they're likely to be wrong.

Related: #424.

### DIFF
--- a/src/System/IO/Resource/Linear/Internal.hs
+++ b/src/System/IO/Resource/Linear/Internal.hs
@@ -158,6 +158,16 @@ hPutStrLn h s = flipHPutStrLn s h -- needs a multiplicity polymorphic flip
     flipHPutStrLn s =
       coerce (unsafeFromSystemIOResource_ (\h' -> Text.hPutStrLn h' s))
 
+-- | See @System.IO.'System.IO.hSeek'@.
+--
+-- @since 0.2.1
+hSeek :: Handle %1 -> System.SeekMode -> Integer -> RIO Handle
+hSeek h mode i = coerce hSeek' h
+  where
+    hSeek' :: Handle %1 -> RIO Handle
+    hSeek' =
+      coerce (unsafeFromSystemIOResource_ (\h' -> System.hSeek h' mode i))
+
 -- new-resources
 
 -- | The type of system resources.  To create and use resources, you need to
--- a/src/System/IO/Resource/Linear/Internal.hs
+++ b/src/System/IO/Resource/Linear/Internal.hs
@@ -114,6 +114,17 @@ openFile path mode = Control.do
       (\h -> Linear.fromSystemIO $ System.hClose h)
   Control.return $ Handle h
 
+-- | See @System.IO.'System.IO.openBinaryFile'@
+--
+-- @since 0.2.1
+openBinaryFile :: FilePath -> System.IOMode -> RIO Handle
+openBinaryFile path mode = Control.do
+  h <-
+    unsafeAcquire
+      (Linear.fromSystemIOU $ System.openFile path mode)
+      (\h -> Linear.fromSystemIO $ System.hClose h)
+  Control.pure $ Handle h
+
 hClose :: Handle %1 -> RIO ()
 hClose (Handle h) = unsafeRelease h
 
--- a/src/System/IO/Resource/Linear/Internal.hs
+++ b/src/System/IO/Resource/Linear/Internal.hs
@@ -168,6 +168,12 @@ hSeek h mode i = coerce hSeek' h
     hSeek' =
       coerce (unsafeFromSystemIOResource_ (\h' -> System.hSeek h' mode i))
 
+-- | See @System.IO.'System.IO.hTell'@.
+--
+-- @since 0.2.1
+hTell :: Handle %1 -> RIO (Ur Integer, Handle)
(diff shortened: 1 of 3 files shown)


src/System/IO/Resource/Linear/Internal.hs
```diff
@@ -147,6 +158,22 @@ hPutStrLn h s = flipHPutStrLn s h -- needs a multiplicity polymorphic flip
...
+  where
+    hSeek' :: Handle %1 -> RIO Handle
+    hSeek' =
+      coerce (unsafeFromSystemIOResource_ (\h' -> System.hSeek h' mode i))
+
+-- | See @System.IO.'System.IO.hTell'@.
+--
+-- @since 0.2.1
+hTell :: Handle %1 -> RIO (Ur Integer, Handle)
+hTell = coerce (unsafeFromSystemIOResource System.hTell)
```
endgame, line 175: It is not clear to me why this needs to consume the handle; maybe `Handle -> RIO (Ur Integer)` is fine since we're only querying? But I cribbed the type from `hIsEOF` where I'd expect similar reasoning to apply.

aspiwack (commented):
A few comments to address below, and it'll be good to go.

src/System/IO/Resource/Linear/Internal.hs
```diff
@@ -105,7 +105,7 @@ instance Control.Monad RIO where
 
 newtype Handle = Handle (UnsafeResource System.Handle)
 
--- | See 'System.IO.openFile'
+-- | See @System.IO.'System.IO.openFile'@
```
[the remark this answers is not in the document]
endgame: Yes:
htell

src/System/IO/Resource/Linear/Internal.hs
endgame: You say "keep in mind" as if that was a thing that had ever entered my mind :-)

I don't understand you when you say "as we have a linear `Handle` in scope, we can't ask for an unrestricted copy of it." I thought linearity in this extension was a property of function arrows, not of values?

What does make sense to me is that I just realised `RIO` only has a linear control monad instance, which means we couldn't compose with a function of type `Handle -> RIO (Ur Integer)`, which means we have to use a linear arrow, which means we have to be careful to return the `Handle`. What other parts of the picture am I missing?

endgame (commented):
I applied all your suggestions.

src/System/IO/Resource/Linear/Internal.hs
aspiwack: > You say "keep in mind" as if that was a thing that had ever entered my mind :-)

:slightly_smiling_face: 

> I thought linearity in this extension was a property of function arrows, not of values?

The terminology is a little ambiguous here. A “linear function” is a function with “linear argument” (it's the argument, the formal parameter, that is linear not its value).

> What does make sense to me is that I just realised `RIO` only has a linear control monad instance, which means we couldn't compose with a function of type `Handle -> RIO (Ur Integer)`, which means we have to use a linear arrow, which means we have to be careful to return the `Handle`. What other parts of the picture am I missing?

Yes, this is exactly what's happening.

aspiwack (approved):
I edited the title of the PR for clarity. I'll be merging now. Thanks for your PR.

PR #427, opened 2022-08-26
Enable builds on both GHC 9.2 and GHC 9.4

matthew-healy:
This PR allows linear-base to be built with both GHC 9.2 and 9.4. There were two main changes which had to be accounted for: firstly that the `~`type equality operator is now a "regular" type operator, exported from `Data.Type.Equality`, and secondly that `GHC.Generic` now exports `Generically` and `Generically1`.

Both of these issues were fixed in a similar way: we defined source directories `ghc92` and `ghc94` which are included in the build based on the current GHC version. This allows us to, for example, define `Generically` in GHC 9.2 builds but re-export it when building with 9.4.

One caveat with this approach is that we also had to move the entry point to `Generics.Linear` into one of these shim modules, as it re-exports `GHC.Generic`, which meant that we were getting unused import warnings in places that imported both `Generics.Linear` and `Prelude.Linear.Internal`. This seems like a reasonable cost to pay, especially since it can be removed when support for GHC 9.2 is dropped.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,6 +36,36 @@ jobs:
     - name: cabal-docspec
       run: nix-shell --pure --run cabal-docspec
 
+  cabal-test-941:
+    name: cabal test with GHC 9.4.1
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: cachix/install-nix-action@v15
+      with:
+        nix_path: "${{ env.nixpkgs-url }}"
+    - name: Cache Cabal dependencies
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cabal/packages
+          ~/.cabal/store
+          dist-newstyle
+        key: cabal-deps-${{ runner.os }}-${{ hashFiles('nix/sources.json') }}-v${{ env.cache-invalidation-key }}-${{ hashFiles('linear-base.cabal') }}-${{ github.sha }}
+        restore-keys: cabal-deps-${{ runner.os }}-${{ hashFiles('nix/sources.json') }}-v${{ env.cache-invalidation-key }}-${{ hashFiles('linear-base.cabal') }}-
+    - name: Build Nix dependencies
+      run: nix-shell --arg ghcVersion '"941"' --pure --run "echo '=== Nix dependencies installed ==='"
+    - name: Update Cabal's database
+      run: nix-shell --arg ghcVersion '"941"' --pure --run "cabal update"
+    - name: Build Cabal's dependencies
+      run: nix-shell --arg ghcVersion '"941"' --pure --run "cabal build --allow-newer --disable-tests --disable-benchmarks --dependencies-only"
+    - name: Build
+      run: nix-shell --arg ghcVersion '"941"' --pure --run "cabal build --allow-newer --disable-tests --disable-benchmarks"
+    - name: Haddock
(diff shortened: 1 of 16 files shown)


aspiwack:
As a note: let's make sure that we enable GHC 9.4 in the nix-shell for the cabal-based CI. To test both versions.

aspiwack:
> warning This PR is based off of the branch `compat-9.4-update-nix-shell` which makes it possible to try to compile linear-base with GHC 9.4. That PR should ideally be merged first.

It's not tremendously important, but just so you kno, we have dpulls installed in the org, so you can use this to mark dependencies instead of managing them manually.

jeicher:
Depends on #429

.github/workflows/ci.yaml
```diff
@@ -36,6 +36,36 @@ jobs:
     - name: cabal-docspec
       run: nix-shell --pure --run cabal-docspec
 
+  cabal-test-941:
```
aspiwack, line 39: It's not worth testing twice. just rename this job into `cabal-test` and delete the above.

linear-base.cabal
```diff
@@ -115,8 +115,16 @@ library
         System.IO.Resource.Linear
         System.IO.Resource.Linear.Internal
         Unsafe.Linear
+    other-modules:
```
aspiwack, line 118: Please no `other-modules`
```suggestion
```

ghc92/Prelude/Linear/Generically/Types.hs
```diff
@@ -0,0 +1,26 @@
+-- | Prior to GHC 9.4, linear-base defined its own versions of @Generically@ and
+-- @Generically1@. As a temporary workaround to enable compilation on both GHC
+-- 9.4 and 9.2, this module exposes linear-base's own implementations of those
+-- types, while the 9.4 version simply re-exports @Generics.Linear@.
+module Prelude.Linear.Generically.Types
```
aspiwack, line 5: I suppose this module (and its alter-ego in `ghc94/`) should be an `Internal` module, shoudn't it?

ghc92/Prelude/Linear/Internal/TypeEq.hs
```diff
@@ -0,0 +1 @@
+module Prelude.Linear.Internal.TypeEq where
```
aspiwack, line 1: Add documentation for why this module exists, and explaining that when the module is removed, then we also should remove the corresponding ` {-# OPTIONS_GHC -Wno-dodgy-exports -Wno-unused-imports #-}`.

aspiwack (commented):
Just a few minor comments from me before we merge.

jeicher:
I know get a build error with `cabal build --allow-newer` because `cabal: Failed to build tasty-hedgehog-1.3.0.0 (which is required by test`. This is not caught by CI because it does `cabal build --allow-newer --disable-tests`.

aspiwack (approved):
For the `--allow-newer` issue, I'm ok with tests not working (though I wonder why this particular failure is happening). But you can also use the fixed version in the workflow.

I'm indifferent to the solution. Merge when you have chosen.

.github/workflows/ci.yaml
```diff
@@ -7,7 +7,7 @@ env:
 
 jobs:
   cabal-test:
-    name: cabal test
+    name: cabal test with GHC 9.4.1
```
aspiwack, line 10: Ah, this is causing the PR to be blocked. Change to
```suggestion
    name: cabal test
```

PR #438, opened 2023-02-16
Compatibility with 9.6

### LINKED ISSUE #437: Fails to build with GHC 9.6 alpha2
**Describe the bug**

Building linear-base 0.3.0 with GHC 9.6 fails with a type error. It seems that the `Solo` tuple is not handled correctly.

**To Reproduce**

```
cabal install linear-base
```

```
src/Data/Monoid/Linear/Internal/Semigroup.hs:168:3: error: [GHC-18872]
    • Couldn't match type ‘Many’ with ‘One’
        arising from a non-linear pattern
    • In the pattern: Solo x
      In an equation for ‘<>’: Solo x <> Solo y = Solo (x <> y)
      In the instance declaration for ‘Semigroup (Solo a)’
    |
168 |   Solo x <> Solo y = Solo (x <> y)
    |   ^^^^^^
```

**Expected behavior**

Success.

aspiwack:
This pull request makes linear-base compatible with GHC 9.6.

I've moved the compatibility files around to permit for many version transitions if we need to (even if I plan to drop 9.2 support soon).

Draft because this is very not done

- [ ] Fix the `Solo` instances
- [ ] Any other fixes?
- [ ] Add 9.6 to CI
- [ ] Formatting

Closes #437


aspiwack:
@matthew-healy since you participated in the previous such occurrence, would you have a look and tell me what you think of the general plan? (I'm sparing @jeicher because I know he wants to focus on his client's work at present).

matthew-healy:
@aspiwack Sure thing! From a very cursory glance it looks reasonable, but I'll take a more in-depth look tomorrow morning just to make sure.

aspiwack:
Using a quick and dirty matrix job to check against more versions of GHC. I'll have to fiddle with the branch protection options to reflect this.

src-version-changes/changed-in-ghc96/ghc94/Data/Tuple/Linear/Compat.hs
```diff
@@ -0,0 +1,16 @@
+{-# LANGUAGE LinearTypes #-}
+module Data.Tuple.Linear.Compat where
```
aspiwack, line 2: I'm pondering whether I want `Data.Tuple.Linear.Compat` to be hidden in Haddock as an internal module (applies to both copies).

aspiwack:
Some bikeshedding thoughts to ponder:

- We could rename `src-version-changes` to `src-something-snappier`
- We could rename `changed-in-ghcxx` to `ghcxx`
  - In which case, we could rename the subdirectory `ghcxx` and `ghcyy` to `before` and `after`, so that the directory structure is more predictable.

aspiwack:
I've tried using a more recent Nixpkgs version, but for whatever reason it ends up rebuilding a large proportion of the Haskell dependencies. I don't have time to investigate, and it would break CI. Maybe I'll join testing on GHC 9.6 with dropping support for GHC 9.2. @guibou I'd be appreciated if you could test this version with GHC 9.6 before I merge.

matthew-healy (approved):

aspiwack:
Actually, it was a bit hacky, but I managed to test with GHC 9.6. I think that most of the issues with Nix are in haskell-language-server and its dependencies. So it may be fixable, but I won't have time today. I'll merge as soon as CI is green.

PR #439, opened 2023-03-15
fix: build with GHC 9.6 (ghc-prim 0.10.0)

guibou:
Recent version of `ghc-prim` changed `Solo`.

With `ghc-prim < 0.10` , it was `data Solo = Solo a`, now it is `data Solo = MkSolo a` and `Solo` is a pattern.

This fixs the build with GHC 9.6 without breaking it with <9.6.

### DIFF
--- a/src/Data/Monoid/Linear/Internal/Semigroup.hs
+++ b/src/Data/Monoid/Linear/Internal/Semigroup.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE CPP #-}
 {-# OPTIONS_HADDOCK hide #-}
 
 -- | This module provides a linear version of 'Semigroup'.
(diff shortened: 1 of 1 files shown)


guibou:
Sorry, I realized that it conflicts with https://github.com/tweag/linear-base/pull/438

aspiwack:
Hi Guillaume,

I haven't merged #438 yet because I need to test against 9.6 for CI. Is this the only compatibility change that you've needed for 9.6? This would be encouraging.

guibou:
Hi Arnaud, nice to interact with you ;)

This change makes `linear-base` build correctly and all the dependencies in our codebase which depends on `linear-base` are building / working. However we are not using `linear-base` directly, so I cannot confirm about the semantic (but I'm rather confident).

May I suggest you merge this (or the similar change from https://github.com/tweag/linear-base/pull/438 which does not need `CPP`) as long as it pass the CI with others version of GHC? If the CI is ok with other GHC version, it won't break anything, and having it merged will ease upgrade to GHC 9.6 for users because they can just point to the main branch of the repo and get the working code, without to tig into the MR to find a candidate.

aspiwack:
The change is correct, indeed. I'll merge #438 tomorrow.

PR #449, opened 2023-05-03
Moving from gauge to tasty-bench

tbagrel1:
This commit replaces the `gauge` benchmark framework with `tasty-bench` for the `linear-base:bench` target. The API is the same in both tools, so the change is mostly transparent.

### DIFF
--- a/bench/Main.hs
+++ b/bench/Main.hs
@@ -1,12 +1,12 @@
 module Main where
 
 import qualified Data.Mutable.Array as Array
-import Data.Mutable.HashMap (hmbench)
-import Gauge
+import qualified Data.Mutable.HashMap as HashMap
+import Test.Tasty.Bench (defaultMain)
 
 main :: IO ()
 main = do
   defaultMain
-    [ hmbench,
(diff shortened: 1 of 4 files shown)


bench/Main.hs
```diff
@@ -1,12 +1,12 @@
...
+import qualified Data.Mutable.HashMap as HashMap
+import Test.Tasty.Bench (defaultMain)
 
 main :: IO ()
 main = do
   defaultMain
-    [ hmbench,
-      Array.benchmarks
+    [ Array.benchmarks,
+      HashMap.benchmarks
```
aspiwack, line 11: Any reason for switching these two around?

tbagrel1: I added some consistency around the import naming scheme, and I reordered them to follow both alphabetical order and the import order.

aspiwack: I don't see much value in alphabetical ordering of the benchmarks. But it doesn't hurt either, I suppose.

aspiwack (approved):

PR #451, opened 2023-05-03
More array benchmarks

aspiwack:
- Benchmark raw allocation
- Benchmark writes
- Benchmark against the `Seq` type

Still a draft because I need to format all this, and maybe I need to revisit a commit message: it would seem that the overhead that my refactoring introduced essentially compensates the difference between Vector and our linear arrays. Which proves that the extra `Ur` and `(,)` are probably where the issue is coming from (probably mostly `Ur` since, as we can see in #328 , the `(,)` are unboxed. So having the ability to unbox `Ur` will really be critical for the extra bit of performance).

That being said, there is something to be said for comparing the same scenarios. There is also something to be said for comparing idiomatic scenarios. In neither case is the comparison completely accurate anyway.

Here are my figures after this PR (note how good we are with writes, the figures are not too shabby honestly):

```
  arrays
    alloc
      Data.Array.Mutable.Linear:                                                 OK (0.41s)
        329  ns ±  29 ns
      Data.Vector:                                                               OK (0.45s)
        451  ns ±  23 ns
      Data.Sequence:                                                             OK (0.29s)
        3.88 μs ± 377 ns
    toList
      Data.Array.Mutable.Linear:                                                 OK (0.36s)
        5.70 μs ± 384 ns
      Data.Vector:                                                               OK (0.35s)
        5.82 μs ± 347 ns
      Data.Sequence:                                                             OK (0.58s)
        3.02 μs ±  95 ns
    map
      Data.Array.Mutable.Linear:                                                 OK (0.32s)
        2.24 μs ± 220 ns
      Data.Vector:                                                               OK (0.33s)
        2.62 μs ± 166 ns
      Data.Sequence:                                                             OK (1.14s)
        7.20 μs ± 436 ns
    reads
      Data.Array.Mutable.Linear:                                                 OK (0.44s)
        9.11 μs ± 355 ns
      Data.Vector:                                                               OK (0.32s)
        11.3 μs ± 755 ns
      Data.Sequence:                                                             OK (0.44s)
        73.8 μs ± 2.7 μs
    successive writes (very unfair to vector)
      Data.Array.Mutable.Linear:                                                 OK (0.37s)
        13.6 μs ± 759 ns
      Data.Vector:                                                               OK (0.57s)
        1.93 ms ± 184 μs
      Data.Sequence:                                                             OK (0.21s)
        116  μs ±  11 μs
```


tbagrel1 (approved):
I'm not sure I can convince myself that this code benchmarks each collection type without introducing extra noise that would impact performance.

That being said, at a syntactic/organisational level, the code is quite readable.

PR #457, opened 2023-06-01
Fix building with master

monoidal:
The holes now have to indicate a context.
`TypeInType` is deprecated and not needed in that file.


aspiwack (approved):
Thanks.

Issue #453, opened 2023-05-20
cycle doesn't terminate


alt-romes:
**Describe the bug**

The function `cycle` from `Data.List.Linear` never terminates

**To Reproduce**

Load up `ghci` with
```
take 5 (Data.List.Linear.cycle ['a','b'])
```

**Expected behavior**

It should return `"ababa"` and terminate. Currently it returns `"ababa"` but never terminates 🙂.

**Environment**

GHC 9.4.4 and 9.6.1

alt-romes:
I diagnosed this incorrectly. I was using linear `take` instead of the Prelude one, and linear `take` traverses the whole list. Apologies!

treeowl:
I believe the essential problem is that it's impossible to ever `consume` an infinite list. So `cycle` seems quite useless in a linear context. Solution: get rid of it, and add an equivalent of `Data.Sequence.cycleTaking`

treeowl:
I'm reopening because I think we should remove `cycle`.

treeowl:
I think the same is true of `repeat`.

treeowl:
I have no idea if something `cycle`-like is appropriate, but for the heck of it I decided to write one (following `Data.Sequence`). It is not at all trivial to

1. Use an optimized `Replicator` when available,
2. Avoid unnecessary `dupR`s, and
3. Avoid unnecessary `next`s.

I came up with a version using the following simple queue:

```haskell
module BoringQueue where
module Data.BoringQueue.Linear where
import Data.Unrestricted.Linear.Internal.Consumable (Consumable (consume), lseq)

infixl 5 :>
infixr 5 :<

-- | Strict cons list.
data CL a = CNil | !a :< !(CL a)

instance Consumable a => Consumable (CL a) where
  consume CNil = ()
  consume (a :< as) = a `lseq` consume as

-- | Strict snoc list.
data SL a = SNil | !(SL a) :> !a

instance Consumable a => Consumable (SL a) where
  consume SNil = ()
  consume (as :> a) = a `lseq` consume as

-- | A boring queue strict in its values.
data Queue a where
  Queue :: !Int -> !(CL a) %1-> !(SL a) %1-> Queue a

instance Consumable a => Consumable (Queue a) where
  consume (Queue _ front rear) = front `lseq` consume rear

size :: Queue a %1-> (Int, Queue a)
size (Queue s fr r) = (s, Queue s fr r)

data SizedQueue a where
[...]

PR #458, opened 2023-06-12
Deprecate cycle, repeat, and iterate

treeowl:
Deprecate `Data.List.Linear.{cycle,repeat,iterate}`. Infinite results cannot be consumed linearly, so they are not really useful in a linear context. They could be consumed by a program that exits via an exception, but I don't think that's something that should really be supported.

Begins to address #453

### DIFF
--- a/src/Data/List/Linear.hs
+++ b/src/Data/List/Linear.hs
@@ -284,14 +284,17 @@ or = foldl' (||) False
 -- # Building Lists
 --------------------------------------------------
 
+{-# DEPRECATED iterate "The result cannot be consumed linearly, so this function is not useful." #-}
 iterate :: (Dupable a) => (a %1 -> a) -> a %1 -> [a]
 iterate f a =
   dup2 a & \(a', a'') ->
     a' : iterate f (f a'')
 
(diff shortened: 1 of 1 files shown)


treeowl:
Alternative: just re-export the versions from `base`, since these versions have no advantage in linear programming.

aspiwack (approved):
Sorry about the delay. Yes, let's deprecate, it's most certainly the right call.

Not so clear about `Stream.delays` though. It could conceivably be used in an infinite process (which can be cancelled with regular thread cancelling).

treeowl:
What do you mean by "regular thread cancelling"? That tends to mean death by exception, no?

aspiwack:
Indeed, this is what I have in mind.

treeowl:
But that goes for the others too, or anything else infinite, right?

aspiwack:
Yes, you're right (also, I'm infinitely slow at responding on this thread).

Somewhat I was thinking “streams have effects”, which is true, but you can also want to do `mapM_` on an infinite list.

These things make sense, but probably shouldn't be front and center. We could have them in a separate module. So that you will have to realise, when importing these definitions, that they're not as innocent as their equivalent in non-linear Haskell. What do you think?

PR #461, opened 2023-08-26
Implements `Semigroup` / `Monoid` instances for `Ur`

konn:
# Description
This PR adds linear `Semigroup` and `Monoid` instances for `Ur`, which turns unrestricted ones into linear ones.

## Motivation

This instance is particuarly useful when one extract unrestricted monoidal value out of linear values; e.g. `foldMap` of Pull and/or Push Arrays.
I used them to [test my custom vector-generic `fromVector` for pull arrays without allocating vector](https://github.com/konn/linear-array-extra/blob/main/test/Data/Array/Polarized/Pull/ExtraSpec.hs#L50) ([corresponding instance definition](https://github.com/konn/linear-array-extra/blob/c6c89ac424620a2ebde41a52dce8241c2799e65f/test/Linear/Array/Extra/TestUtils.hs#L249-L253))


aspiwack (approved):
Thank you!

Issue #466, opened 2023-10-18
Word does not have instances for the Data.Num.Linear typeclasses 


Qqwy:
**Describe the bug**
It is currently impossible to use `Word` in a module using `Prelude.Linear` as prelude, because [instances for Additive, Multiplicative, AddIdentity, MultIdentity, AdditiveGroup, Semiring, Ring, FromInteger and Num](https://github.com/tweag/linear-base/blob/074aeb8502c502f05ac0de142f3ccaf84a6b5ecc/src/Data/Num/Linear.hs#L211-L213) are missing for `Word`:

**To Reproduce**

```haskell
{-# LANGUAGE NoImplicitPrelude #-}
module Example where
import Prelude.Linear

add1 :: Word %1 -> Word
add1 = (+ 1)
```

**Expected behavior**
This should compile.

**Actual behavior**
Compile error: `No instance for (Additive Word) arising from the use of '+'`.

aspiwack:
I would appreciate a PR, indeed. Thanks!

Issue #403, opened 2022-04-10
`Integer` is not `Additive`


andreasabel:
(Sorry, newbie here.)
Trying to add `Integer`s I am getting error:
```
No instance for (Additive Integer) arising from a use of ‘+’
```
Is there a deeper reason for this or is it just not implemented?

aspiwack:
I'm not too sure how to handle `Integer` linearly, to be honest. It's quite possibly out of ignorance: I don't know how `Integer` is implemented.

What we have to consider is that for `(+)` to be linear on `Integer`, it needs to guarantee that it evaluates every thunk (or pass them into the resulting integer without dropping or duplicating any). It's quite obvious for `Int` and friends, which have a single thunk which is necessarily forced. But for `Integer`, I don't know.

treeowl:
Why would every thunk have to be forced? Can't a non-linear implementation expose a linear interface? That said, `Integer` is completely strict internally, so there's nothing to worry about.

aspiwack:
> Why would every thunk have to be forced?

Every linear argument has to be consumed exactly once. This is the contract. A thunk, in a linear argument, can point to linear arguments which haven't been consumed yet (in which case, forcing the thunk will, transitively, consume the linear arguments it points to). So you really need to (ultimately force all the thunks exactly once. Otherwise you would break the contract with these arguments that you've closed over.

treeowl:
I don't think your reasoning is entirely valid.

First, not all thunks are "exposed". If `Integer` were implemented using a data structure with thunks, that would have no impact on its semantics, which are designed and documented to be strict. There are quite a few data structures of that sort—see Okasaki in particular, but also Hinze and Paterson. No one can hang extras off those thunks, so their existence is irrelevant.

Second, you actually can't know for sure how many times a thunk will be forced, short of an expensive operation. GHC's "lazy blackholing" mechanism reserves the right to evaluate the same thunk twice in multiple threads, unless that thunk uses `noDuplicate#` to prevent it. This should only be done when truly necessary, as it is very definitely not free.

Third, while a linear argument can only be *consumed* once, it does not inherently have to be *inspected* only once. Yes, we would write

```haskell
foo :: Maybe Bar %1-> (Maybe Bar, Bool)
foo Nothing = (Nothing, False)
foo (Just x) = (Just x, True)
```

But another valid implementation is

```haskell
foo = toLinear (\m -> let !b = isJust m in (m, b))
```

aspiwack:
[...]

PR #467, opened 2023-10-20
Adds missing Data.Num.Linear.* instances for Word, Integer, Natural, Float, Word8/16/32/64 Int8/16/32/64

Qqwy:
This PR fixes two (related) issues (#403 and #466):

- Movable (and derived `Consumable`/`Dupable`) instances are added for `Integer` and `Natural`. This is safe since these types are represented as strict sum types whose elements are unlifed unboxed primitives so there is no way laziness might break linearity in there. (See also the [last part of the discussion in #403](https://github.com/tweag/linear-base/issues/403#issuecomment-1706679338))
- Instances for the typeclasses in `Data.Num.linear` (`Additive`/`AddIdentity`/`AdditiveGroup`/`Multiplicative`/`MultIdentity`/`Semiring`/`Ring`/`Num`) for the general-purpose number types that exist in the Prelude/`base` which were still missing them:
  - `Word`
  - `Float`
  - `Integer`
  - `Natural`
  - `Word8`/`Word16`/`Word32`/`Word64`
  - `Int8`/`Int16`/`Int32`/`Int64`

Still missing are:
- Instances for `Ratio a` and `Complex a`.
- Instances for `Foreign.C.Types` like `CInt`/`CFloat` etc.
- Instances for `Foreign.Ptr` types like `IntPtr`, `WordPtr`.

Those could be added just as easily but because I was not 100% sure whether adding linear instances for them would be (morally) correct or have unintended consequences that might break linearity in ineffable ways, I left them alone for now.

### DIFF
--- a/src/Data/Num/Linear.hs
+++ b/src/Data/Num/Linear.hs
@@ -46,6 +46,9 @@ import Data.Monoid.Linear
 import Data.Unrestricted.Linear
 import qualified Unsafe.Linear as Unsafe
 import qualified Prelude
+import qualified Data.Word
+import GHC.Num.Natural (Natural)
+import qualified Data.Int
 
 -- | A type that can be added linearly.  The operation @(+)@ is associative and
 -- commutative, i.e., for all @a@, @b@, @c@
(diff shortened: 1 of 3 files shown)


Qqwy:
The `Movable` instances introduces in this PR pattern matches down to the internals of `Integer` and `Natural` just like was done for `Word`/`Int` etc. before. Doing so is only possible by pulling in the `ghc-bignum` package. As far as I can tell `base` itself depends on it so I believe adding this 'extra' dependency does not bloat the library in any way.

src/Data/Num/Linear.hs
```diff
@@ -208,38 +211,143 @@ instance (MultIdentity a) => Monoid (Product a) where
 instance (AddIdentity a) => Monoid (Sum a) where
   mempty = Sum zero
 
+{- ORMOLU_DISABLE -}
```
Qqwy, line 214: I took the liberty of re-ordering these by instance rather than by class because I believe it's easier on the eyes. Ormolu is disabled in this part of the code because it would add a newline between each `deriving` line. I hope you agree. If not, we can always change it back.

src/Data/Num/Linear.hs
```diff
@@ -208,38 +211,143 @@ instance (MultIdentity a) => Monoid (Product a) where
...
+deriving via MovableNum Data.Word.Word64 instance Additive Data.Word.Word64
+deriving via MovableNum Data.Word.Word64 instance AddIdentity Data.Word.Word64
+deriving via MovableNum Data.Word.Word64 instance AdditiveGroup Data.Word.Word64
+deriving via MovableNum Data.Word.Word64 instance Multiplicative Data.Word.Word64
+deriving via MovableNum Data.Word.Word64 instance MultIdentity Data.Word.Word64
+deriving via MovableNum Data.Word.Word64 instance Semiring Data.Word.Word64
+deriving via MovableNum Data.Word.Word64 instance Ring Data.Word.Word64
+deriving via MovableNum Data.Word.Word64 instance FromInteger Data.Word.Word64
+deriving via MovableNum Data.Word.Word64 instance Num Data.Word.Word64
+{- ORMOLU_ENABLE -}
```
aspiwack, line 353: Please add a newline to please git
```suggestion
{- ORMOLU_ENABLE -}

```

src/Data/Num/Linear.hs
aspiwack: Disabling Ormolu is a little unorthodox, but I must admit that in this case, it looks like it's worth it.

aspiwack (commented):
This all looks good. Thank you very much. I noticed that the CI hadn't ran yet. So I triggered it, and I'll (auto)merge when green.

aspiwack:
Actually, can I, instead, ask you to add the newline to the appropriate commit to avoid a silly commit? Sorry about the rather trivial request.

Qqwy:
Ormolu/git should be happy now; of course the CI only runs after being approved once more.

aspiwack:
I'm guessing Github will keep requiring an approval for CI until you have at least one commit on master. I relaunched it.

aspiwack (approved):

PR #469, opened 2023-11-13
Remove the ... & \case -> ... workaround

monoidal:
Since #442 we require GHC >= 9.2. We can drop the workaround, which was needed for 9.0 only.
Tested by building with 9.2 and 9.6.


aspiwack (approved):
Thanks. I've been wanting to do this for a while.

PR #470, opened 2023-11-28
Add safeIndex for Pull Arrays

kephas:
I wonder if there was a reason behind the absence of a `Array a %1 -> Int -> (Maybe a, Array a)` function.

### DIFF
--- a/src/Data/Array/Polarized/Pull.hs
+++ b/src/Data/Array/Polarized/Pull.hs
@@ -28,6 +28,7 @@ module Data.Array.Polarized.Pull
     split,
     reverse,
     index,
+    safeIndex,
   )
 where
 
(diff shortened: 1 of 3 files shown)


b-mehta:
As mentioned in #471, `index` isn't safe and hence neither is `safeIndex`. There may be a save by changing it to return `Maybe (a, Array a)`, where the second array has the element removed (and everything after it shifted forward by one).

kephas:
As `index` isn't safe in terms of linearity, the goal was just to have a function that wouldn't fail with an exception.

aspiwack:
Hi @kephas , thanks for your PR.

Is this `safeIndex` (or just `index`) something that you have some need for. Because in the discussion on #471 , we seem to be leaning rather on the side of removing this function altogether. So I guess, what I'm asking is: do you have opposing arguments that'd help us take an informed decision?

sjoerdvisscher:
@kephas Thanks for pointing out that `index` is used in the example. We have removed `index` now and also updated the example. Given these changes, are you OK with closing this PR?

kephas:
I had wondered why there isn't a `uncons`, the API is so much better with it! The example is really clearer too.

PR #473, opened 2024-01-15
In scope-passing style: use a `Movable b` instead of `Ur b`

aspiwack:
Functions of the form

```haskell
f :: (A %1 -> Ur b) %1 -> Ur b
```

are now of the form

```haskell
f :: Movable b => (A %1 -> b) %1 -> b
```

The new type is strictly more general. Technically this seems to involve some extra allocations here and there. I expect it to be negligible (we don't want to call too many scoped functions anyway).

The extra allocation is interesting, in that it's not always strictly necessary: when returning an `Int`, I don't need to produce an `Ur Int` to make sure that I've actually forced everything, just returning the forced `Int` would do the trick. Yet we use an `Ur Int` to communicate that we've indeed done the job (in some cases the optimiser can actually remove the extra allocation, but not in every case as far as I can tell). Maybe there are cheaper way to tell the compiler that we've, in fact, moved the value. But that's a question for the future.

See https://github.com/tweag/linear-base/discussions/468 for the initial discussion.

### DIFF
--- a/src/Data/Array/Mutable/Linear/Internal.hs
+++ b/src/Data/Array/Mutable/Linear/Internal.hs
@@ -62,11 +62,11 @@ data Array a = Array (Array# a)
 -- | Allocate a constant array given a size and an initial value
 -- The size must be non-negative, otherwise this errors.
 alloc ::
-  (HasCallStack) =>
+  (HasCallStack, Movable b) =>
   Int ->
   a ->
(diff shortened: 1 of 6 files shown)


Qqwy:
@aspiwack Thanks for tagging me! I think you've caught them all. :rocket:

aspiwack:
Let me merge then.

PR #475, opened 2024-04-04
Remove pull array index (unsafe), add uncons.

sjoerdvisscher:


b-mehta:
I agree `uncons` is safe and is what's actually needed in the example, but in the Nothing case `loop` should return an empty array, I'm pretty sure.

sjoerdvisscher:
> I agree `uncons` is safe and is what's actually needed in the example, but in the Nothing case `loop` should return an empty array, I'm pretty sure.

Thanks what it is doing. The error is a filler for a function that should never be called. I kept it from the original example, and my first reaction was the same as yours, so perhaps we should fix it. I guess adding an `empty` function would clarify things.

src/Data/Array/Polarized/Pull/Internal.hs
```diff
@@ -110,6 +114,7 @@ split k (Array f n) = (fromFunction f (min k n), fromFunction (\x -> f (x + k))
 reverse :: Array a %1 -> Array a
 reverse (Array f n) = Array (\x -> f (n + 1 - x)) n
 
--- | Index a pull array (without checking bounds)
-index :: Array a %1 -> Int -> (a, Array a)
-index (Array f n) ix = (f ix, Array f n)
+-- | Decompose an array into its head and tail, returns @Nothing@ if the array is empty.
+uncons :: Array a %1 -> Maybe (a, Array a)
+uncons (Array _ 0) = Nothing
+uncons (Array f n) = Just (f 0, fromFunction (\x -> f (x + 1)) (n - 1))
```
aspiwack, line 120: Thought: a natural generalisation of `uncons` is `splitAt`, with

```haskell
splitAt :: Int -> Array a %1 -> Maybe (Array a, a, Array a)
splitAt i (Array f n)
  | n <= i = Nothing
  | otherwise = Just (fromFunction f i, f i, fromFunction (\x -> f (x+i+1)) (n-i-1)
```

With `uncons` is (up to isomorphism) `splitAt 0`.

Do you think it's worth exposing this generalisation?

sjoerdvisscher: When implementing `uncons` with `splitAt 0`, is there a way to discard the initial array in a total way? It should be empty, but there's no proof.

aspiwack: We don't appear to have made such a function available. But we should. Of course, it'd be partial and be an error if the length is non-0.

sjoerdvisscher: Good point about `splitAt` being the composition of `split` and `uncons`. It feels like that `splitAt` doesn't add much then?

aspiwack: Maybe not. Let's merge this PR at any rate.

You know what's funny though? `splitAt = traverse uncons split` may be the first time I see the `Traversable (a,)` instance show up in practice.

aspiwack (approved):

PR #478, opened 2024-09-11
Improve wording about Ur

### LINKED ISSUE #476: Clarify Ur documentation
Perhaps change

> Its use is restricted while an argument in a non-linear function is unrestricted.

to

> Its use is restricted. By contrast, an argument in a non-linear function is unrestricted.

The former is easy to misinterpret as "Its use is restricted during the time period that an argument in a non-linear function is unrestricted", which is meaningless nonsense.

https://github.com/tweag/linear-base/blob/16795c74588d013bbb2d4cd7954fd129b8ad2c46/src/Data/Unrestricted/Linear.hs#L7

aspiwack:
Fixes #476.

@tomjaguarpaw does this look good?

### DIFF
--- a/src/Data/Unrestricted/Linear.hs
+++ b/src/Data/Unrestricted/Linear.hs
@@ -4,7 +4,7 @@
 -- = /Critical/ Definition: Restricted
 --
 -- In a linear function @f :: a %1-> b@, the argument @a@ must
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Yes! Thanks :)

PR #479, opened 2024-09-11
Bump nixpkgs, bump GHC to 9.6.6 and stackage to LTS 22.34

tbagrel1:
This pull request changes the default version of GHC to be the one used by the latest stackage LTS available at the moment.

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -21,13 +21,26 @@ source-repository head
     type:     git
     location: https://github.com/tweag/linear-base
 
-common warnings
-    ghc-options: -Wall -Wcompat -Wincomplete-record-updates -Wincomplete-uni-patterns -Wnoncanonical-monad-instances
+common build-opts
+    ghc-options: -O2 -Wall -Wcompat -Wincomplete-record-updates -Wincomplete-uni-patterns -Wnoncanonical-monad-instances
                  -- Additional warnings we may consider adding:
                  -- * -Wredundant-constraints : would need deactivating in the modules which use Nat
+common rts-opts-multithread
+    ghc-options: -threaded -rtsopts "-with-rtsopts=-N"
+common rts-opts-monothread-stats
+    ghc-options: -rtsopts "-with-rtsopts=-T"
 
 library
-    import: warnings
+    import: build-opts
+    hs-source-dirs:     src
+    if impl(ghc >= 9.4.0)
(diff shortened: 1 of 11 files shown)


linear-base.cabal
```diff
@@ -21,13 +21,26 @@ source-repository head
...
-common warnings
-    ghc-options: -Wall -Wcompat -Wincomplete-record-updates -Wincomplete-uni-patterns -Wnoncanonical-monad-instances
+common build-opts
+    ghc-options: -O2 -Wall -Wcompat -Wincomplete-record-updates -Wincomplete-uni-patterns -Wnoncanonical-monad-instances
                  -- Additional warnings we may consider adding:
                  -- * -Wredundant-constraints : would need deactivating in the modules which use Nat
+common rts-opts-multithread
+    ghc-options: -threaded -rtsopts -with-rtsopts=-N
+common rts-opts-monothread
+    ghc-options: -threaded -rtsopts -with-rtsopts=-N1
```
aspiwack, line 31: Why not simply removing `-threaded`?
```suggestion
    ghc-options: ""
```

src/Data/Unrestricted/Linear/Internal/Instances.hs
```diff
@@ -186,7 +186,8 @@ instance (KnownNat n, Consumable a) => Consumable (V n a) where
 
 instance (KnownNat n, Dupable a) => Dupable (V n a) where
   dupR (V xs) =
-    V . Unsafe.toLinear (Vector.fromListN (V.theLength @n))
+    V
+      . Unsafe.toLinear (Vector.fromListN (V.theLength @n))
```
aspiwack, line 190: What happened here?

linear-base.cabal
tbagrel1: Apparently there are [some non-negligible](https://discourse.haskell.org/t/make-ghc-threaded-the-default-rts/4220/6) differences, which indicates we should probably use no option at all for the bench case.

src/Data/Unrestricted/Linear/Internal/Instances.hs
tbagrel1: Our ormolu version is given by stackage AFAIK, which I bumped.
I suppose the new ormolu version has some breaking changes related to default operator fixity/formatting.
(I didn't change that manually)

aspiwack (approved):

PR #480, opened 2024-09-11
Project structure: extract examples (quicksort, foreign heap/list, ...) to internal library to allow easy run/test/benchmarking

tbagrel1:
Using an internal library, in addition to the main library, seems to be supported by both Cabal and stack now, despite a warning from stack saying the opposite.

With this change, the `examples` test suite becomes the `examples` internal library, and I introduced `test-examples` to act as the test suite for shared examples. This allows us, for example, to also cherry pick some examples we want to benchmark, and share the code between the test suite and benchmark suite (which wasn't possible before).

Other changes:
- Renamed the default benchmark suite from `mutable-data` to `bench`, for consistency.
- Moved the `genericTraverse` tests from `examples` to the main `test` suite

This PR follows and builds upon https://github.com/tweag/linear-base/pull/479


linear-base.cabal
```diff
@@ -171,38 +189,28 @@ test-suite test
         tasty,
         tasty-hedgehog >= 1.2,
         mmorph,
-        vector
+        vector,
+        linear-generics
 
-test-suite examples
+test-suite test-examples
```
aspiwack, line 195:
```suggestion
test-suite examples
```
Should work. Because different units are namespaced between their types, so it'll be referred to as something like `test:examples` (instead of `test:test-examples` in your case).

aspiwack (approved):
Up to small point below, merge when you're ready.

linear-base.cabal
tbagrel1: It doesnt seem to work:

```
library examples
    import: build-opts
    hs-source-dirs: examples
    
[...]

test-suite examples
    import: build-opts
    import: rts-opts-multithread
    type:             exitcode-stdio-1.0
    main-is:          Main.hs
    hs-source-dirs:   test/examples
```

gives

```
Configuring linear-base-0.4.0...
There exist several components with the same name: 'examples'. This is a bug in your .cabal file.
CallStack (from HasCallStack):
  error, called at libraries/Cabal/Cabal-syntax/src/Distribution/PackageDescription/Utils.hs:18:15 in Cabal-syntax-3.10.3.0:Distribution.PackageDescription.Utils
Progress 1/2

Error: [S-7282]
       Stack failed to execute the build plan.
       
       While executing the build plan, Stack encountered the error:
       
       [S-7011]
       While building package linear-base-0.4.0 (scroll up to its section to see the error)
       using:
[...]
```

aspiwack: My bad then.

PR #490, opened 2025-08-05
Support for Data.Complex

ksrky:
I'm aiming to make [Data.Complex](https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-Complex.html) in the base package  to be an instance of each linear version of type class.
I would like your opinion on which type classes are preferable for `Complex a` and which functions in `Data.Complex` should be reimplemented in linear-base.

### DIFF
--- a/src/Data/Num/Linear.hs
+++ b/src/Data/Num/Linear.hs
@@ -42,6 +42,7 @@ where
 
 -- TODO: flesh out laws
 
+import Data.Complex
 import qualified Data.Int
 import Data.Monoid.Linear
 import Data.Unrestricted.Linear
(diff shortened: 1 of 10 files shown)


src/Data/Num/Linear.hs
```diff
@@ -350,4 +351,41 @@ deriving via MovableNum Data.Word.Word64 instance Semiring Data.Word.Word64
 deriving via MovableNum Data.Word.Word64 instance Ring Data.Word.Word64
 deriving via MovableNum Data.Word.Word64 instance FromInteger Data.Word.Word64
 deriving via MovableNum Data.Word.Word64 instance Num Data.Word.Word64
+
+deriving via
+  MovableNum (Complex a)
+  instance
+    (Movable a, Prelude.RealFloat a) => Additive (Complex a)
```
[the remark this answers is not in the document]
ksrky: Thanks for replying. Actually, this PR has become unnecessary for me in the last few days.
My use case was to represent a complex vector for quantum computation, i.e. datatype `Complex` is only used like `type CC = Complex Double`.
Reading your comments, I thought the type `Complex a` alone where `a` is arbitrary does not need to be treated like a complex number.
In this respect, it may be more appropriate to view `Complex` as quadratic extension field rather than a complex number. This is because complex numbers are quadratic extensions of real numbers, whereas `Complex` is defined like for any field.
However I don't know this view suggest a reasonable solution.
Either way, this is no longer necessary at this point.

src/Data/Num/Linear.hs
```diff
@@ -350,4 +351,41 @@ deriving via MovableNum Data.Word.Word64 instance Semiring Data.Word.Word64
...
+  instance
+    (Movable a, Prelude.RealFloat a) => AddIdentity (Complex a)
+deriving via
+  MovableNum (Complex a)
+  instance
+    (Movable a, Prelude.RealFloat a) => AdditiveGroup (Complex a)
+deriving via
+  MovableNum (Complex a)
+  instance
+    (Movable a, Prelude.RealFloat a) => Multiplicative (Complex a)
```
ksrky, line 370: Sorry, we can't just derive `Multiplicative`.
This is probably calculated as (x1+iy1)*(x2+iy2)=(x1*x2)+i(y1*y2)

aspiwack: Indeed, but the `MovableNum` implementation takes care of that just fine. It's just unclear how useful it is to add these instances.

PR #492, opened 2025-10-21
Make List.zipWith as lazy as expected

### LINKED ISSUE #491: zipWith throws NonTermination exception
**Describe the bug**
If I define Fibonacci sequence and use zipWith from linear-base, it throws exception starting at 3rd number

```
euler-hs-linear: Uncaught exception ghc-internal:GHC.Internal.Control.Exception.Base.NonTermination:
<<loop>>
```

I reduced my code to
```

import Data.Word
import qualified Data.List.Linear as Linear
import qualified Data.Num.Linear as Linear
import Data.Unrestricted.Linear (Consumable, lseq)

test :: Word32
test = fib !! 2
  where
    fib = 1 : 2 : Linear.zipWith (Linear.+) fib (tail fib)

-- my implementation of linear zipWith
linZipWith ∷ (Consumable a, Consumable b) ⇒ (a ⊸ b ⊸ c) → [a] ⊸ [b] ⊸ [c]
linZipWith _ [] ys = ys `lseq` []
linZipWith _ xs [] = xs `lseq` []
linZipWith f (x:xs) (y:ys) = f x y : linZipWith f xs ys
```

**To Reproduce**
Using index greater than 1 into fib will throw exception.

**Expected behavior**
I expected type error, if any. My implementation linZipWith works fine, no type errors, no exception.

**Environment**
ghc 9.12.2
linear-base 0.5

**Additional context**
I have no experience with Linear Types. I wanted to try it in some simple code.

aspiwack:
Fixes #491

### DIFF
--- a/src/Data/List/Linear.hs
+++ b/src/Data/List/Linear.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE LinearTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE NoImplicitPrelude #-}
@@ -79,6 +80,7 @@ module Data.List.Linear
     zip3,
     zipWith,
     zipWith',
+    zipFold,
     zipWith3,
     unzip,
     unzip3,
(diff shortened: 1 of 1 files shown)


src/Data/List/Linear.hs
```diff
@@ -354,18 +354,34 @@ zip3 :: (Consumable a, Consumable b, Consumable c) => [a] %1 -> [b] %1 -> [c] %1
...
+    (\a as -> ([], Just (Left (a :| as))))
+    (\b bs -> ([], Just (Right (b :| bs))))
+
+zipWithk :: forall r a b c. (a %1 -> b %1 -> c) -> (c %1 -> r %1 -> r) -> r -> (a %1 -> [a] %1 -> r) -> (b %1 -> [b] %1 -> r) -> [a] %1 -> [b] %1 -> r
+zipWithk f cons nil lefta leftb =
+  go
+  where
+    go :: [a] %1 -> [b] %1 -> r
+    go [] [] = nil
+    go (a : as) [] = lefta a as
```
tbagrel1: It's not entierely clear why the `left{a,b}` functions are taking head and tail as separate arguments. I guess the motivation is to make a `NonEmpty` list in `zipWith'` without having to call `NonEmpty.fromList`?

src/Data/List/Linear.hs
```diff
@@ -354,18 +354,34 @@ zip3 :: (Consumable a, Consumable b, Consumable c) => [a] %1 -> [b] %1 -> [c] %1
...
-    (cs, rest) -> (f a b : cs, rest)
+zipWith' f =
+  zipWithk
+    f
+    (\c !(cs, rest) -> ((c : cs), rest))
+    ([], Nothing)
+    (\a as -> ([], Just (Left (a :| as))))
+    (\b bs -> ([], Just (Right (b :| bs))))
+
+zipWithk :: forall r a b c. (a %1 -> b %1 -> c) -> (c %1 -> r %1 -> r) -> r -> (a %1 -> [a] %1 -> r) -> (b %1 -> [b] %1 -> r) -> [a] %1 -> [b] %1 -> r
```
tbagrel1:
```suggestion
zipWithK :: forall r a b c. (a %1 -> b %1 -> c) -> (c %1 -> r %1 -> r) -> r -> (a %1 -> [a] %1 -> r) -> (b %1 -> [b] %1 -> r) -> [a] %1 -> [b] %1 -> r
```

test/Test/Data/List.hs
```diff
@@ -41,3 +44,19 @@ take_length = property $ do
...
       annotate "Prelude.length xs < n"
       Prelude.length (List.take n xs) === Prelude.length xs
+
+zipWith_lazy :: Property
+zipWith_lazy = property $ do
+  _ <- eval $ Prelude.head xs
+  Prelude.return ()
+  where
+    xs :: [Word]
+    xs = List.zipWith (Num.+) (0 : error "bottom") [0 .. 42]
```
tbagrel1: What is the motivation for using static bounds for ranges instead of property-supplied ones?
(same below)

src/Data/List/Linear.hs
```diff
@@ -354,18 +354,34 @@ zip3 :: (Consumable a, Consumable b, Consumable c) => [a] %1 -> [b] %1 -> [c] %1
...
+zipWithk f cons nil lefta leftb =
+  go
+  where
+    go :: [a] %1 -> [b] %1 -> r
+    go [] [] = nil
+    go (a : as) [] = lefta a as
+    go [] (b : bs) = leftb b bs
+    go (a : as) (b : bs) = cons (f a b) (go as bs)
 
 zipWith3 :: forall a b c d. (Consumable a, Consumable b, Consumable c) => (a %1 -> b %1 -> c %1 -> d) -> [a] %1 -> [b] %1 -> [c] %1 -> [d]
```
tbagrel1: Why does `zipWith3` doesn't get the same treatment with a `zipWithK3`?

src/Data/List/Linear.hs
```diff
@@ -354,18 +354,34 @@ zip3 :: (Consumable a, Consumable b, Consumable c) => [a] %1 -> [b] %1 -> [c] %1
...
 
 zipWith :: (Consumable a, Consumable b) => (a %1 -> b %1 -> c) -> [a] %1 -> [b] %1 -> [c]
-zipWith f xs ys =
-  zipWith' f xs ys & \(ret, leftovers) ->
-    leftovers `lseq` ret
+zipWith f =
+  zipWithk f (:) [] consume2 consume2
+  where
+    consume2 :: forall x y z. (Consumable x, Consumable y) => x %1 -> y %1 -> [z]
+    consume2 x y = x `lseq` y `lseq` []
```
tbagrel1: In `zipWith3` you do ```(x, y) `lseq` []``` instead, any reason to prefer one or the other?

aspiwack: I don't think it makes much difference after optimisation. In principle, even without optimisation ``x `lseq` y `lseq` []`` is going to be one fewer allocation.

src/Data/List/Linear.hs
aspiwack: Because there's no `zipWith3'` with rests. And that, in turn, is because we didn't give it a type (using `Either` is a bit of a shortcut for `zipWith'`, maybe it was a bad shortcut).

We could make a `zipWith3K` if we decide to export it.

tbagrel1:
Thanks for taking time to address my comments! LGTM

PR #493, opened 2025-12-13
Control Monad instance for linear arrows

konn:
`FUN r` has only a data Functor instance, but it can be given Control Monad structure provided that `r` is `Dupable`. This PR just adds that instance.

### DIFF
--- a/src/Control/Functor/Linear/Internal/Reader.hs
+++ b/src/Control/Functor/Linear/Internal/Reader.hs
@@ -1,5 +1,9 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# OPTIONS -Wno-orphans #-}
 {-# LANGUAGE LinearTypes #-}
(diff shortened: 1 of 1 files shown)


aspiwack (approved):

konn:
> Indeed. In fact, I believe it's an equivalence: if `FUN 'One r` is an applicative functor, then `r` is dupable (_i.e._ a comonoid).

Agreed! We can use `(,) <$> id <*> id` to implement `dup`.

> I think this free with the lax-monoidal functor presentation of applicatives (but I didn't work out if the laws actually align, surely they do).

Sounds interesting, let me know if you make it out!

> Question: do you need that soon (aka should I cut a release)?

Well, they are definetely useful (for, e.g., duplicating Linearly tokens until linear constraints land in GHC), but we can do the same thing with `Reader` monads. So I'm not in a hurry :-)

aspiwack:
It's likely that I avoided adding a monad for `FUN 'One` because I don't like the error messages that you get when you forget an argument in a monad. But they aren't really much better without a monad instance, presumably. So let's merge this.

I'll cut a release in January.

PR #494, opened 2025-12-30
Fix: use Unsatisfiable for banned instances (unrestricted Monoid / Semigroup for linear containers)

konn:
Currently, `Prelude.Semigroup` instance for (linear) HashMap / Set results in a *runtime* error.
This PR explicitly bans these instances by using `Unsatisfiable` type constraint family.

### DIFF
--- a/src/Data/HashMap/Mutable/Linear/Internal.hs
+++ b/src/Data/HashMap/Mutable/Linear/Internal.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE InstanceSigs #-}
@@ -25,7 +27,9 @@ import qualified Data.Functor.Linear as Data
 import Data.Hashable
 import qualified Data.Maybe as NonLinear
 import Data.Unrestricted.Linear
+import GHC.TypeLits (ErrorMessage (..))
 import Prelude.Linear hiding (filter, insert, lookup, mapMaybe, read, (+))
+import Prelude.Linear.Unsatisfiable (Unsatisfiable, unsatisfiable)
 import Unsafe.Coerce (unsafeCoerce)
 import qualified Unsafe.Linear as Unsafe
 import Prelude ((+))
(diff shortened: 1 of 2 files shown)


konn:
Ah, indeed there is already one - `Prelude.Linear.Unsatisfiable. In this time, I just use them.

src/Data/HashMap/Mutable/Linear/Internal.hs
```diff
@@ -481,8 +485,8 @@ instance Data.Functor (HashMap k) where
         )
         arr
 
-instance Prelude.Semigroup (HashMap k v) where
-  (<>) = error "Prelude.<>: invariant violation, unrestricted HashMap"
+instance (Unsatisfiable ('Text "Prelude.<>: invariant violation, unrestricted HashMap")) => Prelude.Semigroup (HashMap k v) where
```
aspiwack: The error message wasn't meant to be read as the function can never be called. If we are making it into an error which reaches the user, then we should have a more user-friendly message, such as:

```suggestion
instance (Unsatisfiable ('Text "Using Prelude's Semigroup methods on a Data.HashMap.Mutable.Linear is vacuous as there can't be an unrestricted such Hashmap")) => Prelude.Semigroup (HashMap k v) where
```

src/Data/Set/Mutable/Linear/Internal.hs
```diff
@@ -70,8 +75,11 @@ fromList xs f =
 -- # Typeclass Instances
 -------------------------------------------------------------------------------
 
-instance Prelude.Semigroup (Set a) where
-  (<>) = Prelude.error "Prelude.(<>): invariant violation, unrestricted Set"
+instance
+  (Unsatisfiable ('Text "Prelude.<>: invariant violation, unrestricted Set")) =>
```
aspiwack:
```suggestion
  (Unsatisfiable ('Text "Using Prelude's Semigroup methods on a Data.Set.Mutable.Linear is vacuous as there can't be an unrestricted such Set")) =>
```

treeowl:
An `Unsatisfiable` constraint on an instance serves two main purposes: 

1. It documents (in the type error) that the instance is not *unintentionally* missing. This prevents, e.g., feature requests to have it added. 
2. It ensures that no one wastes (further) time trying to write said instance. It can't be written and someone has checked carefully that that's the case.

aspiwack:
Fair enough.

Are the messages I suggest clear enough? I haven't spend much time wordsmithing.

konn:
@treeowl Thank you for clarifying the purpose of `Unsatisfiable`. Yes, that's exactly my intention to prevent illegal orphan instances.

@aspiwack Thanks! I think it looks good, and committed into the PR.

PR #495, opened 2025-12-30
feat: Adds missing instances for `NonEmpty` and adds some Linear counterpart of `Data.List.NonEmpty`

konn:
I had some linear code involving `NonEmpty`, and misses its `Traversable` instance and some helper functions.
This PR adds those.


aspiwack (approved):
This is great. Thanks.

PR #501, opened 2026-02-23
Add instance Consumable/Dupable/Movable for ByteString, ShortByteString and Text

### LINKED ISSUE #412: Add instance Consumable/Dupable/Movable for ByteString and Text
`linear-base` already depends on `bytestring` (via `hashable`) and `text`. It would be awesome to provide `Consumable`, `Dupable` and `Movable` instances for `ByteString` and `Text`, because they cannot be defined in respective packages. I could have used them in https://github.com/Bodigrim/linear-builder/

Bodigrim:
Closes #412. 

`ByteString`, `ShortByteString` and `Text` are simple immutable data, so they are trivially `Movable`. `linear-base` already depends on `text` and (transitively) on `bytestring`, so this does not incur new dependencies.

### DIFF
--- a/linear-base.cabal
+++ b/linear-base.cabal
@@ -135,6 +135,7 @@ library
     default-language: Haskell2010
     build-depends:
         base >=4.16 && <5,
+        bytestring,
         containers,
         ghc-prim,
         ghc-bignum,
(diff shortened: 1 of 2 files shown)


aspiwack:
Immutable isn't sufficient. But because they have a single closure at the root, I agree, all these things are `Movable` by virtue of just forcing this one thunk (`Text` is quite subtle but appears to work).

Bodigrim:
@aspiwack pushed a fix for Ormolu, it should be good now.

aspiwack (approved):

aspiwack:
I'll cut a release with this (I think it's going to be purely a type-class-instance related release), probably tomorrow.

PR #503, opened 2026-03-16
Added System.IO.Resource.withFile

heades:
Similar to System.IO.withFile, opens a file, runs a callback on the handle, and then releases the handle. Just makes porting code over to the linear base easier.

### DIFF
--- a/src/System/IO/Resource/Linear.hs
+++ b/src/System/IO/Resource/Linear.hs
@@ -42,6 +42,9 @@ module System.IO.Resource.Linear
     openBinaryFile,
     System.IOMode (..),
 
+    -- *** Compatibility functions
+    withFile,
+
     -- ** Working with Handles
     hClose,
     hIsEOF,
@@ -52,10 +55,7 @@ module System.IO.Resource.Linear
     hPutStrLn,
     hSeek,
     System.SeekMode (..),
-    hTell,
-
-    -- *** Compatibility functions
-    withFile,
+    hTell,    
 
     -- * Creating new types of resources
     -- $new-resources
--- a/src/System/IO/Resource/Linear.hs
+++ b/src/System/IO/Resource/Linear.hs
@@ -38,8 +38,7 @@ module System.IO.Resource.Linear
     Handle,
 
     -- ** File I/O
-    openFile,
-    withFile,
+    openFile,    
     openBinaryFile,
     System.IOMode (..),
 
@@ -55,13 +54,16 @@ module System.IO.Resource.Linear
     System.SeekMode (..),
     hTell,
 
+    -- *** Compatibility functions
+    withFile,
+
     -- * Creating new types of resources
     -- $new-resources
     Resource,
     release,
     unsafeAcquire,
     unsafeFromSystemIOResource,
-    unsafeFromSystemIOResource_,
+    unsafeFromSystemIOResource_,    
 
     -- * Deprecated symbols
     UnsafeResource,
--- a/src/System/IO/Resource/Linear.hs
+++ b/src/System/IO/Resource/Linear.hs
@@ -39,6 +39,7 @@ module System.IO.Resource.Linear
 
     -- ** File I/O
     openFile,
+    withFile,
     openBinaryFile,
     System.IOMode (..),
 
--- a/src/System/IO/Resource/Linear.hs
+++ b/src/System/IO/Resource/Linear.hs
@@ -38,7 +38,7 @@ module System.IO.Resource.Linear
     Handle,
 
     -- ** File I/O
-    openFile,    
+    openFile,
     openBinaryFile,
     System.IOMode (..),
 
@@ -55,7 +55,7 @@ module System.IO.Resource.Linear
     hPutStrLn,
     hSeek,
     System.SeekMode (..),
-    hTell,    
+    hTell,
 
     -- * Creating new types of resources
     -- $new-resources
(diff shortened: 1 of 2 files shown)


src/System/IO/Resource/Linear/Internal.hs
```diff
@@ -115,6 +116,14 @@ instance Control.Monad RIO where
 
 type Handle = Resource System.Handle
 
+-- | See @System.IO.'System.IO.withFile'@
```
aspiwack: I think I want something to avoid misleading the reader into thinking that `withFile` is the recommended pattern with `RIO` as it is with `IO`.

An option would be something like:

```suggestion
-- | This function is provided for compatibility with @System.IO.'System.IO.withFile'@, to facilitate porting code to the @RIO@ monad. Since, the @RIO@ monad guarantees the resources are properly released, even when an exception is raised, it isn't recommended to use 'withFile' in new code.
```

Another option would be to list the function under `-- *** Compatibility functions` section in the export list of `System.IO.Resource.Linear`, maybe?

aspiwack (commented):
Thanks, yes, this sounds quite reasonable.

src/System/IO/Resource/Linear/Internal.hs
heades: Thanks for the feedback! This makes sense, I figured I'd need to add something. I'll make these changes later today.

heades:
I committed the changes. I did both, I figure it's safer that way.

src/System/IO/Resource/Linear.hs
```diff
@@ -54,13 +54,16 @@ module System.IO.Resource.Linear
     System.SeekMode (..),
     hTell,
 
+    -- *** Compatibility functions
+    withFile,
+
```
aspiwack, line 59: I think it would be better to move the sub-subsection at the end of `File I/O` rather than the end of `Working with Handles`, wouldn't it?

(sorry, would have done a suggestion, but it seems to handle lines out of the diff poorly, and I don't want to accidentally create broken diffs)

heades: No sorry needed! Even if these are small changes, I'm learning a lot about the processes y'all like to follow so I can be better prepared for future contributions. :-)

heades: I made the changes, see the most recent commit.

treeowl:
Why doesn't this have exception handling like the version in `System.IO`?

heades:
> Why doesn't this have exception handling like the version in `System.IO`?

I started to write it in the same style as `System.IO.withFile`, but I ended up doing the above based on the comments in the file. I was worried maybe using `bracket` style was non-linear or some other reason. I can try to rewrite it so that it's more than just a "use this, but eventually replace it" function if that's preferred.

treeowl:
> > Why doesn't this have exception handling like the version in `System.IO`?
> 
> I started to write it in the same style as `System.IO.withFile`, but I ended up doing the above based on the comments in the file. I was worried maybe using `bracket` style was non-linear or some other reason. I can try to rewrite it so that it's more than just a "use this, but eventually replace it" function if that's preferred.

I don't know terribly much about the linear stuff, but exception safety is the raison d'etre of `withFile`. It seems super weird not to have that. I imagine that the style in `System.IO` is overkill here, since I don't imagine we deal with handles that carry finalizers, but I haven't dug into that code.

heades:
> > > Why doesn't this have exception handling like the version in `System.IO`?
> > 
> > 
> > I started to write it in the same style as `System.IO.withFile`, but I ended up doing the above based on the comments in the file. I was worried maybe using `bracket` style was non-linear or some other reason. I can try to rewrite it so that it's more than just a "use this, but eventually replace it" function if that's preferred.
> 
> I don't know terribly much about the linear stuff, but exception safety is the raison d'etre of `withFile`. It seems super weird not to have that. I imagine that the style in `System.IO` is overkill here, since I don't imagine we deal with handles that carry finalizers, but I haven't dug into that code.

In `RIO`, when we run it using `run`, exceptions will be raised just like when we use `bracket`. So I don't think it's necessary to do it here, and to make it work we would have to do some non-linearity. This is really just a convenience function to port code without having to abandon `withFile` right away.

treeowl:
Now I understand. The reason for not recommending it in new code is that it's not necessary and it can keep the file open longer than necessary? Perhaps the documentation could clarify that?

heades:
> Now I understand. The reason for not recommending it in new code is that it's not necessary and it can keep the file open longer than necessary? Perhaps the documentation could clarify that?

I reworded the doc to be a little more clear.

aspiwack:
Sorry. I got swamped and this sort of skipped my mind for a while. I pushed a bit of simplification on the documentation (plus some whitespaces).

Now, with a bit of a clearer head, I see an issue with the function. It has type.

```haskell
withFile :: FilePath -> System.IOMode -> (Handle %1 -> RIO (Ur r, Handle)) -> RIO (Ur r)
```

So here is some funny code you can do with this

```haskell
withFile "a/path" ReadWriteMode $ \h -> Linear.do
  h' <- openFile "another/path"
  release h
  …
  return (Ur …, r)
  ```

aspiwack:
I clicked “send” before finishing my message, sorry. Question: is this a problem we care about? It's probably not too serious, but I'd like your opinion on that as a client for this function.

Another issue that I can spot is that the callback function is unrestricted, this means it can't talk of other linear resources. Is this a problem? (but we take a linear callback instead, we can have even funkier behaviour in the style of the example above).

heades:
Good point! Let me close this for now, and think about it a bit.

PR #505, opened 2026-04-27
Coerce `System.IO` and `Linear.IO` into `RIO`

dcastro:
As discussed in https://github.com/tweag/linear-base/discussions/504, this PR adds/exports these functions from the RIO module: `fromIO`, `fromSystemIO` and `fromSystemIOU`.

It also adds a `MonadIO` instance for `RIO`.

### DIFF
--- a/src/Control/Monad/IO/Class/Linear.hs
+++ b/src/Control/Monad/IO/Class/Linear.hs
@@ -7,6 +7,8 @@ import qualified Control.Functor.Linear as Linear
 import Prelude.Linear
 import qualified System.IO as System
 import qualified System.IO.Linear as Linear
+import System.IO.Resource.Linear (RIO)
+import qualified System.IO.Resource.Linear as RIO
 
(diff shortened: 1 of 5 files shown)


aspiwack:
I think this works. Give me a couple of days to try and find a way to break things. I don't think there is, but I want to give it a good go just in case.

dcastro:
> I think this works. Give me a couple of days to try and find a way to break things. I don't think there is, but I want to give it a good go just in case.

Yes of course, I appreciate it! :grin:

aspiwack (approved):
I can't shake an uneasy feeling that I'm missing something. But the problem is probably me, not your proposal. I've tried quite hard to find a fault with it, I'm pretty sure it's all correct. Let me merge.

dcastro:
> I can't shake an uneasy feeling that I'm missing something. But the problem is probably me, not your proposal. I've tried quite hard to find a fault with it, I'm pretty sure it's all correct. Let me merge.

Many thanks for being thorough! Cheers

PR #506, opened 2026-05-01
Add MonadIO instances for StateT and ReaderT

dcastro:
This small PR adds MonadIO instances for `StateT` and `ReaderT`


aspiwack (approved):
Good idea. Thanks.

PR #507, opened 2026-05-12
Flakify the development shell

### LINKED ISSUE #431: Provide Ormolu via Nix
Ormolu has very slow to build dependencies (essentially ghc-lib-parser). This has been an obstacle to contributing, and makes the CI very slow on the first push of a new branch.

Currently, the `./format` script build Ormolu itself, via `stack build`. This is convenient (calling `./format` is machine independent). The main reason why we went this route back then, was that the version of Ormolu that was available on Nixpkgs was too old, we needed a recent one. Now, I believe that it should be easy to provide `Ormolu` via Nixpkgs.

This doesn't prevent contributors who want to provide Ormolu via Stack to call `stack build ormolu && stack exec -- ./format`. It makes the management of dependencies more package-manager agnostic. And will let us download cached Ormolu when using Nix (in particular on CI).

aspiwack:
I'm shaving a yak. I've been having trouble with the version of Stack that the shell provides getting older and it not being able to upload to Hackage. And this time I had enough. But the upgrade process is cumbersome with Niv, and so I preferred converting to a flake.

[ETA: this also bumps the Nix and GHC versions; to avoid playing with the `flake.lock` manually]

Incidentally, this closes #431 . It was time.


aspiwack:
This has implied
- A couple of warnings to silence to comply with newer GHC
- Bumping Cabal's index state to get new versions of some packages (notable `hashtables`) which are hopefully compatible with the newer GCC provided by Nix.

bench/Data/Mutable/HashMap.hs
```diff
@@ -10,6 +10,11 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TupleSections #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
+
+-- `-Wno-unused-imports`: so that we can import `foldl'` without GHC 9.10+
+-- complaining. Inelegant, but this is just a test file, not the end of the
+-- world.
```
aspiwack: Not my proudest moment, to be honest.

.github/workflows/ci.yaml
```diff
@@ -27,23 +27,23 @@ jobs:
         key: cabal-deps-${{ runner.os }}-${{ hashFiles('nix/sources.json') }}-${{ matrix.ghc-version }}-v${{ env.cache-invalidation-key }}-${{ hashFiles('linear-base.cabal') }}-${{ github.sha }}
         restore-keys: cabal-deps-${{ runner.os }}-${{ hashFiles('nix/sources.json') }}-${{ matrix.ghc-version }}-v${{ env.cache-invalidation-key }}-${{ hashFiles('linear-base.cabal') }}-
     - name: Build Nix dependencies
-      run: nix-shell --arg ghcVersion '"${{ matrix.ghc-version }}"' --arg installHls 'false' --pure --run "echo '=== Nix dependencies installed ==='"
+      run: nix develop .#devShells.x86_64-linux.ci-"${{ matrix.ghc-version }}" --command echo '=== Nix dependencies installed ==='
```
tbagrel1, line 30: Can we still use `nix develop` without extra args, and get a suitable dev env?

bench/Data/Mutable/HashMap.hs
tbagrel1: I'm just wondering why it is part of this PR. Is using flake having an impact here?

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; inherit overlays; };
+
+        # need to match Stackage LTS version from stack.yaml snapshot
+        ghc-version = "ghc9103";
```
tbagrel1: I would rename to `stack-ghc-version`

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+          ];
+
+        mkCIShellFor = ghc-ver: pkgs.mkShell {
+          # Set UTF-8 local so that run-tests can parse GHC's unicode output.
+          LANG="C.UTF-8";
+
+          buildInputs = buildTools "${ghc-ver}";
+        };
+
+        ci-shells = builtins.listToAttrs (map (ver:
```
tbagrel1, line 44: Does these shells allow `stack`? Or are they only cabal-based?

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+
+        devShells = ci-shells // {
+          default = pkgs.mkShell {
+            # Set UTF-8 local so that run-tests can parse GHC's unicode output.
+            LANG="C.UTF-8";
+          
+            buildInputs = buildTools ghc-version ++ devTools;
+          };
+
+          ci-stack = mkCIShellFor ghc-version;};
```
tbagrel1: It seems the other ci-shells also have `stack` in their dep list? Or am I missing something.

In general, I would expect:
- `shell-cabal-9.6`, `shell-cabal-9.8`, `shell-cabal-9.10`,..., `shell-stack-stackage` if cabal shells are not providing/not compatible with stack and vice-versa
- `shell-cabal-9.6`, `shell-cabal-9.8`, `shell-cabal-9.10`,..., `shell-stackage` if cabal shells are not providing/not compatible with stack but the last one allow to work with both cabal+stack
- `shell-9.6`, `shell-9.8`, `shell-9.10`,..., `shell-stackage` if all shells are providing both cabal+stack; with `shell-stackage` being the default one

I'm not sure the `ci-` prefix is bringing much here, since I don't think we have different devshell targets for actual developpers?

tbagrel1 (approved):
I have a couple of questions:
1. Is it necessary to have two different dev-shell targets `ci-<ghc>` and `ci-stack`? It is a bit confusing since per `flake.nix`, it seems the slack dev-shell target depends on a GHC version parameter, despite not having the GHC version explicitely in its name like the other target. 
2. if we omit the parameters to `nix develop`, do we still get a

flake.nix
aspiwack: I called them `ci` to signify that they have minimal build tools. Whereas the real dev shell has dev tools as well, such as hls. Would there be a better name?

bench/Data/Mutable/HashMap.hs
```diff
@@ -10,6 +10,11 @@
...
```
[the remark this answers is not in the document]
aspiwack: I certainly won't do this. But this point at the comment being confusing, so let me fix.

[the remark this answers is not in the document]
aspiwack: Ok, I found a “better” way. Import `foldl'` qualified and use it qualified, don't disable the warning. I think I'm less ashamed of this solution.

flake.nix
tbagrel1: Indeed, I got confused here. 

I'm wondering if it wouldn't be better to have these with dev tools enabled by default (conditionned by a switch), and flip the switch off in the CI (to not use the dev tools)

Because when cabal CI breaks, and one wants to see what went wrong and reproduce locally, I guess it may be good to have access to dev tools? That being said, dev tools are currently not indexed by `ghc-ver`, but by the global `ghc-version`, so it would require a slightly larger change.

aspiwack:
@tbagrel1 
(there's no thread for the following, for whatever reason, answering here)
> I'm just wondering why it is part of this PR. Is using flake having an impact here?

I bumped the Nixpkgs and GHC versions at the same time, by laziness (I should have split it in another PR or something, but it didn't seem important enough). And the consequence is that in newer GHC, `foldl'` is in the `Prelude`. Which trigger warnings. But per https://github.com/tweag/linear-base/pull/507#discussion_r3227062571 I found a less annoying solution.

flake.nix
aspiwack: There's no switch in a Flake. Only a list of named shells.
We usually don't want HLS to be present in the CI, it can be a problem (it tends to get rebuilt and stuff). But it's possible to double these shells and make a CI and a Dev version for each. Is it worth it?

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+
+        buildTools = ghc-ver:
+          let hask = pkgs.haskell.packages."${ghc-ver}"; in
+          [
+            hask.ghc
+            hask.cabal-install
+            stack-wrapped
+            cabal-docspec
+            haskell.ormolu # Haskell formatter
+          ];
```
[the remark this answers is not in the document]
aspiwack: It isn't quite necessary to do that. But some of the dev shells do need to do some rebuild and take significantly more time, so maybe minimising the CI shells can save significant time.

.github/workflows/ci.yaml
aspiwack: Yes. With everything and the stack ghc version.

flake.nix
tbagrel1: I'll let you decide what is the best course of action here.

Having ci shells + dev shells is a bit verbose, but the most complete option while still being efficient

Having only ci shells can be a bit annoying when trying to debug something that breaks specifically in a particular cabal version

Having dev shells only can be quite heavy/annoying in normal circumstances where the CI is not supposed to break ^^ So I guess it's the worst option

flake.nix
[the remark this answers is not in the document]
aspiwack: A night of sleep did me good. The tools that I could strip out of these CI shells are all independent of `hask`. And the dev shell at the stack version takes about 30s to install; this is an upper limit on how much time can be saved. I think it's not worth the extra complexity in the flake, so I'll leave it this way.

flake.nix
aspiwack: They have stack, but stack won't work unless the installed GHC is at the right version. It could be optimised by removing Slack when it's not relevant but per https://github.com/tweag/linear-base/pull/507#discussion_r3231969819 I won't be doing that.

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+        cabal-docspec = import ./nix/cabal-docspec.nix { inherit pkgs; };
+
+        buildTools = ghc-ver:
+          let hask = pkgs.haskell.packages."${ghc-ver}"; in
+          [
+            hask.ghc
+            hask.cabal-install
+            stack-wrapped
+            cabal-docspec
+            haskell.ormolu # Haskell formatter
```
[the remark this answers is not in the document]
aspiwack: Likewise https://github.com/tweag/linear-base/pull/507#discussion_r3231969819 ; I won't be doing that.

flake.nix
aspiwack: I added the dev shells then.

```
$ nix flake show
…
    └───x86_64-linux
        ├───ci-910: development environment 'nix-shell'
        ├───ci-912: development environment 'nix-shell'
        ├───ci-96: development environment 'nix-shell'
        ├───ci-98: development environment 'nix-shell'
        ├───ci-stack: development environment 'nix-shell'
        ├───default: development environment 'nix-shell'
        ├───dev-910: development environment 'nix-shell'
        ├───dev-912: development environment 'nix-shell'
        ├───dev-96: development environment 'nix-shell'
        └───dev-98: development environment 'nix-shell'
```

flake.nix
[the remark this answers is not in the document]
aspiwack: I ended up doing as suggested anyway because it simplified the flake with the changes from https://github.com/tweag/linear-base/pull/507#discussion_r3232043039 . Still going for simplicity as much as possible, at the cost of some potential rebuild if Ormolu isn't properly cached in some version of GHC.

aspiwack:
Ok, so, most of the shells spent ~20min (which is wasteful, even we got almost all the relevant feedback within 2 min) building Cabal (and Ormolu, but mostly Cabal). I realised (remembered? (stopped being an idiot?)) that Cabal and Ormolu were available at toplevel (and they don't need in any way to be built with the same compiler we install; the toplevel version of Cabal supports all available compilers).

So while it's not the end of the world, it's easy to make the CI much less wasteful. So I did that. Which concludes this session of yak-shaving.

PR #509, opened 2026-05-17
Import Multiplicity from GHC.Types

tomsmeding:
GHC 9.2 did not yet export Multiplicity from GHC.Exts. The rest of the repository imports Multiplicity from ghc-prim:GHC.Types, from which the type has always been exported, but this one occurrence in Control.Functor.Linear.Internal.Reader didn't.

### DIFF
--- a/src/Control/Functor/Linear/Internal/Reader.hs
+++ b/src/Control/Functor/Linear/Internal/Reader.hs
@@ -35,7 +35,7 @@ import qualified Data.Functor.Linear.Internal.Functor as Data
 import Data.Kind (FUN)
 import Data.Unrestricted.Linear.Internal.Consumable
 import Data.Unrestricted.Linear.Internal.Dupable
-import GHC.Exts (Multiplicity (..))
+import GHC.Types (Multiplicity (..))
 import Prelude.Linear.Internal (runIdentity', ($), (.))
 
 -- # Linear ReaderT


aspiwack:
Does linear-base even compile with GHC 9.2 (other than this infelicity?). I confess that I'm rather a burn-bridges kind of person, and I tend to drop support for old versions as soon as they become inconvenient.

I agree that it's better to be uniform, but on the other hand, `GHC.Types` is is `ghc-prim` and `GHC.Exts` is in `base`. So it's generally preferable to import from the latter (maybe we can even drop our direct dependency on `ghc-prim`, that could be nice; even if for a library like linear-base, it's not particularly surprising to depend on ghc-prim).

On the other other hand, I assume that this pull request comes from a need. Could you elaborate the need a little?

tomsmeding:
For me personally, it doesn't matter if you fix the import or increase your lower bound on `base`. :)

I operate the [playground](https://play.haskell.org) and had to rebuild the package set for GHC 9.2; my script happily told me that `cabal` was happy with a package set that included linear-base, but the build then proceeded to fail. Seeing it was linear-base I decided to spend a few minutes seeing if I could fix it, and because it was just an import I opened a PR.

I appreciate your arguments, but `cabal test` succeeds, and if really all that is needed is importing some things from ghc-prim to support 9.2, I'd suggest potentially keeping a few additional users happy.

aspiwack (approved):
This is fair enough. I'm really surprised that we don't have more incompatibilities with GHC 8.2, though.

aspiwack:
Published as 0.8.1 https://hackage.haskell.org/package/linear-base-0.8.1

I should have done this earlier, sorry, it skipped my mind for some reason.

